PR #1419, opened 2013-04-03
Improved iPad Support

tqc:
This includes a couple of fixes to make CodeMirror work better on an iPad.

1. Mobile Safari handles z-index within a specific parent element and positioning type - with the previous version the selection would hide the code and the gutter background would hide the line numbers. I have moved lineDiv in front of selectionDiv and sizer in front of gutters. This does mean that the gutter background won't cover horizontally scrolled code, but using a fixed gutter runs into other issues on iOS anyway.

2. I have replaced onMouseDown with a set of touch event handlers when running on iOS.
 * Mobile safari does not generate mouse events until it knows if it is a click or double click - i.e. it waits 300ms then fires all the events at once, so anything timing related won't work. The new implementation moves the cursor immediately and decides later if it also needs to select the word/line.
 * To avoid conflicts with scrolling, drag and selection are triggered with either a 100ms pause before movement or movement on the second touch.

### DIFF
--- a/demo/ipad.html
+++ b/demo/ipad.html
@@ -1,0 +1,83 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>CodeMirror: iPad Demo</title>
+
+        <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


marijnh:
This should be superseded by the recently landed mobile work.

PR #2328, opened 2014-03-04
Add delLineRight command for sake of symmetry

kolya-ay:


jasonLaster:
Sublime and many other text editors support deleting to the right of the cursor with. `<cmd>+delete` or `<meta>+delete`



It'd be great to have support for `delLineRight` so that it'd be possible to add this key map if we want it.

marijnh:
You could add it to `keymap/sublime.js`, along with the default binding for the Sublime keymap, and submit a pull request. I still don't want it in the core library.

PR #2581, opened 2014-05-24
auto expand folded regions when fold ranges become invalid

thehogfather:
This addresses an issue where fold markers do nothing if a folded region is made invalid  as a result of an undo or delete action. To reproduce:
* Enter the following fragment
```javascript
function () {

}
```
* Fold the function
* Delete the closing brace

Result is that the fold marker no longer expands the folded region, although one is still able to click text marker or indeed expand the fold by navigating the cursor into the region.

In an attempt to address this issue, this pull request expands a folded region when they become invalid.
Credits to @JeffryBooher for spotting the issue.

### DIFF
--- a/addon/fold/foldgutter.js
+++ b/addon/fold/foldgutter.js
@@ -69,11 +69,16 @@
     var opts = cm.state.foldGutter.options, cur = from;
     cm.eachLine(from, to, function(line) {
       var mark = null;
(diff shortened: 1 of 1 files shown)


marijnh:
I am not sure how I failed to respond to this. Apologies for that. I think a more lightweight solution is to fix the unfold markers so that they work even if the foldable range is no longer there. Attached patch does this.

PR #2853, opened 2014-10-10
fixed issue where extra spaces added to class attr

decherneyge:
when ever a class was added to a node, regardless of if a space was present at the end of the class attribute an additional one was added to ensure the new class was not appended to the end of the last class. This caused an ever increasing amount of spaces in the class attribute. Updated the addClass function to trim the current value of the class attribute, append a space and then append the new class


marijnh:
This appears to be fixed by 4bce5342 . Next time, when adding a patch to a pull request, also add a comment, so that the repo owner gets notified that you want something reviewed.

PR #2858, opened 2014-10-12
Added the possibility to automatically jump to first diff ocurrence

plafue:
The merge addon now can jump to the first difference if configured so. I'm not quite happy with the hard coded 8 in line 64, maybe someone has a tip how to easily get the font size of the editor?


marijnh:
You should add a comment when you add patches -- otherwise, I won't notice them.

Anyway, your implementation still isn't really what I was looking for. See my attached patch for an alternative.

PR #2995, opened 2014-12-19
Modified markText to allow css

GabrielGhe:
I modified `markText()` to allow css.

Before, you could only style it by using a className

```
cm.markText(startPos, endPos, { className: "someClass" });
```

Now you can add css dynamically.

```
var marker = cm.markText(startPos, endPos, { css: "background-color: blue; color: red" });
marker.clear();
marker = cm.markText(startPos, endPos, { css: "background-color: orange" });
```


charlieroberts:
Just another thank you for this to Gabriel... exactly what I needed.

PR #3013, opened 2015-01-01
2156

ankitstarski:
To fix this I had to introduce new state parameter `cursorHighlight` that tells if
cursor is before colon or not.


marijnh:
Thank you. Merged as 9e56052a7a902229b

PR #3014, opened 2015-01-02
Add JSONiq mode

wcandillon:
The syntax highlighting is computed by [XQLint](https://github.com/wcandillon/xqlint).
XQLint is heavily tested against the W3C test suite and is already used in the ACE editor.

### DIFF
--- a/mode/jsoniq/index.html
+++ b/mode/jsoniq/index.html
@@ -34,161 +34,246 @@
 
     <div class="cm-s-default">
 	<textarea id="code" name="code">
-xquery version &quot;1.0-ml&quot;;
-(: this is
- : a
-   "comment" :)
-let $let := &lt;x attr=&quot;value&quot;&gt;&quot;test&quot;&lt;func&gt;function() $var {function()} {$var}&lt;/func&gt;&lt;/x&gt;
-let $joe:=1
-return element element {
-	attribute attribute { 1 },
-	element test { &#39;a&#39; },
-	attribute foo { &quot;bar&quot; },
-	fn:doc()[ foo/@bar eq $let ],
-	//x }
+import module namespace config = &quot;http://apps.28.io/config&quot;;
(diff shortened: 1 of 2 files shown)


marijnh:
I do not want to include all of XQLint into the mode file, since it's huge, under another license, and likely to go out of date. Could you rewrite this to load it as a dependency? (And have the demo get it from some CDN, so that we don't have to include it in the distribution.)

wcandillon:
@marijnh Thank you so much for the review! 
My use case is to have the JSONiq code highlighting available on http://brackets.io. Anything that helps me towards that, I'm in.

Can you give me a pointer on how such dependency work? Is it like an npm dependency? I couldn't find a build system for the editor. I assume for other mode have similar needs (like javascript and jshint). Couldn't find a relevant pointer in the source code.

The part form XQLint that is included in the mode, is only the browserify build of the JSONiq lexer classes: https://github.com/wcandillon/xqlint/blob/master/Gruntfile.js#L107
The file is big because the lexer is complete.

Regarding the lexer going out of date, I am sure to understand how this situation is different from any other mode where the language might evolve.

marijnh:
Since this is a rather specialized use case, and I really don't want to pull in blobs of 3rd party code into my distribution, I think it is best if you create a separate repository for this, and base your own brackets plugin on it.

wcandillon:
@marijnh that's a bit harsh. Tools like jshint or csslint are not used in Codemirror? My use case is the same. XQLint is used in both ACE and Atom to syntax highlighting and markers, please let me know what I have to do to bring this to codemirror as well.

marijnh:
> Tools like jshint or csslint are not used in Codemirror?

There is integration code for them under `addon/lint`, but they are specified as _depdendencies_ of that code, not pasted in wholesale. There is simply no way I'm going to include thousands of lines of duplicated code (which is under a different license, no less).

PR #3016, opened 2015-01-03
fix GLSL keywords

sapphire-al2o3:


marijnh:
Thank you. Merged as 5619d83

PR #3018, opened 2015-01-04
fixed onKeyEventTargetKeyDown in keymap/vim.js

esdras:
when CodeMirror.keyName returns a non string object like false or true, the indexOf function is undefined. This causes the undefined is not a function error to be thrown.


mightyguava:
Ah good catch, f97df90 adds checking for falsy return from `keyName`

esdras:
glad I could help :-)

PR #3021, opened 2015-01-08
add svxverilog mode theme and demo files

ratchup:
Here is the pull req to add svxverilog mode. (Ref: Jan 6 e-mail)


marijnh:
Is this an entirely different language from [Verilog/SystemVerilog](http://codemirror.net/mode/verilog/)? If not, would it make sense to write this as a configuration of that mode instead?

ratchup:
Re: Is this an entirely different language from Verilog/SystemVerilog
<http://codemirror.net/mode/verilog/>?
No, this language is an extended form of system verilog. (The extended
system verilog is submitted to a preprocessing tool which produces system
verilog.) Right now, this extension is experimental and not part of a
standard.

Re:  If not, would it make sense to write this as a configuration of that
mode instead?
I did consider adding this extension to the verilog mode module. (I'm
guessing that would be the impact of adding a configuration of the verilog
mode.) However, I had a couple of concerns: 1) I was unsure how this
approach (of adding configurations) would scale (in terms of
maintainability) for two (or more)  pre-processor modes; 2) This tool (SVX)
is currently in development so the editor support needs might change. (And
I thought a separate mode would allow for more flexibility than an
integrated mode. I could be wrong about this, but it seemed to me that
having multiple indentation flows [one for the main mode and (potentially)
others for the different configuration flows] would make for a busy module.)

(I tried to make a svx mixed mode [like html mixed

ratchup:
Re: (I tried to make a svx mixed mode [like html mixed ...

Sorry, I was trying to type too quickly and I accidentally sent the message
before I was done.

So,
I tried to make a svx mixed mode [like html mixed mode] which would call
svxverilog and verilog modes. But right now, that approach isn't working
for me. (I'm having CM framework problems.)

Regards,

Ray

marijnh:
If the language is unstable, maybe it is better to hold off on integrating this in the distribution for now?

But even if we put this in the main distribution, I think it would be preferable to add configurability to the Verilog mode. Duplicated code is a maintenance burden and will prevent improvements to one of the modes from helping the other mode. See also the clike, css, and sql modes for examples of modes that handle families of languages.

ratchup:
Re:
If the language is unstable, maybe it is better to hold off on integrating
this in the distribution for now?

I think of the preprocessor language as experimental- within Intel the
consumer tool (the preprocessor) underwent several iterations before being
released as open source. Some flows supported by the tool, such as the HDL
pipeline extensions, are more developed than other flows.  So, there is
room for more development and hence, I consider the language experimental.
Anyway and however, I think that  holding off on integrating this into the
distribution for now is not unreasonable from the maintenance point of
view. I want to be able to provide to those  who are interested (which, I
guess, is a small population of the code editor user community) this addon
mode to support user use of the preprocessor tool and to facilitate
development of the tool. (I think there is a chicken and egg issue here-
the tool needs to be used in order to facilitate further development so
that the tool can be used more...)   (I guess I'm asking the question:
"What is the best way to provide EDA Website code editor for the
preprocessor tool and also minimize the ongoing code editor maintenance
effort?" However, I don't know much about the CM mode interface and so
can't give a good answer to the second clause...)

Re:
But even if we put this in the main distribution, I think it would be
preferable to add configurability to the Verilog mode. Duplicated code is a
maintenance burden and will prevent improvements to one of the modes from
helping the other mode. See also the clike, css, and sql modes for examples
of modes that handle families of languages.

I'll look at the suggested modes and see what I can do...

 Regards,

Ray

ratchup:
I redid the request as a config option in verilog mode. I tried to use an aspect oriented approach to add configurability to the Verilog mode.

marijnh:
I took a look at this, but as it stands the code really isn't clean enough to be integrated. For a start, its indentation and whitespace are different from the other modes, and rather inconsistent internally. More importantly, the actual code itself seems alarmingly confused. You are initializing variables to an empty object only to directly overwrite them, you are returning objects from which you use only one field (in the hooks), you are needlessly touching the internals of the string stream abstraction (its `pos` and `string` fields), and your hooks' logic, instead of simply dispatching on the content of the line, first sets some variables, and then later dispatches on those variables. It is also a really bad idea to _mutate_ the mode's configuration object, I'm not sure what you were trying to accomplish with that, but it's going to touch other instances of the same mime mode, since they share that object.

So if you want this merged, tighten up the code style, remove superfluous code and indirection so that someone who is not you has a chance of understanding what is going on, and please also try to make the configuration less invasive -- there's not need to have multiple `vx`-specific properties on the state even when in another configuration -- your hooks can add those when they need them. Also try making the hook mechanism general, not vx-specific, and try to give the individual hooks more obvious names.

ratchup:
Thanks for looking at my code.  I'll work on cleaning up the hygienic
problems of indentation, respecting the API ("touching the internals of the
string stream abstraction"),  removing superfluous code and indirection,
making the svx configuration less invasive, and making the hook mechanism
more general.  I hadn't intended to *mutate* the mode's configuration
object, and would like to fix that problem. However, I'm in  need of some
guidance on this exact issue. Is there someone that I could talk to or a
pointer that I could look at to resolve the mutation problem?  Thanks...

mode/verilog/verilog.js
```diff
@@ -354,11 +372,185 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
...
+               if ((0 == stream.pos) && (/\\SV/.test(stream.string))) {
+                 curPunc = (/\\SVX_version/.test(stream.string))
+                   ? "\\SVX_version" : stream.string;
+                 stream.skipToEnd();
+                 if (curPunc=="\\SV" && state.vxCodeActive) {codeRegion = "2v";};
+                 if (/\\SVX/.test(curPunc) && !state.vxCodeActive) {codeRegion ="2vx";};
+                 if (curPunc=="\\SVX_version" && state.vxCodeActive) {codeRegion = "2vx";};
+                 style=(thisMode.svxkeywords.propertyIsEnumerable(curPunc))
+                   ? thisMode.svxkeywords[curPunc] : "builtin";
+                 thisMode.svxCurCtlFlowChar  = thisMode.svxPrevPrevCtlFlowChar
```
marijnh: There are various points where you are assigning to properties of `thisMode`, which is, unless I'm really confused, referring to the `parserConfig` object passed to the mode. This object is shared by all instances of the mode (it is the value passed to `defineMIME`), and thus you probably shouldn't be changing it.

ratchup: Re: There are various points where you are assigning to properties of
thisMode, which is, unless I'm really confused, referring to the
parserConfig object passed to the mode... you probably shouldn't be
changing it.

Yes. And thanks for the clarification about the mutation problem...

ratchup:
I fixed the "mutate the mode's configuration object" problem; made the vx configuration flow less invasive to the verilog mode flow; gave the individual hooks more obvious names;  removed superfluous code; and tried to make the hook mechanism more general. Please let me know what I should do next. Thanks.

mode/verilog/verilog.js
```diff
@@ -354,11 +360,189 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
...
+    svxindentUnit: 2,
+    svxkeywords: {
+      "\\SVX":"keyword", "\\SV":"keyword", "m4":"def", "M4":"def", "\\SVX_version":"keyword"},
+    svxkpScopePrefixs: {
+      "**":"variable-2", "*":"variable-2", "$$":"variable", "$":"variable",
+      "^^":"attribute", "^":"attribute"},
+    svxchScopePrefixs: {
+      ">":"property", "->":"property", "-":"hr", "|":"link", "?$":"qualifier", "?*":"qualifier",
+      "@-":"variable-3", "@":"variable-3", "?":"qualifier"},
+    svxisOperatorChar:  /[\[\]=:]/
```
marijnh: The 5 properties above could simply be variables. That'd also remove the need to pass the `thisMode` around, I believe.

mode/verilog/verilog.js
```diff
@@ -354,11 +360,189 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
...
+        this.vxIndentRq = 0;
+      },
+      "token": function(stream, state, _thisMode) {
+        if (state.vxContext.vxCodeActive==true && stream.sol() && state.vxContext.svxCurCtlFlowChar != "") {
+          state.vxContext.svxPrevPrevCtlFlowChar = state.vxContext.svxPrevCtlFlowChar;
+          state.vxContext.svxPrevCtlFlowChar = state.vxContext.svxCurCtlFlowChar;
+          state.vxContext.svxCurCtlFlowChar  = "";
+        }
+      },
+      "svxGenIndent": function(stream, state, thisMode) {
```
marijnh: How about making this a function, rather than putting it in the hooks object, though no dynamic dispatch is needed?

mode/verilog/verilog.js
```diff
@@ -354,11 +360,189 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
...
+          style=(thisMode.svxkeywords.propertyIsEnumerable(curPunc))
+            ? thisMode.svxkeywords[curPunc] : "builtin";
+          state.vxContext.svxCurCtlFlowChar  = state.vxContext.svxPrevPrevCtlFlowChar
+            = state.vxContext.svxPrevCtlFlowChar = "";
+          if (codeRegion == "2vx") {
+            state.vxContext.svxCurCtlFlowChar  = "\\";
+            vxIndent=thisMode.vxHooks["svxGenIndent"](stream, state, thisMode);
+          }
+          if ("2v"  == codeRegion)  {state.vxContext.vxCodeActive=false;}
+          if ("2vx" == codeRegion)  {state.vxContext.vxCodeActive=true;}
```
marijnh: Why indirect through `codeRegion`, rather than simply doing this right away in the `if` statements above?

mode/verilog/verilog.js
```diff
@@ -17,7 +17,8 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
       statementIndentUnit = parserConfig.statementIndentUnit || indentUnit,
       dontAlignCalls = parserConfig.dontAlignCalls,
       noIndentKeywords = parserConfig.noIndentKeywords || [],
-      multiLineStrings = parserConfig.multiLineStrings;
+      multiLineStrings = parserConfig.multiLineStrings,
+      vxHooks = parserConfig.vxHooks || {}; // vx- verilog extensions
```
marijnh: I'd call this `hooks`, since it could also be used by other dialects, and there's nothing vx-specific about it.

mode/verilog/verilog.js
```diff
@@ -107,7 +107,10 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
   var statementKeywords = words("always always_comb always_ff always_latch assert assign assume else export for foreach forever if import initial repeat while");
 
   function tokenBase(stream, state) {
-    var ch = stream.peek();
+    var ch = stream.peek(), style;
+    if (vxHooks[ch] && ((style=vxHooks[ch](stream, state, parserConfig)) != false)) {return style;}
+    if (vxHooks["tokenBase"] && ((style=vxHooks["tokenBase"](stream, state, parserConfig)) != false)) {return style;}
```
marijnh: Please put spaces around binary operators, everywhere.

ratchup:
I put spaces around the binary operators in the svx flow; changed vxhooks to hooks; removed the codeRegion placeholder var; moved the "svxGenIndent" function out of the object literal hooks area (since dynamic dispatch is not needed); moved the 5 literal properties in the svx mime to variables; and removed thisMode from the parameter lists.

marijnh:
I've merged this as aa46ff58b8c and cleaned it up further in c7562e82f022b6.I am not sure how you failed to notice that the mode didn't work at all when you edited text, because its state didn't properly support copying. I have left out the theme and the demo page, since this appears to be an exceedingly obscure language, and I don't want to pull in any extra baggage because of it.

PR #3025, opened 2015-01-09
Add selectAll to the emacs key map

mrkrd:
Equivalent to the Emacs' command mark-whole-buffer.


marijnh:
Thank you! Merged as 5dd07d60b8b

Issue #2940, opened 2014-11-22
Missing mode file extensions in CodeMirror.modeInfo


silverwind:
It looks like there are a few modes which aren't defining any file extensions. As I've come to rely on extensions for displaying the correct mode for a given filename, it'd be great if at least those modes that have a unique file extension also have them defined in `mode/meta.js`.

I'll be researching some modes myself to see if I can find the extensions for it, but any help is appreciated!

Here's the current list of modes that don't define a file extension:

```
Mode          Mime Type
-----------------------------------------
asterisk      text/x-asterisk
cypher        application/x-cypher-query
django        text/x-django
dockerfile    text/x-dockerfile
gfm           text/x-gfm
http          message/http
javascript    application/ld+json
jinja2        null
mirc          text/mirc
nginx         text/x-nginx-conf
pegjs         null
pig           text/x-pig
rpm           text/x-rpm-changes
sieve         application/sieve
slim          text/x-slim
smartymixed   text/x-smarty
solr          text/x-solr
sql           text/x-mariadb
sql           text/x-mssql
sql           text/x-mysql
sql           text/x-plsql
stex          text/x-stex
textile       text/x-textile
tiddlywiki    text/x-tiddlywiki
tiki          text/tiki
toml          text/x-toml
tornado       text/x-tornado
vbscript      text/vbscript
```

PR #3027, opened 2015-01-11
[meta] Add findModeByFileName, let all meta functions accept uppercase

silverwind:
As discussed in #2940, this adds `findModeByFileName` which takes a filename and checks each mode in `modeInfo` for a new property `file`, which contains a regexp to test against the given filename. If no filename matches, it tries to extract a extension and fall back to `findModeByExtension`.

I also saw the opportunity to let all four meta functions accept data in upper and mixed case too.

Finally, the modeInfo has been extended with filenames, and extensions.This is the result of me googling up modes with no extension associated. There are still modes with no mime or extension, but adding some to these would've been plain guessing.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -2654,12 +2654,13 @@ editor.setOption("extraKeys", {
       properties <code>mimes</code>, which holds an array of MIME
       types for modes with multiple MIMEs associated,
       and <code>ext</code>, which holds an array of file extensions
-      associated with this mode. Three convenience
+      associated with this mode. Four convenience
       functions, <code>CodeMirror.findModeByMIME</code>,
-      and <code>CodeMirror.findModeByExtension</code>,
+      <code>CodeMirror.findModeByExtension</code>,
+      <code>CodeMirror.findModeByFileName</code>
       and <code>CodeMirror.findModeByName</code> are provided, which
-      return such an object given a MIME or extension string. Note
-      that, for historical reasons, this file resides in the
+      return such an object given a MIME, extension, file name or mode name
+      string. Note that, for historical reasons, this file resides in the
       top-level <code>mode</code> directory, not
       under <code>addon</code>. <a href="../demo/loadmode.html">Demo</a>.</dd>
 
--- a/mode/meta.js
+++ b/mode/meta.js
@@ -13,15 +13,15 @@
 
   CodeMirror.modeInfo = [
     {name: "APL", mime: "text/apl", mode: "apl", ext: ["dyalog", "apl"]},
-    {name: "Asterisk", mime: "text/x-asterisk", mode: "asterisk"},
+    {name: "Asterisk", mime: "text/x-asterisk", mode: "asterisk", file: /^extensions\.conf$/i},
     {name: "C", mime: "text/x-csrc", mode: "clike", ext: ["c", "h"]},
-    {name: "C++", mime: "text/x-c++src", mode: "clike", ext: ["cpp", "c++", "hpp", "h++"], alias: ["cpp"]},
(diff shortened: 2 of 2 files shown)


mattpass:
This looks interesting but a couple of the filetypes shouldn't be there?

GitHub Flavored Markdown - isn't just README.md but could be anything with a .md extension such as CONTRIBUTING.md or HISTORY.md
Nginx - can include other .conf files too thru the use of includes say ```include vhosts/more-sites.conf``` or wildcards (``*.conf```)

The other filetype values (Asterisk etc) I'm not sure on but the above I am.

silverwind:
> GitHub Flavored Markdown - isn't just README.md but could be anything with a .md extension such as CONTRIBUTING.md or HISTORY.md

README.md is the only one rendered by default on Github as far as I know. I could see adding others here, but I wonder if these are just loosely based on convention or if there's a defininte list. 

> Nginx - can include other .conf files too thru the use of includes say include vhosts/more-sites.conf or wildcards (*.conf`)

I first thought of allowing paths and using regexp pattern to match stuff like the common names for vhosts files, but decided against it for simplicty. I could easily extend it if requested.

mattpass:
GitHub will render any file ending with the .md extension. See examples here: https://github.com/icetbr/markdown-examples

You also couldn't list Nginx conf files as they too could be any name with the extension .conf

Just don't think you can name specific filenames for either of these languages.

silverwind:
> GitHub will render any file ending with the .md extension.

Yeah, but what I meant is, only a README.md is displayed below a directory's root, making it kind of special. I just went ahead and added your suggested filenames to `gfm`.

> You also couldn't list Nginx conf files as they too could be any name with the extension .conf

If there's path support, I could possibly use `/nginx[\/\\]\.conf$/` and get a good success rate, because these most likely reside in `/etc/nginx.conf`. But I think that would require a `findModeByPath` which would then fall through to `findModeByFileName`, just like I do for extensions. 

Paths would probably a very rare use case, and probably not worth the extra weigth it adds to CM. While I could myself use it, most commonly you only have a file name available in a web app.

silverwind:
Fixed the test failure with a force-push (missing semicolons).

hitsthings:
Just wanted to voice that I also think you should just add the .md extension instead of specific filenames. Nothing about the filename describes the flavor of Markdown, and .md files will always be Markdown.

I like this patch overall, thanks for it! Means we can simplify some of our own code that worked similarly.

silverwind:
> Just wanted to voice that I also think you should just add the .md extension instead of specific filenames. 

I would have if i could, but the extension is already mapped to the vanilla `markdown` mode. If we can agree that Github Markdown is more popular, we *could* remap the extension.

hitsthings:
Ah in that case, I'm the opposite. :) Idremove the added filenames entirely. There is really nothing about those filenames that says their flavor.

silverwind:
Reduced the function a bit. Filenames are defined by a single regexp now.

marijnh:
I think this looks good as-is, except that I don't think case-insensitivity should be hard-coded into the function. Rather, add an `/i` to the regexps that need to be case-insensitive.

silverwind:
Will do later today.

For the case of Markdown (if you agree), I think it's the best possible guess we can make. GitHub-based filenames are pretty common, and (if I remember correctly) the difference between regular and GitHub markdown are subtle enough, so it will render mostly correct in the wrong mode too.

silverwind:
Updated and squashed the commits. Case sensitivity now depends entirely on the regexp.

Also changed `Dockerfile` to be case-sensitive.

marijnh:
Thank you! Merged as 78e06ac. I've removed the line that downcases extensions by default. I understand the rationale, but I've seen people use uppercase letters in extensions in a significant way, so I don't think this is a decision that should be forced on all users of the library.

silverwind:
Awesome, thanks!

PR #3036, opened 2015-01-19
[perl] Add lineComment attribute to Perl mode

vincentwoo:
Also clean up style a bit. `perl.js` is oddly spaced!


marijnh:
Yes, that whole file is rather strange (and I already did some cleaning up). Thanks for the patch. Merged as 5a7affd

PR #3038, opened 2015-01-21
also mention coffeescript-lint addon

forresto:


marijnh:
Thanks. Merged as bf2e307

PR #3043, opened 2015-01-23
[vim] fix join when 'head' before 'anchor' (closes #3042)

### LINKED ISSUE #3042: [vim] Join `J` fails when cursor at top of selection
From http://codemirror.net/demo/vim.html. I set Vim to visual line mode, and attempt to join lines with `J`. The join fails if the caret is at the beginning of the selection. See:

vim-join

I'm guess that, on join, the start / end markers need to be swapped / set appropriately?

kevinushey:


kevinushey:
I also notice that `joinLines` does not place the cursor in the correct location after a join (it should lie on the joined line; instead it is placed ahead)

kevinushey:
I think the issue was related to the 'exitVisualMode' command moving the cursor; we want to ensure we set the cursor position after exiting visual mode.

mightyguava:
Merged, thanks!

PR #3047, opened 2015-01-25
Added Stylus mode

dmitrykiselyov:


marijnh:
Nice! But the indentation doesn't seem to work very well yet -- even after the opening brace in the example, I end up at column zero when I press enter.

marijnh:
Thank you, looks much better now! Merged as ac72e8d23653c47.

(When adding patches to a pull request, the repo owner does not get a notification, so it's usually best to follow up with a comment when you want your new patches to be reviewed -- it was pure luck that I reopened this page and noticed the new commit.)

dmitrykiselyov:
Ok. Now I'll know. Thanks.
It was not pure luck, I mentally asked you to reopen :)

PR #3051, opened 2015-01-26
fix issue - loop fixated in big comment with JSON notation

andreich2013:
Hello, I am a developer on a big service. We use your codeMirror as script editor. Our support team found bug - the loop fixated when JSON notation is inside the big comment (/*...*/) - http://prntscr.com/5xbxph. As the result - script editor crashes and the page is not available for any manipulations. Besides Google Chrome actually crashes too and I need to restart it again. 

I fix it (I hope so :) ), but it should be tested for real.

Best Regards, Andrew!

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -5190,8 +5190,10 @@
     sol: function() {return this.pos == this.lineStart;},
     peek: function() {return this.string.charAt(this.pos) || undefined;},
     next: function() {
-      if (this.pos < this.string.length)
-        return this.string.charAt(this.pos++);
+      if (this.pos < this.string.length) {
+          this.pos = this.pos + 1;
+          return this.string.charAt(this.pos);
+      }
     },
     eat: function(match) {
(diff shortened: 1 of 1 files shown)


marijnh:
You are changing the meaning of the `next` method, which pretty much all 90 currently existing modes rely on. That does not sound like a good idea.

Could you describe the problem in a bit more detail? For one thing, which mode are you using? And is the input in the screenshot (`"/*\n{"`) the input that triggers the issue?

andreich2013:
Yes, maybe that is not a good idea, but it works incorrectly... As I know Douglas Crockford recomended use i = i + 1 instead i++ in his books. 

We are using javascript mode. Yes - "/*\n{" - it is. For more details, this is Appery platform and this is link where our editor is - https://appery.io/servercode/

marijnh:
Entering that code in http://codemirror.net/mode/javascript/ does not cause an infinite loop. Please verify that your issue has not already been fixed.

PR #3054, opened 2015-01-28
Mobile branch: Fix for copying large chunks of text

mihailik:
Selecting and copying a large chunk of text places a single dash into clipboard.

This is related to line 1285, where the input's content is trimmed for performance:

https://github.com/codemirror/CodeMirror/blob/mobile/lib/codemirror.js#L1285

    minimal = hasCopyEvent &&
      (range.to().line - range.from().line > 100 || (selected = cm.getSelection()).length > 1000);
    var content = minimal ? "-" : selected || cm.getSelection();

The problem is apparently a typo during refactoring/renaming from 'd' on the master branch to 'input' here.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1210,7 +1210,7 @@
       function prepareCopyCut(e) {
         if (cm.somethingSelected()) {
           lastCopied = cm.getSelections();
-          if (this.inaccurateSelection) {
+          if (input.inaccurateSelection) {
             input.prevInput = "";
             input.inaccurateSelection = false;
             te.value = lastCopied.join("\n");


marijnh:
Thanks for tracking that down and fixing it. I've merged your patch. (This kind of bug is making me consider trying to start using a typed JS dialect at some point.)

mihailik:
TypeScript CodeMirror? That would be so cool!

Besides, there seems to be another pasting bug still hiding, where newlines are converted to spaces on Android's Chrome Mobile. Hoping to catch it soon.

marijnh:
I'm playing around with [Flow](http://flowtype.org), which seems more modern and flexible, but it's also rather unfinished and rough at the moment, so I guess I'm not converting any projects over to it on the short term.

PR #3056, opened 2015-01-28
[tern addon] open [doc] link in a new tab/window

cpetrov:
Opening the documentation in the same window navigates away from the
code editor.


marijnh:
My assumption was that people would middle-click such a link anyway, but I guess on Mac that's more involved, and there's no real reason not to make a new tab the default. Merged as 893a12e

PR #3059, opened 2015-01-30
Making tooltip go away on scroll or blur

mihailik:
Easiest to see the difference on touch-enabled device. Go to Tern demo page (link below), click on a variable name and press Ctrl+I. Now pan the editor's content. The tooltip stays at a fixed position, whilst the variable name detaches and moves off.

The fix listens to scroll and blur, switching the tooltip off.

### DIFF
--- a/addon/tern/tern.js
+++ b/addon/tern/tern.js
@@ -595,6 +595,8 @@
       cm.state.ternTooltip = null;
       if (!tip.parentNode) return;
       cm.off("cursorActivity", clear);
+      cm.off('blur', clear);
+      cm.off('scroll', clear);
       fadeOut(tip);
     }
     var mouseOnTip = false, old = false;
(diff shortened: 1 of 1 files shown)


mihailik:
Not sure if cm.off(...) calls need to move above if (!tip.parentNode) check.

When would tip.parentNode be null? Would we still want to unsubscribe from the editor

marijnh:
Thanks. Merged as b002600. We're kind of re-solving problems that have already been solved for the `show-hint` addon, and are probably going to pop up in the `lint` addon as well at some point. I will try to take some time to factor out the common logic into a tooltip addon at some point.

nickasd:
I found that this issue still happens for argument hints. In the tern demo, just move the cursor to the arguments of `slice()` on line 6 and scroll down: the tooltip will not scroll and stay fixed.

adrianheine:
Since this is a closed pull request, could you just open a new issue and link to this PR?

PR #3061, opened 2015-01-31
Fixed missing quotation marks.

lukebrdn:


marijnh:
Omitting quotation marks is perfectly okay in HTML, and is intentionally done quite a lot in the files bundled with the CodeMirror codebase. As such, I don't feel it is helpful to start adding them.

PR #3063, opened 2015-02-02
Fixed an issue with invalid repeating regular expressions flags being highlighted as valid ones

giovannicalo:
I fixed an issue where invalid repeating regular expressions flags (e.g. `gg`, `mim`, `gmigy`) would be recognized and highlighted as they were valid.


marijnh:
See what happens when you enter `/foo/i + i`. Attached patch (a7d44a5) fixes that.

PR #3064, opened 2015-02-02
[css mode] teach it about CSS variables

KwanEsq:
Currently CSS variables e.g. --main-color not only fail to highlight properly, but also screw up the parsing later in the document, so auto-indenting fails. (It indents an extra step after using a variable in var(), and sometimes un-indents a step after declaring the variable).
This fixes both of these problems by correctly parsing occurrences of --variable-name

Tests run locally, all pass.

### DIFF
--- a/mode/css/css.js
+++ b/mode/css/css.js
@@ -57,6 +57,11 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
       if (/[\d.]/.test(stream.peek())) {
         stream.eatWhile(/[\w.%]/);
         return ret("number", "unit");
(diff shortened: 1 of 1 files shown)


marijnh:
Wonderful. Merged as 752c785f83f4, and test added in 941916482 (consider including a test in the pull req in the future).

KwanEsq:
Great, thanks!  Apologies for the lack of a test, I didn't immediately grok your test syntax, and figured if you wanted one you would let me know.  I'm attempting to write one for #3065, though not sure if I'm doing it correctly.

PR #3065, opened 2015-02-02
[css mode] teach it about @document

KwanEsq:
Add support for the @document rule to the CSS parser
Fixes part of https://bugzilla.mozilla.org/show_bug.cgi?id=977665

### DIFF
--- a/mode/css/css.js
+++ b/mode/css/css.js
@@ -66,7 +66,9 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
       return ret("qualifier", "qualifier");
     } else if (/[:;{}\[\]\(\)]/.test(ch)) {
       return ret(null, ch);
-    } else if (ch == "u" && stream.match("rl(")) {
+    } else if ((ch == "u" && stream.match(/rl(-prefix)?\(/)) ||
+               (ch == "d" && stream.match("omain(")) ||
+               (ch == "r" && stream.match("egexp("))) {
       stream.backUp(1);
       state.tokenize = tokenParenthesized;
       return ret("property", "word");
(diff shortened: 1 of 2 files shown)


marijnh:
@document appears to be a Gecko-specific non-standard feature, for now. I'm not too keen on adding support for it to the CSS mode, especially when it is relatively heavyweight (+29 lines).

Maybe we can come up with a solution where unknown @-directives are parsed in a simple, defensive way, matching brackets and, if the brackets are braces, treating the stuff between them like top-level directives? That'd be a little simpler, I believe, and involve less special-case code.

KwanEsq:
I've attempted to add some tests for this, but I'm not sure if I'm being over- or under-thorough.  They do at least all pass, though I had to wrestle with indenting errors for the first test for a while (hence making the smaller ones as possible replacements).

Re: your comment
Yes that's correct (I did mean to link to https://developer.mozilla.org/en-US/docs/Web/CSS/@document in my PR, but forgot)
That might be possible.  The original (dumb) way I did it when I was patching Firefox directly was to just treat @document (and @supports) as @media via changing
 ```javascript
} else if (type == "@media") {
```
to
```javascript
} else if (/@(media|supports|(-moz-)?document)/.test(type)) {
```
which got it most of the way.  The main reason for separating it was for the non-url functions like
```css
@document url-prefix(http://www.mozilla.org/) {
```
since it wouldn't parse their contents as strings.
Making just the tokenBase() changes (with document-as-media) seems to be enough to fix that, but the functions themselves are still tagged as errors.

KwanEsq:
Okay, I've tried merging @media, @supports, and @document into one in KwanEsq/CodeMirror@5561e21355892db8c3defb78f331be2c27d13e4c, and it seems to work as it passes all tests.
This isn't general for any future @rules, since I'm not sure how one would distinguish between one-line @rules, like @import and @namespace, and @ blocks like the three in question, or would you not bother?

marijnh:
Thanks. I've merged your alternative patch as 2d544462f25e. I was thinking to simply see whether the rule ended with a semicolon or a braced block to determine whether it supported a block or not, but I just realized that the fact that Less and SCSS also use `@` for variables is likely to create ambiguity, so for now let's simply go with a hard-coded list of @ directives.

KwanEsq:
Awesome, thanks a lot for the merge.

PR #3066, opened 2015-02-02
[css mode] teach it about @supports

KwanEsq:
Add support for the @supports rule
Fixes part of https://bugzilla.mozilla.org/show_bug.cgi?id=977665


marijnh:
Merged as 6b9511d11d2, test in 3469f317b5eb.

If we add another expression language (like the one in `@supports` and `@media`) we should generalize it so that we don't end up copy-pasting all the boilerplate stuff.

PR #3069, opened 2015-02-04
[tests] Tidy up test/index.html

KwanEsq:
Changes to tidy up test/index.html and make it cleaner.  Main change is moving all mode imports currently in div#testground into the header, and then alphabetising both the mode and test imports (with one exception).  This will hopefully avoid double imports in the future (as xml.js is now), and also makes the ordering less fragile, since some depend on multiple imports.

This work was mainly motivated by the desire to make future imports into Firefox easier.


marijnh:
Thanks. Merged as f3f57a3 (except for the empty alt attribute, which seemed useless).

KwanEsq:
Great, thanks a lot!  Looking forward to 4.13.

PR #3071, opened 2015-02-06
Check foldGutter for exists

mloginov:
To prevent uncaught exception if 'signalLater' send


marijnh:
Thanks for fixing that. Merged as e454aa8

PR #3076, opened 2015-02-09
Ability to destroy tern worker

mloginov:


marijnh:
Thanks. Merged as 011b947

PR #3077, opened 2015-02-11
[idl mode] Fix red brackets and commas

lstagner:
The red brackets and commas were getting annoying. This "fixes" the problem, although I am not sure this is the best way to fix it.

Before:
screenshot from 2015-02-10 16 48 01
 
After:
screenshot from 2015-02-10 16 47 19

### DIFF
--- a/mode/idl/idl.js
+++ b/mode/idl/idl.js
@@ -275,7 +275,7 @@
 
     // Handle non-detected items
     stream.next();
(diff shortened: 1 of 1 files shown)


marijnh:
Agreed -- valid syntactic elements shouldn't blindly be styled `error`. Merged as 9d9337f

PR #3080, opened 2015-02-13
Missed changes from the @document work

KwanEsq:
So I noticed today I missed a couple changes when doing the work for @document.  The most important is in the indent() function, where it referred to the now renamed media and media_parens states.
Also I've added documentTypes to the parserConfig, allWords hint array and CSS mime definition, but I'm not sure if that's needed.  I didn't add it to the LESS and SASS definitions because I'm not sure if that is necessary/they interact well.


marijnh:
Thank you for following up with those. Merged.

Issue #238, opened 2011-11-08
Copying is broken on windows in the keymap branch


marijnh:
Ctrl-a - ctrl-c - ctrl-v gives weird results.

PR #3083, opened 2015-02-14
[closebrackets addon] Option to override expansion of triple quotes #238...

ngn:
When the closebrackets addon is enabled, typing three quotes in a row causes six quotes to be inserted and puts the cursor in the middle.  Triple quotes make sense in Python and CoffeeScript, but not in the majority of languages.  I added an option to disable this behaviour, usage: `cm = CodeMirror(element, {autoCloseBrackets: {triples: ''}})`.  The default value for `triples` is `"'\""`.

### DIFF
--- a/addon/edit/closebrackets.js
+++ b/addon/edit/closebrackets.js
@@ -10,6 +10,7 @@
     mod(CodeMirror);
 })(function(CodeMirror) {
   var DEFAULT_BRACKETS = "()[]{}''\"\"";
+  var DEFAULT_TRIPLES = "'\"";
   var DEFAULT_EXPLODE_ON_ENTER = "[]{}";
   var SPACE_CHAR_REGEX = /\s/;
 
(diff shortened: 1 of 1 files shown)


marijnh:
What kind of situation requires putting three quotes in a row in a language that doesn't do triple-quoting?

ngn:
In some languages quotes inside strings are escaped by [doubling them up] (https://en.wikipedia.org/wiki/String_literal#Doubling_up).  For instance, to represent the JavaScript string `'\'a'` in APL, one would type `'''a'`.

marijnh:
That makes sense. Merged as 39409d6

anntzer:
Can this be made part of the individual language modes?  e.g. I tried adding `closeBrackets: {triples: ""}` to octave.js (which behaves similarly to APL), but this didn't seem to have any effect.

marijnh:
> Can this be made part of the individual language modes? 

That should already work. The Python mode has `closeBrackets: {triples: "'\""}`, for example.

anntzer:
Perhaps this doesn't really belong here (I am not sure) but I tried editing this out (setting triples to the empty string) of python.js as shipped by jupyter notebook (/usr/lib/python3.6/site-packages/notebook/static/components/codemirror/mode/python/python.js) and this had no effect.  Any suggestions?

marijnh:
What is the effect you're trying to achieve? The default value for `triples` _is_ the empty string, so setting it to that isn't expected to have any effect.

anntzer:
I am trying to make typing three quotes in a row NOT insert six quotes (well, the real goal is to fix octave.js which should have that setting  as it uses the same syntax as mentioned in https://github.com/codemirror/CodeMirror/pull/3083#issuecomment-74395375).

marijnh:
That had nothing to do with triple-quoting. Patch a5d7882c should fix it.

anntzer:
thanks!

PR #3084, opened 2015-02-15
[css mode] add support for @counter-style

KwanEsq:
The [CSS3 Counter Style spec](http://www.w3.org/TR/css-counter-styles-3/) has reached CR, and [Firefox has implemented it](https://hacks.mozilla.org/2015/02/introducing-counter-styles/).  Therefore I'd like to add it to the CSS mode.
[My first attempt](/KwanEsq/CodeMirror/tree/css-at-counter-style) just naïvely added the descriptors to the keywords, but I didn't like how this meant they were highlighted everywhere.  Fortunately I then realised/noticed you'd already implemented @font-face in an isolated way that'd be perfect for @counter-style.  Thus this second attempt, which I'm much happier with.
Main issues/questions:
Is the way I'm propagating what type of block it is in, via state.restrictedType, okay?  I couldn't see any pre-existing way to access that data, but maybe I just missed it.  And should it be named something else?  I could see the same approach being useful for distinguishing regular at-blocks in future, should the property be named the same for both?
Is the way I'm handling counter-style names in restricted_atBlock_before correct?


marijnh:
Looks good. Merged, and modified slightly in attached patch, with the intent to have a `stateArg` property dedicated to keeping local state, so that we don't end up adding more and more state properties for specific purposes. I've removed the `delete` use, since on some JS engines that still causes objects to become slow general dictionaries, which we don't want to happen for the `state` object.

PR #3085, opened 2015-02-16
[sql-hint] Better SQL table/column name completion, fixes #2323

### LINKED ISSUE #2323: SQL Show hint doesn't support schema name
For example I have table name like schemaname.tblName then show hint can't show column list with following code:

CodeMirror.commands.autocomplete = function(cm) {
    CodeMirror.showHint(cm, CodeMirror.hint.sql, { 
        tables: {
            "schemaname.table1": [ "col_A", "col_B", "col_C" ],
            "schemaname.table2": [ "other_columns1", "other_columns2" ]
        }
    } );
}

If I remove dbname form table name key, it works as expected.

nisargjhaveri:


marijnh:
Could you maybe add test cases? A big problem with the SQL completion is that I'm never sure what it's supposed to do, and whether a change I make breaks some of its functionality.

The idea would be to add an `sql-hint-test.js` file in `addon/hint`, which hooks into the simple testing framework defined in `test/driver.js`, and calls the completion function on some interesting inputs, verifying that it returns the right completions. Then include the new test script in `test/index.html`, and it will be run as part of the test suite.

nisargjhaveri:
Sure. I'll look into it.

nisargjhaveri:
@marijnh, I have added some tests for sql-hint. Please have a look.

Please check that I wrote the tests in right way, as I did not find any docs and wrote on basis of other tests and driver.

nisargjhaveri:
Sorry, I didn't follow the style guilds properly. Please review the tests. If it looks fine, I'll recommit this.

marijnh:
Great! I've cleaned them up a little (a1cdedd) and merged your work. (I had to move them to `test/` to prevent the automatic linter from going bonkers over them -- it is apparently not really suited for test code yet).

nisargjhaveri:
Thanks. :)

PR #3088, opened 2015-02-17
[css mode] add missing functions to valueKeywords

KwanEsq:
I've noticed that currently, only url() and the colour functional notations like rgba() are recognised by the parser, meaning that e.g. calc() is styled very differently from url().  This adds the rest of the CSS functions to the keywords array so that they are styled the same, and also updates three tests which include such functions.  I don't _think_ I've missed any (though obviously can't guarantee), and I've deliberately left out ones that seem too experimental/unimplemented (like element() and the two conic-gradient functions).

[attr, calc, counter, counters, linear-gradient, matrix, matrix3d, perspective, radial-gradient, repeating-linear-gradient, repeating-radial-gradient, rotate, rotate3d, rotateX, rotateY, rotateZ, scale, scale3d, scaleX, scaleY, scaleZ, skew, skewX, skewY, symbols, translate, translate3d, translateX, translateY, translateZ, var]


marijnh:
Looks good. Merged as 6a68467

Issue #3049, opened 2015-01-25
[vim] How can I define and map custom Vim commands?


kevinushey:
(this is a slight repeat of https://github.com/codemirror/CodeMirror/issues/2840)

I understand that I can define custom ex commands through `defineEx()`, and define key-to-key mappings through `map()`, but is there any way to map a particular (sequence of) keys to execute an arbitrary Javascript function?

Would it be sufficient to just append an appropriate object to the `defaultKeymap`? Or is there an alternative API available for this?

(It would be useful if some of the documentation re: extending Vim were placed here: http://codemirror.net/demo/vim.html)

Thanks! `vim.js` is really incredible :)

mightyguava:
> I understand that I can define custom ex commands through defineEx(), and define key-to-key mappings through map(), but is there any way to map a particular (sequence of) keys to execute an arbitrary Javascript function?

No, there is not. I'm open to adding one. Just never had a request (oddly), and have not had time to think of a good API
> 
> Would it be sufficient to just append an appropriate object to the defaultKeymap? Or is there an alternative API available for this?

Adding to defaultKeymap will not work unless the command is already defined in vim.js

kevinushey:
Some thoughts on how I imagine this might work:

1. Add an API for adding specific entries to the `operators`, `actions` and `motions` objects. I imagine this would be with e.g.

```js
addMotion("name", function(cm, head, motionArgs) { ... });
```

Or even:

```js
add("motion", "name", function(...) {});
```

Validating the function could be tricky; maybe it could be applied to some dummy text and see if it actually e.g. returns a `Cursor` object (in the case of a `motion`)?

After doing this, one could bind a (sequence of) keys to a particular action by either appending to the default keymap (probably not desired), or by asking the appropriate parts of `vim.js` to check both a `defaultKeymap` and a `userKeymap` for dispatching on a set of keys. Does that make sense?

If it does, I'd be happy to prototype and PR an implementation (time permitting)

mightyguava:
I think I like the `addMotion` option a bit better (would call it `defineMotion` to better align with CodeMirror). I'd remove `head` from the argument list though, it's a bit controversial whether it should have been added to the list of arguments to start with...

For binding to them, I'd have a
```js
mapCommand('keys', 'type', 'name', {motionArgs: {}});
```
and corresponding `unmapCommand` function.

Adding a `userKeymap` for storing the user-defined map is something I've wanted to add for a while now, but I don't consider it necessaary for this.

[...]

PR #3090, opened 2015-02-18
[vim] add API for defining custom commands (#3049)

kevinushey:
This PR adds a simple API allowing users of `vim.js` to add their own key mappings, through:

- `define*` commands allow users to map functions to particular 'names' for different commands within a particular group, and
- `mapCommand` allows users to map particular key combinations to these defined commands.

I also added `_mapCommand()` as the barest low-level API just so that users can also:

1. Define custom commands that might not fit well into the `mapCommand()` interface, e.g. complex `operatorMotion` commands, and
2. To allow that function to e.g. use a custom keymap in the future (rather than appending to the default keymap).

Comments appreciated!

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -776,7 +776,16 @@
       },
       handleEx: function(cm, input) {
         exCommandDispatcher.processCommand(cm, input);
-      }
+      },
+
+      defineMotion: defineMotion,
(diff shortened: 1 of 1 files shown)


mightyguava:
Merged as e7598d0, thanks!

PR #3094, opened 2015-02-18
honor indentWithTabs when handling "Tab" key

andrey012:


marijnh:
The current behavior is intentionally the way it is. If you want a different behavior, simply bind Tab to something else.

andrey012:
Marijn,

Sorry, I'll revert the change and will try to find another way without having to modify CodeMirror core.
I just thought, that there is certain disparency:
when I use Tab key without selection - it adds one \t char which is 4 spaces.
but when I use Tab key with selection - it adds 2 space chars.

http://software-developments.com/u/150220-1752-12-125097723101278069255506862254471279130238-0_codemirror_disparency.mp4

AG
Marijn Haverbeke wrote:

marijnh:
I agree there's a case to be made for the behavior you describe, but at this point I can't just change the behavior of a commonly used command out of the blue, since it would cause unexpected behavior for existing users.

PR #3095, opened 2015-02-19
Support advance properteis for sql hint

gzzhanghao:
Add support for advance properties (`displayText` for example) for sql-hint addon.

Here is an example for the new feature:
```
options.tables = [{
  text: 'table-name',
  displayText: 'table-name | comment',
  columns: [{
    text: 'column-name',
    displayText: 'column-name | column-type'
  }]
}]
```

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -26,9 +26,37 @@
     return CodeMirror.resolveMode(mode).keywords;
   }
 
+  function getText(item) {
+    if (typeof item == 'string') {
+      return item;
+    }
+    return item.text;
+  }
+
+  function getItem(list, item) {
+    if (Array.isArray(list)) {
+      for (var i = list.length - 1; i >= 0; i--) {
+        if (getText(list[i]) == item) {
+          return list[i];
+        }
+      }
+    }
+    return list[item];
+  }
+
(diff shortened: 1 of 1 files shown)


marijnh:
You are changing the table format in an incompatible, breaking way. That is not acceptable.

gzzhanghao:
@marijnh The patch is backward compatible, the original table format is supported as well.

marijnh:
Fair enough. Merged as 8f13aa615b8f7 and simplified and tested in 1eb72e8ccc17412.

addon/hint/sql-hint.js
```diff
@@ -26,9 +26,37 @@
     return CodeMirror.resolveMode(mode).keywords;
   }
 
+  function getText(item) {
+    if (typeof item == 'string') {
+      return item;
+    }
+    return item.text;
```
prathik457: Shouldn't this be item.displayText?
Let's suppose I have a table object {columns:{....}, text: '{{table1}}', displayText:'table1'}
Then when I type 'ta' it does not return me anything. Am I missing something?

marijnh: > Shouldn't this be item.displayText?

No, this is used to get the actual completion text, not the label.

PR #3097, opened 2015-02-21
[sql-hint addon] Fix defaultTable for new tables format

gzzhanghao:
Fix the `defaultTable` option when using the new tables format.


marijnh:
Thanks. Merged.

PR #3104, opened 2015-02-25
[lint addon] fix startLinting, onChange

coderaiser:
When option `lint` set to jshint config and `mode` set to `javascript`,  `file.js` is loaded - everything is OK.
But, if after that mode is changed to another one with no `lint` support (`gfm` for example) and set `lint` to `false` - after setting up value, errors are occur:
```js
Uncaught TypeError: Cannot read property 'options' of undefined in startLinting
Uncaught TypeError: Cannot read property 'timeout' of undefined in onChange
```

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -118,12 +118,15 @@
   }
 
   function startLinting(cm) {
-    var state = cm.state.lint, options = state.options;
-    var passOptions = options.options || options; // Support deprecated passing of `options` property in options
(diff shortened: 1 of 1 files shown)


marijnh:
Good point. Attached patch solves the problem in a slightly different way (making sure the timeout is canceled when the option is reset). Could you verify that this works for you?

coderaiser:
Everything works good, thank you. And it's definitely better to `clearTimeout` then just ignore it as I did.

PR #3111, opened 2015-03-02
[closebrackets addon] Add JS template strings (`)

### LINKED ISSUE #3110: Backquote is a string delimiter in JavaScript
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/template_strings

Presumably ``` should be added to https://github.com/codemirror/CodeMirror/blob/b4f8836c8f08324283b9cb9cb45714547c3f1f8d/addon/edit/closebrackets.js#L12

KwanEsq:
Fixes #3110

### DIFF
--- a/addon/edit/closebrackets.js
+++ b/addon/edit/closebrackets.js
@@ -9,7 +9,7 @@
   else // Plain browser env
     mod(CodeMirror);
 })(function(CodeMirror) {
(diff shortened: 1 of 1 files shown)


marijnh:
Patch e1cda70 implements something better -- the ability to configure closing behavior per-mode, so that modes where backticks shouldn't be closed don't close them.

Issue #3114, opened 2015-03-04
Backslash anywhere in a line is treated like a continuation character in properties mode


kevinearls:
It should only apply if it's the last character on a line

PR #3115, opened 2015-03-04
#3114 backslash anywhere in a line is treated like a continuation charac...

kevinearls:
...ter in properties mode


marijnh:
Thanks! Merged as 4485ae7

PR #3116, opened 2015-03-05
Fix typo in SASS mode that was introduced by PR #3013

peterflynn:
Typo was in commit 9e56052a

### DIFF
--- a/mode/sass/sass.js
+++ b/mode/sass/sass.js
@@ -232,7 +232,7 @@ CodeMirror.defineMode("sass", function(config) {
 
       if (stream.eatWhile(/[\w-]/)){
         if(stream.match(/ *: *[\w-\+\$#!\("']/,false)){
-          return "propery";
+          return "property";
         }
         else if(stream.match(/ *:/,false)){
           indent(state);


peterflynn:
Btw, the token types used in SASS are really inconsistent with the set used by the CSS/LESS/SCSS modes.

| | CSS/LESS/SCSS | SASS |
|---|---|---|
| Tag name selector | `tag` | `atom` |
| Class selector | `qualifier` | `atom` |
| Id selector | `builtin` | `atom` |
| Property name | `property` | `property` |
| Numeric value | `number` (including any unit suffix) | `number` (but unit suffix is `unit`) |
| Enumerated value | `atom` | `attribute` |

Would there be any objections to changing SASS to match the others more closely?  And maybe porting the `unit` thing into the others, since that seems like a nice enhancement...  (CC @ankitstarski since you seem to have worked on the SASS code most recently)

marijnh:
I've merged your patch as dad7abf . I'd be okay with changing the styles output by the SASS mode to match the CSS dialects more closely, and also with copying the trick of giving units a separate style -- though I'd prefer to use one of the standard styles, rather than `unit`, which is not covered by any of the themes.

PR #3118, opened 2015-03-06
[vim] Fix a bug in linewise visual mode 

heppe:
ex commands (such as :s) operating on linewise block would do nothing if the block happened to end on an empty line.


mightyguava:
The check was a red-herring. `handleExternalSelection` should not have been triggered in this case. Fixed in 11fa758

mightyguava:
Unfortunately, because the behavior that was broken is async, I couldn't find a good way to test it

PR #3124, opened 2015-03-10
When using async mode, if there are multiple calls to show the widget (s...

adammartin1981:
...taggered data loading from multiple sources) then close the widget prior to adding data.


adammartin1981:
If there is a better way that I've missed to circumvent this issue then feel free to tell me, I was looking for some kind of Widget.updateData method but couldn't see one.

adammartin1981:
Didn't test fully whilst typing - please ignore

PR #3125, opened 2015-03-13
new mode: cmake

darealshinji:
For *.cmake and CMakeLists.txt files. I made this based on the puppet mode code.


marijnh:
Wonderful. Merged as fd9f66c3fb1970. And I've reduced the size of the example code in 03a49e971 (since it counts towards library download size, until I get around to splitting the docs from the library package)

darealshinji:
Yeah, the example was really a bit too long.

PR #3126, opened 2015-03-13
new mode: troff (for manpages)

darealshinji:


marijnh:
Looks great. Merged as 30a118a (and wow, I didn't troff was such an ugly format)

PR #3127, opened 2015-03-13
[sublime] Sublime-style smart backspace

vincentwoo:
In Sublime, when you hit backspace, the editor checks to see if everything proceeding your cursor is whitespace. If it is, and your cursor is on an even multiple of the tab size, it will simply delete one full tab, which is pretty nice.

I've tried to mimic this here. It's a bit hacky without some defined API for figuring out the real length of hard vs soft tabs but I think this suffices. Thoughts?

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -409,6 +409,22 @@
 
   map[cK + ctrl + "Backspace"] = "delLineLeft";
 
+  cmds[map["Backspace"] = "smartBackspace"] = function(cm) {
+    if (cm.somethingSelected()) {
+      return CodeMirror.Pass;
+    }
+
+    var tabSize = cm.getOption('tabSize');
+    var cursor = cm.getCursor();
+    var toStartOfLine = cm.getRange({line: cursor.line, ch: 0}, cursor);
+    var column = CodeMirror.countColumn(toStartOfLine, null, tabSize);
+
+    if (/^\s+$/.test(toStartOfLine) && column % tabSize == 0) {
+      return cm.indentSelection('subtract');
+    }
+    return CodeMirror.Pass;
+  };
+
   cmds[map[cK + ctrl + "K"] = "delLineRight"] = function(cm) {
     cm.operation(function() {
       var ranges = cm.listSelections();


keymap/sublime.js
```diff
@@ -409,6 +409,24 @@
...
+  map["Backspace"] = function(cm) {
+    if (cm.somethingSelected()) {
+      return CodeMirror.Pass;
+    }
+
+    var tabSize = cm.getOption('tabSize');
+    var spaces = new Array(tabSize).join(' ');
+    var cursor = cm.getCursor();
+    var toStartOfLine = cm.getRange({line: cursor.line, ch: 0}, cursor)
+      .replace("\t", spaces);
```
marijnh: You should be able to use `CodeMirror.countColumn(cm.getLine(cursor.line), cursor.ch, tabSize)` to get the column at which the cursor is. That's probably cleaner.

vincentwoo:
Okay, I took your advice on `countColumn`. I also registered the function in the commands map, which I hope you don't mind. I think it's a useful one.

marijnh:
Thanks. Merged as 8dbb84f44d3f and tweaked in 33e56fbd728ad0. I figured `indentUnit` is a more appropriate value to check the modulo of, since that's what `indentSelection` will actually shift the line by, and what most modes use as the base unit of indentation.

vincentwoo:
Thanks! Out of curiosity, can you explain how/when `indentUnit` and `tabSize` differ?

marijnh:
They should usually be set to the same value, but `indentUnit` is the unit of indentation, and tabsize is the width of a tab. Those are not necessarily the same.

marijnh:
I've reverted this in 55a47a5, because

 * It breaks when there are multiple cursors, which is something the rest of the sublime bindings take care to handle

 * I couldn't get sublime text to actually show this behavior by default. If I press backspace after leading spaces, even when on a tab-count boundary, it still simply deletes the space before the cursor

See #4023

vincentwoo:
My colleagues and I see this default backspace behavior on sublime 2 and 3, with no customizations to the keybindings. It might be relevant that we all use OSX. Would you be okay with merging a version of this that works correctly with multiple cursors? I can take a crack at it.

marijnh:
Sure, I'd merge a proper implementation.

PR #3131, opened 2015-03-14
Repurpose verilog extension mode as transaction level verilog (suffix: ....

ratchup:
...tlv) mode. The new mode escape-to token prefix is  "\TLV".

### DIFF
--- a/mode/verilog/verilog.js
+++ b/mode/verilog/verilog.js
@@ -375,73 +375,73 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
     name: "verilog"
   });
 
-  // SVXVerilog mode
+  // TLVVerilog mode
 
-  var svxchScopePrefixes = {
+  var tlvchScopePrefixes = {
     ">": "property", "->": "property", "-": "hr", "|": "link", "?$": "qualifier", "?*": "qualifier",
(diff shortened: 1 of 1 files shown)


marijnh:
Did the targeted language change its name? Again, are you sure this language is mature and stable enough that it makes sense to keep a mode for it in the CodeMirror core library?

ratchup:
Yes, the language is now called transaction level verilog (TLV). I think of
this change as connotative of a language scope reduction.  In regard to the
questions about the stability and maturity of the language, I will ask the
language architect about these topics. I know that Steve has additional
plans to enhance the design flow supported by TLV, but I'm not sure how (or
if) these plans will impact the  CodeMirror support.

ratchup:
I received feedback from Steve: It's desirable for TLV mode to be part of
the CM release. (Reduces maintenance, distribution, and installation
issues.)  But this may be a burden for the CM maintainer.  To be fair and
honest, the expectation is that the language is evolving (as all languages
are) and that  periodic update requests will be submitted. (The expectation
is that these updates will not happen  on less than a monthly interval.)
The name change has to do with the language being adopted by the industry.
The language is currently being deployed at sandpiper.edaplayground.com.
Support in CM will be a draw for CM during the anticipated rapid adoption.

marijnh:
Fair enough. Merged as 78a35ed

PR #3148, opened 2015-03-23
More options for panel positioning.

ju1ius:
This PR allows more fine-grained control of panels postion.
See the updated [demo](https://github.com/ju1ius/CodeMirror/blob/ju1ius/panels/demo/panel.html) and [manual](https://github.com/ju1ius/CodeMirror/blob/ju1ius/panels/doc/manual.html#L2762) for details.

### DIFF
--- a/addon/display/panel.js
+++ b/addon/display/panel.js
@@ -19,19 +19,19 @@
         wrapper = info.wrapper,
         cmWrapper = this.getWrapperElement()
     ;
-    if (options.position === "bottom") {
-      wrapper.appendChild(node);
-    } else if (options.position === "before-bottom") {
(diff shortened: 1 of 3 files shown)


marijnh:
That looks useful. Merged as 79e162b51c1, and tweaked in e0cf6f8d3854. If you are contributing to a project it is generally a good idea to look at the code style it uses, and conform to that.

ju1ius:
> If you are contributing to a project it is generally a good idea to look at the code style it uses, and conform to that.

Sorry for that. I did read CONTRIBUTING.md, but the "double-equal", "double-quotes", "multiple-vars" et al. weren't mentioned. Maybe the contributing notes should be a little more specific about this ?

Thanks for your time !

PR #3150, opened 2015-03-24
[lint addon] add support for jshint globals option

mgh:
`options` with a `globals` key as defined in a standard `.jshintrc` is actually separated into two arguments to the `JSHINT` function in the `jshint` cli: `https://github.com/gameclosure/jshint/blob/master/src/cli.js#L487`.  This forwards `options.globals` properly to `JSHINT`.


marijnh:
Thanks! Merged as 8b5e659

PR #3153, opened 2015-03-26
Add 'origin' parameter to searchCursor's 'replace' method

danieleds:


marijnh:
Thanks! Merged as 92d0d3d

PR #3155, opened 2015-03-27
Fix variable color in the monokai and solarized themes.

kreeger-google:
I've noticed that the colors for the Monokai and the Solarized theme are wrong. After inspecting around, I've noticed that any non-keyword gets marked as 'cm-variable'. This makes looking at files like Java and C++ (which we do a lot at Google) very hard with theses themes.

Here is the before:
Solarized Dark: https://drive.google.com/file/d/0B_n2OUyge9Hac185Y3JtOWM3WG8/view?usp=sharing
Monokai: https://drive.google.com/file/d/0B_n2OUyge9HaTkNTaXVvWnJmNTQ/view?usp=sharing

Solarized Dark: https://drive.google.com/file/d/0B_n2OUyge9Hacjc5bmpfZmFFMUU/view?usp=sharing
Monokai: https://drive.google.com/file/d/0B_n2OUyge9HaTkNTaXVvWnJmNTQ/view?usp=sharing

### DIFF
--- a/theme/monokai.css
+++ b/theme/monokai.css
@@ -18,7 +18,7 @@
 .cm-s-monokai span.cm-keyword {color: #f92672;}
 .cm-s-monokai span.cm-string {color: #e6db74;}
 
-.cm-s-monokai span.cm-variable {color: #a6e22e;}
+.cm-s-monokai span.cm-variable {color: #f8f8f2;}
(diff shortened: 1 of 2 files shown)


marijnh:
Good idea. Merged as ba90b71

PR #3165, opened 2015-03-30
Fix for StringStream.match() when a RegExp starts with \b

cassareh:
Here's a fix for a small bug with highlighting search matches.  If a user wants to perform a whole word search they will typically wrap their pattern with a leading and trailing \b.  (eg. /\btext\b/)  Currently the search highlighter will highlight additional matches which are not whole word matches.  This patch is a fix for this issue.

Thanks!

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -5857,7 +5857,7 @@
           // If the regex starts with \b than we need to consider the previous
           // character as the pattern implies that any previous
           // character should be a non-word character to satisfy the regex.
-          // However, when parsing the string as a stream, every search will 
+          // However, when parsing the string as a stream, every search will
           // match a leading \b pattern because each successive search moves the
           // start of line forward and ignores the previous characters. The
           // following check will honor the leading \b in the pattern.
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -5852,6 +5852,20 @@
       } else {
         var match = this.string.slice(this.pos).match(pattern);
         if (match && match.index > 0) return null;
+
+        if (pattern.source.indexOf('\\b') == 0) {
+          // If the regex starts with \b than we need to consider the previous
+          // character as the pattern implies that any previous
+          // character should be a non-word character to satisfy the regex.
+          // However, when parsing the string as a stream, every search will 
+          // match a leading \b pattern because each successive search moves the
+          // start of line forward and ignores the previous characters. The
+          // following check will honor the leading \b in the pattern.
+          if (this.pos > 1 && /\w/.test(this.string.charAt(this.pos-1))) {
+              return;
+          }
+        }
+
         if (match && consume !== false) this.pos += match[0].length;
         return match;
       }


marijnh:
You are required to start the regexps you pass to `match` with `^`, so adding a `\b` at the start is not meaningful. 

(This patch also doesn't really solve the issue -- you'll still go wrong for `/(\bfoo)/` for example.)

cassareh:
Yes good point, it is a poor fix as your example illustrates, however that aside, do you see an issue with `match` as it is implemented today or am I thinking about it incorrectly?  As I currently understand the `match` function, it would still perform incorrectly in some cases even if the regexp starts with `^`.  

Consider this *simplified* example [`mode`](http://codemirror.net/doc/manual.html#token) object with `token` function defined as follows.  The intended behavior is to find all matches on this line that satisfies the regexp.

``` javascript
var modeObject = ({
  token: function(stream) {
    var pattern = /^test/;
    while(!stream.eol()) {
      if (stream.match(pattern, /* consume */ true)) {
        return 'my-style';
      }
      stream.next();
    }
  }
});
```

If pattern is set to `/^test/` (what should be valid) and the underlying string in `stream` were `"test test test"`, than this function would find three matches where only one is expected. `match` would find three  matches because it does a pattern match from `stream.pos` forward.  As the `stream` object is incremented, each call to `match` is handled as if it is the start of line and this behavior is what causes a regexp that starts with `^` to find additional matches. I am hitting this exact issue but with `\b`.  Can you help clarify, am I using `match` incorrectly or is this an issue with the `match` function?

Thanks for your time!

marijnh:
The required `^` in the regexp is only there because of a limitation in JS' regexp interface (we can't force a match to start at a given position in a string, so we have to call `slice` on the string and match that, pinning the start of the match to the start of the substring). Thus, you can't make your regexp depend on anything that came before the current stream position, and a `\b` at the start of the regexp is meaningless (always matches).

cassareh:
Thanks, I follow your reasoning as I have also come across this limitation in javascript.  However I don't understand how a leading `^` in the regexp is any different from the leading `\b`.  As far as I understand it, they are both meaningless as they will both always match.  It seems to me that a regexp should not be allowed to start with a `[^\b\B]` characters when using the `StringStream.match()` function.

That aside, I found a workaround for this limitation which could be used to make the `StringStream.match()` more general.  The [String.prototype.replace()] (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter) function allows passing a function instead of replace text.  It is possible to piece this together such that you can use `str.replace()` to find the offsets of all matches on a given line.  Here is a quick example:

``` javascript
function finder(match) { 
  console.log( 'Found match:' + match + ' at offset:' + arguments[arguments.length-2]);
  return match;
}

"hellohellohello hello Hello".replace(/\bhello\b/gi, finder)
```
```
2015-04-06 10:51:58.375 Found match:hello at offset:16
2015-04-06 10:51:58.375 Found match:Hello at offset:22
"hellohellohello hello Hello"
```

I am proposing that I can build a function which will use `str.replace()` to find the offsets of all matches on a line.  `StringStream.match()` can then use this function and its current position in the stream to determine if the current position is a match.  What do you think about this solution?  Should I continue forward with this?

marijnh:
The `index` property on the return value from `match` and `exec` already allows us to trivially find the position where a match is located. The reason we need `^` is to prevent the regexp engine from doing a lot of extra work. Until the `/y` regexp flag from ES6 is widely supported, we'll have to stick with the current kludge.

cassareh:
Yeah, unfortunately as you know, `index` doesn't help in the case of a leading `\b` in the regexp because it only returns the first match.  To use to find all matches on a line would require parsing the string as a stream which has the issues we already discussed, hence the work around I proposed above.

Having said that, I agree with you that the `/y` flag is the only clean solution and I can understand the reasoning behind waiting until it is available instead of introducing hacks or workarounds.  

Also I finally understand the reasoning behind the leading `^`, thanks for explaining it.  It was confusing to me since the leading `^` doesn't affect the matches found as it normally would in a regexp. Instead the leading `^` is used only as a performance optimization, which wasn't clear to me until now.

Thanks again for taking the time!

PR #3176, opened 2015-04-02
Create mumps.css

LloydMilligan:
Alternate syntax coloring for MUMPS language


marijnh:
This time it worked! But as I said in the other discussion, this is too specialized for the distribution.

LloydMilligan:
I agree Marijn.  After uploading I realized the only reason for changing the color scheme was to make the editor look more like an accustomed one, which is pointless.

Again, thanks for  your time.

Best wishes,
Lloyd

PR #3185, opened 2015-04-07
Basic Handlebars syntax highlight

libo:
Rendering of HBS template.

screen shot 2015-04-07 at 16 18 57

### DIFF
--- a/mode/handlebars/handlebars.js
+++ b/mode/handlebars/handlebars.js
@@ -1,0 +1,53 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"));
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror", "../../addon/mode/simple"], mod);
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
+  "use strict";
+
+  CodeMirror.defineSimpleMode("hbs", {
+    start: [
+      { regex: /\{\{!--/, push: "dash_comment", token: "comment" },
+      { regex: /\{\{!/,   push: "comment", token: "comment" },
+      { regex: /\{\{/,    push: "handlebars", token: "tag" }
+    ],
+    handlebars: [
+      { regex: /\}\}/, pop: true, token: "tag" },
+
+      // Double and single quotes
+      { regex: /"(?:[^\\]|\\.)*?"/, token: "string" },
+      { regex: /'(?:[^\\]|\\.)*?'/, token: "string" },
+
+      // Handlebars keywords
+      { regex: />|[#\/]([A-Za-z_]\w*)/, token: "keyword" },
+      { regex: /(?:else|this)\b/, token: "keyword" },
+
+      // Numeral
+      { regex: /\d+/i, token: "number" },
+
+      // Atoms like = and .
+      { regex: /=|~|@|true|false/, token: "atom" },
+
+      // Paths
+      { regex: /(?:\.\.\/)*(?:[A-Za-z_][\w\.]*)+/, token: "variable-2" }
+    ],
+    dash_comment: [
+      { regex: /--\}\}/, pop: true, token: "comment" },
+
+      // Commented code
+      { regex: /./, token: "comment"}
+    ],
+    comment: [
+      { regex: /\}\}/, pop: true, token: "comment" },
+      { regex: /./, token: "comment" }
+    ]
+  });
+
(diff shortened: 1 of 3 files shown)


marijnh:
I don't do early feedback on half-finished code, but if you have something that you feel is finished, I'll review it.

libo:
I think we reach a mature support for the Handlebars language. 
Please feel free to review.

mode/handlebars/handlebars.js
```diff
@@ -0,0 +1,53 @@
...
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror", "../../addon/mode/simple"], mod);
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
+  "use strict";
+
+  CodeMirror.defineSimpleMode("hbs", {
+    start: [
+      { regex: /\{\{!--/, push: "multiline_comment" },
```
marijnh: Doesn't it look better when you also style comment-delimiting braces with the `"comment"` style?

mode/handlebars/handlebars.js
```diff
@@ -0,0 +1,53 @@
...
+      { regex: /\{\{!/,   push: "comment" },
+      { regex: /\{\{/,    push: "handlebars" }
```
marijnh: Similarly, I'd also apply some token style to these (and the corresponding closing braces). Maybe `"bracket"` is appropriate, or `"tag"`.

mode/handlebars/handlebars.js
```diff
@@ -0,0 +1,53 @@
...
+    handlebars: [
+      { regex: /\}\}/, pop: true },
+
+      // Double and single quotes
+      { regex: /"(?:[^\\]|\\.)*?"/, token: "string" },
+      { regex: /'(?:[^\\]|\\.)*?'/, token: "string" },
+
+      // Handlebars keywords
+      { regex: />|[#\/]([A-Za-z_]\w*)/, token: "keyword" },
+      { regex: /else|this/, token: "keyword" },
```
marijnh: You'll probably want to use something like `/(?:else|this)\b/`, so that the start of 'elsee' does not match 'else'.

mode/handlebars/handlebars.js
```diff
@@ -0,0 +1,53 @@
...
+      // Numeral
+      { regex: /\d+/i, token: "number"},
+
+      // Atoms like = and .
+      { regex: /=|~|@|true|false/, token: "atom"},
+
+      // Paths
+      { regex: /(?:\.\.\/)*(?:[A-Za-z_][\w\.]*)+/, token: "variable-2"}
+    ],
+    multiline_comment: [
```
marijnh: Since all Handlebars comments are multiline, maybe 'dash_comment' or so is a better name.

marijnh:
I've added a few notes. Are you combining this mode with a content (HTML or so) mode using `addon/mode/overlay`, or aren't you?

libo:
@marijnh thanks a lot for feedback!
I have applied all of them.

To answer your question, yes I believe that the best case scenario is to go in overlay with XML or so.
Shall we add it to the HTML demo? I felt like it was better to leave it isolated...

marijnh:
Wonderful. Merged as f7a00e0. I was just curious whether a simple mode would work with an overlay, but I guess there's no reason why it wouldn't.

libo:
Awesome thanks!

Overlay works and I am using it on a project, there is a small glitch with stuff like 
```html
<img src="{{ user.avatar.url size="small" }}">
```

That actually is valid Handlebars+HTML but the code linter report as error since the double quotes confuse it. 

The issue is that the XML lexer swallow the entire content between `" "` and hence doesn't detect the opening of Handlebars mode's token "{". 

@ilkkao and me ended up patching the xml.js in our project to handle this, I was wondering if we could think of more flexible way to handle overlay in case like these or I misunderstanding something.

ilkkao:
Like Libo said, XML mode treats quoted string as a single token and eats it. Could there be a "escape sequence'" option in XML mode (in this case '{{' that would stop eating the input and return a token immediately. That way the top level mode would get a chance to switch to an another submode.

marijnh:
Using the [`multiplex`](http://codemirror.net/doc/manual.html#addon_multiplex) addon instead of overlay should help here (the main difference being that the outer mode does not get to see the content inside of the inner mode, but before now, that would not feed the delimiters (`{{` and `}}`) to the inner mode, which would prevent the handlebars mode from highlighting any handlebars code. In patch 9c524fd, I've added an option `parseDelimiters` that changes this, and I've updated the handlebars demo to use a multiplexing mode.

**Also note**: I've changed the mode name of the handlebars mode to `handlebars', since all CodeMirror modes define a mode with the same name as the file/directory they live in, and it'd confuse people (as well as the `loadmode` addon) if that no longer holds.

libo:
Thanks a ton @marijnh ! This is awesome!

Is this going to be released in 5.1? How often do you release CodeMirror btw?

marijnh:
Yes, and I release once a month, around the 20th (so that'll be early next week this month)

PR #3190, opened 2015-04-09
Direction

jasonzhao3:
I manually tested this branch. The cursor is significantly improved. It works fine with the mouse cursor now. Thanks! 

But I find it still has issue with the rtl space. For example, if you type something شسيب then type space, the cursor doesn't move.  Can we merge this branch first and then keep working on it?

### DIFF
--- a/demo/bidi.html
+++ b/demo/bidi.html
@@ -9,6 +9,7 @@
 <script src="../mode/xml/xml.js"></script>
 <style type="text/css">
       .CodeMirror {border-top: 1px solid black; border-bottom: 1px solid black;}
+      .CodeMirror pre > * { text-indent: 0px; }
     </style>
 <div id=nav>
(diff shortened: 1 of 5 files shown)


jasonzhao3:
Also I found if you use arrow key to navigate through words mix of ltr and rtl characters, you will get quite weird behavior -- different from what I tried in a  rtl textarea.

marijnh:
This branch was mostly written by last year's Summer of Code student, but he never really finished it properly or added tests, and I don't want to merge it in the state it is currently in. This is extremely tricky code, so finishing and merging it would take some time and energy. I have been too busy to take this time. If you are part of an organization that can allocate some money for this, that would help, I guess. If not, I guess I'll eventually feel bad enough about this to work on it, but that might take a while.

Flimm:
BTW, for anybody stumbling upon this issue in Google, it seems that CodeMirror does now support bi-directional text:

https://codemirror.net/demo/bidi.html

PR #3192, opened 2015-04-10
[clike mode] Remove C 'entry' keyword

mtaran-google:
It was never standardized and is no longer a reserved word: http://publications.gbdirect.co.uk/c_book/chapter2/keywords_and_identifiers.html


marijnh:
Thanks for noticing that. Merged as 5f44dd5

PR #3193, opened 2015-04-10
Add ASN.1, TTCN/CFG and TTCN modes

oswcab:
Tests are passing OK.

### DIFF
--- a/mode/asn.1/asn.js
+++ b/mode/asn.1/asn.js
@@ -1,0 +1,244 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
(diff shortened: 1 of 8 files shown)


marijnh:
Please make sure the linter also passes (install node, `npm install`, run `bin/lint`).

And remove boilerplate code that isn't actually needed. I notice the CFG mode has an `indent` function that A) seems useless, since the config format doesn't seem to involve indentation, and B) doesn't work (crashes with "ctx.indented is not a function"). That looks like you simply copy-pasted the code and didn't think about it very much. Similarly, since you're only defining a single mime type, the `def` function is a waste of space in that file.

Finally, please rename the `cfg` mode to `ttcn-cfg`, since it is too obscure to take such a short name.

marijnh:
Superseded by #3213

PR #3194, opened 2015-04-10
[python mode] better highlighting for and/or/not

mtaran-google:
Currently and/or/not in most themes look like normal text because cm-operator doesn't usually get special styling. This patch highlights them as keywords to make them stand out better.


marijnh:
Sounds good. Merged as 2e16ef6

PR #3195, opened 2015-04-11
[z80 Mode] Improve z80 mode, repair example, add ez80 mode.

KermMartian:
These changes were implemented for an online graphing calculator IDE/program editor, SourceCoder. It recently gained ASM support, so I took some time to fix problems with the existing z80 mode and add support for ez80 assembly. I also repaired stylistic and functional problems in the z80 demo's ASM code.


marijnh:
Thanks! Merged as 5c90705

PR #3196, opened 2015-04-11
Add support for configuring language mode via vim 'filetype' setting

dbarnett:
Allows using the vim `:set` command to query and override the language mode:
```vim
:set filetype?
:set ft?
:set filetype=clike
:set filetype=
```

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -4151,16 +4151,37 @@
           // Calling set with a boolean option sets it to true.
           value = true;
         }
-        if (!optionIsBoolean && !value || forceGet) {
-          var oldValue = getOption(optionName);
-          // If no value is provided, then we assume this is a get.
+        // 'ft' is an alias for 'filetype'.
+        if (optionName == 'ft') {
+          optionName = 'filetype';
+        }
+        // If no value is provided, then we assume this is a get.
+        if (!optionIsBoolean && value === undefined || forceGet) {
+          var oldValue;
+          if (optionName == 'filetype') {
+            // The 'filetype' option proxies to the CodeMirror 'mode' option.
+            oldValue = cm.getMode().name;
+            if (oldValue == 'null') {
+              oldValue = '';
+            }
+          } else {
+            oldValue = getOption(optionName);
+          }
           if (oldValue === true || oldValue === false) {
             showConfirm(cm, ' ' + (oldValue ? '' : 'no') + optionName);
           } else {
             showConfirm(cm, '  ' + optionName + '=' + oldValue);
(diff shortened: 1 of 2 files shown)


dbarnett:
Caveat: Changing the mode doesn't seem to totally work. I tried setting to clike in the vim demo and it had some coloring and indentation behavior that was not quite clike. But that could be fixed in setOption, which already tries to handle 'mode' specially, and even just the querying is something I've been wanting for a while.

Some parts like the alias handling could be generalized. Could do that as follow-up or on the second setting that needs it.

dbarnett:
Oh, the mode quirks might have been specific to how 'clike' is configured on the demo/vim.html page. I tried hacking some other pages and it seemed to work fine.

mightyguava:
I'd rather `set` didn't special case `filetype`. It opens up the path for a lot of clutter in `set`. Can you use the new callback support in ddbf33f and use `defineOption` to build this behavior instead? Aliasing `ft` to `filetype` in `set` is fine to me though.

dbarnett:
Awesome, agree that will be much better. One thing, though: What's the best way to get access to the `cm` object from this callback?

mightyguava:
Ah I didn't think that far. The options API here doesn't actually support local options. I've expanded it in https://github.com/codemirror/CodeMirror/pull/3200. Please take a look and see if it fits your requirements

dbarnett:
Updated to use option callbacks and refactored alias registration. PTAL.

I did change the arg order of `Vim.defineOption` to add `aliases` before `callback` and keep the callback last. Shouldn't affect any callers since the callback arg was just added, but I could just as easily make it a separate `Vim.defineOptionAlias` func or whatever makes sense to you.

mightyguava:
Thanks. Merged as 346d16c and added 2c45249

dbarnett:
> Caveat: Changing the mode doesn't seem to totally work. I tried setting to clike in the vim demo and it had some coloring and indentation behavior that was not quite clike.

I figured out what was going on with the clike mode quirks here. You have to set the mode to "text/x-csrc" instead of "clike" for it to load the right options. So `setOption('mode', getOption('mode'))` loses information, which is unfortunate.

marijnh:
@dbarnett At the CodeMirror level, doing `instance.setOption("mode", instance.getOption("mode"))` should not lose any information (the original value is preserved at the mode option's value). Is that what you were talking about?

dbarnett:
Aha, I had a small discrepancy getting the 'mode' option in a different way than I was setting it. Fix in #3218.

PR #3197, opened 2015-04-11
Radically improve Django mode

parisk:
This pull request improves radically the Django mode by introducing more features and fixing some bugs.

In detail, this pull request introduces the following

- Comment highlighting (both with inline tag and block comment tag)
- String highlighting
- Variable and variable attribute highlighting
- Operator highlighting
- Highlighting for filters
- Definition of `blockCommentStart` and `blockCommentEnd`
- Appropriate highlighting according to tag type: (`variable`, `tag`, `comment`)

This pull request also fixes two major bugs that made highlighting break:

- Including keywords inside strings highlighted them as keywords instead of string
- Mixing Django template highlighting with HTML highlighting in the same line broke the HTML highlighting

The only oddball in this pull request is the `"null"` token styling. Since the `django` mode is an overlay of `django:inner` over `html`, when using Django templates in HTML tag attributes, returning `null` results in string styling (because of the HTML tag attribute), instead of no styling. For this reason I returned `"null"` to avoid this.

The Django template language is described extensively at https://docs.djangoproject.com/en/1.8/ref/templates/language

This pull request can be tested at https://codemirror-paris.apps.lair.io/mode/django/index.html

### DIFF
--- a/mode/django/django.js
+++ b/mode/django/django.js
@@ -18,7 +18,10 @@
                     "loop", "none", "self", "super", "if", "endif", "as", "not", "and",
                     "else", "import", "with", "endwith", "without", "context", "ifequal", "endifequal",
                     "ifnotequal", "endifnotequal", "extends", "include", "load", "length", "comment",
-                    "endcomment", "empty"];
+                    "endcomment", "empty", "url", "static", "trans", "blocktrans", "now", "regroup",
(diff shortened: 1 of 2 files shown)


marijnh:
Looks great! Merged as e22891b

PR #3198, opened 2015-04-11
Exit vim ex/search on last backspace and clear input with Ctrl-U

dbarnett:
Makes backspacing over the leading `:` or `/` exit ex/search modes just like in real vim.

Also implements `c_CTRL-U` to remove all input characters in ex/search prompts (normally this removes "all characters between the cursor position and the beginning of the line", but I didn't see a way to detect the cursor position implemented yet).


mightyguava:
Merged as a405048

PR #3201, opened 2015-04-13
[jsp mode] Add

noragrossman:


marijnh:
There's quite a lot of tricky aspects involved with doing your own mode-switching like this. Any reason you didn't use a multiplexing mode (as in the [multiplex](http://codemirror.net/doc/manual.html#addon_multiplex)?

noragrossman:
No, I should have, and researching multiplexing mode actually revealed that there is actually already support for JSP within the HtmlEmbedded mode, it was just not obvious at first. I am closing the pull request!

PR #3202, opened 2015-04-13
[mobile] disable autocomplete to ensure key event support

mtaran-google:
See https://code.google.com/p/chromium/issues/detail?id=118639#c209

You may have a better place to put this, or want to gate it on being Chrome, etc.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1146,6 +1146,7 @@
   }
 
   function disableBrowserMagic(field) {
+    field.setAttribute("autocomplete", "off");
     field.setAttribute("autocorrect", "off");
     field.setAttribute("autocapitalize", "off");
     field.setAttribute("spellcheck", "false");


marijnh:
Could you elaborate on the problem that is being solved by this? In principle, the editor should be able to handle compose events.

mtaran-google:
I believe the original issue was that Chrome on Android would send the backspace key event but would not fill in the key code properly. It'd probably be easy to test it out when the next Android Chrome beta comes out, since I'm not sure if this will actually be an issue.

marijnh:
Turning off autocomplete seems like a very indirect and complication-prone way to address this. So yeah, I'd prefer to test first and only go ahead if the problem persists.

marijnh:
Closing this due to lack of response, feel free to comment/reopen if new information is available.

PR #3204, opened 2015-04-14
Preserving the last search query.

amshali:


marijnh:
Good idea. Merged as 82e86ef74a55, fixed in d6318cbbb8a0597395

PR #3208, opened 2015-04-15
TSQL has a `declare` keyword

Crisfole:
https://msdn.microsoft.com/en-us/library/ms189822.aspx


marijnh:
Thanks! Merged as 7ea815d

PR #3211, opened 2015-04-17
[markdown mode] Include optional space character after atx style headers in formatting

JonasDoebertin:
This changes will make the optional space character after ATX style headers be included within the `cm-formatting` element.

As an example, this:

```html
<span class="cm-formatting cm-formatting-header cm-formatting-header-1 cm-header cm-header-1">#</span><span class="cm-header cm-header-1"> Header</span>
```

will become this:

```html
<span class="cm-formatting cm-formatting-header cm-formatting-header-1 cm-header cm-header-1"># </span><span class="cm-header cm-header-1">Header</span>
```

I know that this is a very small change but it allows for more consistent styling, since this is how quotes are handled, too. The optional space character after the `>` is included in the formatting, as well.


JonasDoebertin:
Dang! Forgot about the tests...

marijnh:
Looks reasonable. Merged as 21e6d2f

PR #3212, opened 2015-04-17
Add support for controlling theme with vim :colorscheme command

dbarnett:
Two caveats:
  * Should probably propagate an event to the host app that the colorscheme has changed. Some apps might want to completely disable changing theme, too (but in that case at least being able to read out the theme name is still useful).
  * CodeMirror has no concept of what themes are available, so it never warns you about nonexistent themes.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -3918,6 +3918,7 @@
     // pair of commands that have a shared prefix, at least one of their
     // shortNames must not match the prefix of the other command.
     var defaultExCommandMap = [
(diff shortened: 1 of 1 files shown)


mightyguava:
Merged as de9393f, made a small edit to pass lint

dbarnett:
Is it worth loading some themes along with the vim demo page now that there's a related feature? If you go to the vim demo page now and do `:colorscheme ANYTHINGBUTDEFAULT`, it loses all theming.

PR #3213, opened 2015-04-17
Add TTCN, ASN.1 and TTCN-CFG 

oswcab:
Test suite is passing OK
bin/lint is passing OK
Maintainer comments has been addressed.

### DIFF
--- a/mode/asn.1/asn.js
+++ b/mode/asn.1/asn.js
@@ -54,13 +54,14 @@
       stream.eatWhile(/[\w\-]/);
       var cur = stream.current();
       if (keywords.propertyIsEnumerable(cur)) return "keyword";
-      if (cmipVerbs.propertyIsEnumerable(cur)) return "cmipVerbs";
-      if (compareTypes.propertyIsEnumerable(cur)) return "compareTypes";
-      if (status.propertyIsEnumerable(cur)) return "status";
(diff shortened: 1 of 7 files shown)


marijnh:
Thanks for updating the code. Another comment is that for many tokens, you are returning types that are not styled in the existing themes. Adding extra information for your custom styling is okay, but try to combine it with at least one 'standard' token style (for example, return `"builtin timerOps"` rather than just `"timerOps"`), so that other themes can also color the code halfway properly.

oswcab:
Done. Thanks for guidance.

marijnh:
Thanks! Merged as bdd6c0edc8031, 92ec62d2785a71e37, 46e9a5e08f44687, and fa11b729130b91be77. I made small changes to the ASN.1 mode, renaming the file to asn.1.js and the mode definition to define the string `"asn.1"`, so that the mode name matches the file and directory name (which holds for all other modes).

oswcab:
Thanks and kudos for a great project!

silverwind:
These new modes seem to be missing in [meta.js](https://github.com/codemirror/CodeMirror/blob/master/mode/meta.js).

oswcab:
That's right, thanks for noticing. I'll made a new pull request with this change.

PR #3223, opened 2015-04-28
configuration option 'inline' on css mode

loraderon:
Adds a 'inline' configuration option for the css mode.

Forum thread: ["Inline" CSS](http://discuss.codemirror.net/t/inline-css/215)

### DIFF
--- a/mode/css/css.js
+++ b/mode/css/css.js
@@ -12,7 +12,9 @@
 "use strict";
 
 CodeMirror.defineMode("css", function(config, parserConfig) {
+  var provided = parserConfig;
   if (!parserConfig.propertyKeywords) parserConfig = CodeMirror.resolveMode("text/css");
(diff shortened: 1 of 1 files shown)


mode/css/css.js
```diff
@@ -336,9 +336,9 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
   return {
     startState: function(base) {
       return {tokenize: null,
-              state: "top",
+              state: config.inline ? "block" : "top",
```
marijnh: I think it makes more sense to put this on the `parserConfig` object -- that's where mode-specific options go.

marijnh:
Merged as 52ad899. When adding commits to a PR, you should usually also add a comment, since I don't get notified of commits and I don't go through open PRs very often.

PR #3226, opened 2015-04-28
Added a missing argument to getSearchState

amshali:


marijnh:
Thank you. Merged as a5f2de4

PR #3227, opened 2015-04-29
Changed the linter code to work with one instance of editor and multiple...

amshali:
Changed the linter code to work with one instance of editor, multiple linters and multiple doc(as a side effect). 

The current lint code works fine when one does the linting per editor instance. In situations when you have one editor and are swapping docs in it, it will run into issues. 

The logic for picking up a linter for an editor picks up whatever was last registered(I think). This does not allow one to select a linter based on the mode of the editor. In addition, it is not possible to apply multiple linters to one editor. 

This patch changes the logic in css, json and javascript linters so that they register as global helpers with a predicate function which chooses them when the right mode is passed in. It also changes the lint.js code so that it runs all of the registered linters.

Pardon me if the logic for state management is a bit fishy. Feel free to change it however you want or let me know how to do it. 

Thank you!

### DIFF
--- a/addon/lint/css-lint.js
+++ b/addon/lint/css-lint.js
@@ -15,7 +15,7 @@
 })(function(CodeMirror) {
 "use strict";
 
-CodeMirror.registerHelper("lint", "css", function(text) {
+CodeMirror.registerGlobalHelper("lint", "css", function(mode) { return mode === "css"; }, function(text) {
   var found = [];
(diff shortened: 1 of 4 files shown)


marijnh:
Attached patch fixes the problem of not being able to change mode in a different way.

In what kind of situation do you want to apply multiple linters?

PR #3229, opened 2015-05-02
Fixing non-firing of "close" on completion close

mihailik:
The manual says that completion should fire "pick", "close", "select" and "shown".

Here: https://github.com/codemirror/CodeMirror/blob/master/doc/manual.html#L2556

Currently only "pick" and "select" seem to be firing. This change fixes "close".


marijnh:
Seems that one, along with the `"shown"` event, was accidentally dropped in the recent refactors. I've attached a patch which restores both, and checks for the necessary preconditions when firing the `"close"` event.

PR #3242, opened 2015-05-06
Make demo link back to addon doc

anaran:
All demos currently have an a.active link with `href="#"`, which is of fairly limited value.

This proposed patch links back to the related addon section in the manual, which is the most useful purpose I could think of for that link.

Currently the link, when clicked, *does nothing*.

As always, there were some exceptions to the rule:

* Some demo filenames did not match the `id` in the manual.
* Some demos cover multiple addons (folding demo covers addons foldcode and foldgutter).

I fixed these manually.

There are some demos which are not in the demo folder. I have left those alone.

### DIFF
--- a/demo/activeline.html
+++ b/demo/activeline.html
@@ -20,7 +20,7 @@
     <li><a href="https://github.com/codemirror/codemirror">Code</a>
   </ul>
   <ul>
-    <li><a class=active href="../doc/manual.html#addon_activeline">Active Line</a>
+    <li><a class=active href="../doc/manual.html#addon_active-line">Active Line</a>
   </ul>
 </div>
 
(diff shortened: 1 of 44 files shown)


anaran:
I'm working to resolve this mess. In the end this will be a clean pull request of two commits, which I can squash to one if desired.

anaran:
OK, that is cleaned up now. Please let me know when it is necessary to squash to a single commit.
I find it valuable to seen the automated for changed separate from the manual fixup in the second commit.

marijnh:
These links (the red ones in the navigation area on the left) are supposed to indicate your current position. Them linking to the thing you are currently looking at is intentional and not something that should be fixed.

Issue #3246, opened 2015-05-07
SQL-Hint does not respect a table alias, but it DOES detect them


rathboma:
Here's an example:

```sql
select t. from super_long_complicated_table t
```
If you call showHint while the cursor is at `t.` it will suggest (correctly) all the column names that belong to alias t. However, if you select one it will replace the selection with the fully qualified name.

So when I want it to insert `t.name`, it will autocomplete to `super_long_complicated_table.name`, even though it knows that t is an alias for the table, and it has already detected that.

So the result should be:

```sql
select t.name from super_long_complicated_table t
```

but it comes out as:

```sql
select super_long_complicated_table.name from super_long_complicated_table t
```

marijnh:
Fixed via #3248

PR #3248, opened 2015-05-07
potential (hacky) fix for alias tables

rathboma:
Potential fix for issue #3246. It's hacky but it works


marijnh:
Looks good enough. Merged as 736140c06f675e and added a test in d47bc479c884

PR #3249, opened 2015-05-08
Add addon/help/help.js with tests and documentation

anaran:
This is a squashed commit of a merge --no-commit from my gh-pages branch.

It includes the addon, the associated manual.html update, and a new help_test.js module already hooked into the test suite.

The tests were passing locally.

You can look at the demo at http://anaran.github.io/CodeMirror/demo/help.html, the documentation at http://anaran.github.io/CodeMirror/doc/manual.html#addon_help, and the new tests at http://anaran.github.io/CodeMirror/test/#addon_help_,verbose

gh-pages (hosting all the links above) will become an integration branch for this pull request along with #3242.

Now I wonder what I may have missed...

anaran

### DIFF
--- a/addon/help/help.js
+++ b/addon/help/help.js
@@ -1,0 +1,115 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+(function (mod) {
+  if (typeof exports == 'object' && typeof module == 'object') // CommonJS
+    mod(require('../../lib/codemirror'));
+  else if (typeof define == 'function' && define.amd) // AMD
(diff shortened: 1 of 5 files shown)


marijnh:
I still do not believe this belongs in the distribution. Firstly, Emacs-style "buffers" are not something CodeMirror interfaces are based on, so this isn't something that many people would be able to use. Secondly, the actual functionality provided here seems very minimal (JSON-ifying some existing information). So, since people will want to put their own interface on this anyway (showing it in a popover as an HTML list, or as you did, in a buffer, or in some kind of tooltip, or as a separate help page...), we might as well leave them to implement the whole thing in their own system, rather than providing a half-solution that isn't going to fit properly most of the time.

PR #3259, opened 2015-05-13
A Mathematica mode for CodeMirror

cbarbat:
This is a mode for the Mathematica syntax.
Mathematica is copyright by Wolfram Research Inc.


marijnh:
Thanks. Are you sure it is worth including a list of a thousand character names (weighting over a 100kb), rather than just matching anything that looks like `\[foo]` as a character reference?

marijnh:
Actually, the same goes for the ridiculously long list of keywords. This would be, if merged, by far the biggest mode file in the distribution. Since a lot of people are going to be downloading the file without ever using it, I don't think that's okay. If you can't cut down the list of keywords/builtins to something more reasonable, this is probably best hosted in a separate repository.

cbarbat:
@marijnh I managed to reduce the size of mathematica.js to about 5.5 kbytes.

marijnh:
Wonderful! Merged as de9a6b9

PR #3265, opened 2015-05-17
[clike mode] Add C# keywords/builtins: async, await, Action, Func, Task

dorny:


marijnh:
Thanks. Merged as e3cc695

PR #3269, opened 2015-05-20
MySQL function is GROUP_CONCAT, not GROUPBY_CONCAT

madhuracj:
See https://dev.mysql.com/doc/refman/5.6/en/group-by-functions.html#function_group-concat


marijnh:
Thank you. Merged as daae9be

Issue #3261, opened 2015-05-15
Selection problem in IE11, when leaving an iframe while selecting


mboehlke:
We found an issue with codemirror 5.2.0 (and previous).
We have a codemirror instance embedded in an iframe. 
While selecting some text with pressed left mousebutton and thereby unintentional leaving the iframe 
and release the mousebutton there: the selection of text occurres while mousemove and we found the 
only way to return to normal behaviour by clicking with the right button. 
We have created a minimal sample "EditorMan.html" with a left area outside an iframe on the right (see below). 
The misbehavior occurs with Win7 IE11.  (actual versions) 
It does not occur with Win7 FF and Chrome; and on OSX Safari.
Other combinations not tested. 
Despite that: Thank you for the great code.
Manuel 

The HTML file: EditorMan.htm
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
	<meta charset="utf-8" />
	<title>Editor MainView</title>

	<script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
	<script type="text/javascript">
		$(document).ready(function ()
		{
			document.getElementById("idRecipeCod").src = "Editor.htm";
		});
	</script>
</head>
<body>
	<div id=idLeft style="position:absolute; top:0; left:0; width:150px; bottom:0; background:beige">LEFT</div>
	<iframe id=idRecipeCod style="position:absolute; top:0; left:158px; right:0; bottom:0"></iframe>
</body>
</html>

The HTML file: Editor.htm
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
	<head>
		<meta charset="utf-8"/>
		<title>Text Editor</title>

		<link rel="stylesheet" type="text/css" href="./codemirror/lib/codemirror.css" />

		<script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
	    <script type="text/javascript" src="./codemirror/lib/codemirror.js"></script>

		<script type="text/javascript">
			$(document).ready(function () {
				var mycodemirror = CodeMirror(document.body, {
					value: "function myScript()\n{ var a = 1;\n var b = 42;\n return 100;\n}\n",
					mode: "javascript"
				});
			});
		</script> 
	</head>
	<body>
	</body>
</html>

marijnh:
I am aware of this, but have not been able to come up with a solution.

- We normally rely on the `mouseup` event to abort a drag-select
[...]

PR #3272, opened 2015-05-21
Workaround for  #3261 Selection problem in IE11, when leaving an iframe while selecting

pstadermann:
If the mouse leaves the iframe during text selection, the MouseEvents are no longer sent to Codemirror. If the user lets go of the left mouse button, CodeMirror does not know that the text selection has stopped.

Chrome and Firefox allow to check the states of the mouse buttons on mouse move. Codemirror does this and stops the text selection. This does not work on any version of IE.

Workaround for IE: Check on each left click, if text is currently being selected. If yes, the selection is finished. This is not perfect, but prevents Codemirror from starting a second text selection and going crazy.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -87,6 +87,7 @@
       focused: false,
       suppressEdits: false, // used to disable editing during key handlers when in readOnly mode
       pasteIncoming: false, cutIncoming: false, // help recognize paste/cut edits in input.poll
+      selectingText: false,
       draggingText: false,
       highlight: new Delayed(), // stores highlight worker timeout
       keySeq: null,  // Unfinished key sequence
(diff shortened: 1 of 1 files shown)


lib/codemirror.js
```diff
@@ -3607,6 +3609,12 @@
     var display = cm.display, doc = cm.doc;
     e_preventDefault(e);
 
+    // #3261: make sure, that we're not starting a second selection
+    if (cm.state.selectingText) {
+      cm.state.selectingText(e);
+      return;
```
marijnh: Why not check this in `onMouseDown`? That'd also save the extra condition above (when deciding between a drag and a select).

marijnh:
Merged in slightly changed form as b22783d

PR #3274, opened 2015-05-22
Respect alias here as well

madhuracj:
This is further to https://github.com/codemirror/CodeMirror/issues/3246 and as a fix for https://sourceforge.net/p/phpmyadmin/bugs/4916/

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -127,13 +127,13 @@
 
     if (columns) {
       addMatches(result, string, columns, function(w) {
+        var tableInsert = table;
+        if (alias == true) tableInsert = aliasTable;
         if (typeof w == "string") {
-          var tableInsert = table;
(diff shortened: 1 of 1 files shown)


marijnh:
Could you give a more descriptive commit message and add a test to `test/sql-hint-test.js` that verifies that whatever it is you are fixing here stays fixed in the future?

madhuracj:
I went through the tests, but all of them seems to test what hints are shown rather than what is being added to the editor. I'm sorry, i'm not very familiar with the code base, so I decided to describe the issue in detail in https://github.com/codemirror/CodeMirror/issues/3281 referring this as a potential fix.

marijnh:
I've merged this as 172a78163d3ad8

The tests that use display text test for both the inserted text and the display text, so it was not complicated at all to add a test. See attached patch.

madhuracj:
Great. Thanks!

PR #3283, opened 2015-05-26
[monokai theme] Add support for cm-header

vincentwoo:
Matches color for `cm-link` style


marijnh:
Sure, merged as 64b11dc

PR #3289, opened 2015-05-27
[python mode] Add async and await keyword to python3

Carreau:
Since May 5th. `async` and `await` are Python3 keywords. 
cf https://www.python.org/dev/peps/pep-0492/

The new version of Python is still in beta, but start to get some use. And getting the correct highlighting out will help people not use `async` and `await` as variable name.


marijnh:
Thanks. Merged as 6482803

PR #3294, opened 2015-05-30
Fix JSFiddle link

baconface:


marijnh:
Thanks. Merged as f0b870d

PR #3299, opened 2015-06-02
Twig mode added

anthonydugois:
Just a mode to support the Twig template engine (http://twig.sensiolabs.org/).

### DIFF
--- a/mode/twig/index.html
+++ b/mode/twig/index.html
@@ -1,0 +1,45 @@
+<!doctype html>
+
+<title>CodeMirror: Twig mode</title>
+<meta charset="utf-8"/>
+<link rel=stylesheet href="../../doc/docs.css">
+
+<link rel="stylesheet" href="../../lib/codemirror.css">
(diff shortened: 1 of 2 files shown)


bobrocke:
@anthonydugois Wonderful! I had been hoping someone would write one.

Will it do multiple-modes like Twig together with HTML and JavaScript?

marijnh:
Nice! Merged as 23a03a9b7960 and linked in 00d968d6cbf.

The state maintenance could use some cleaning up, for example by unifying everything in a single state machine and dispatching to the tokenizer function that can handle the current state (by storing the function in the state object, rather than a plethora of flags).

For mixing this with HTML, you should be able to use the [multiplex](http://codemirror.net/doc/manual.html#addon_multiplex) addon to multiplex between the outer and the Twig mode, by dispatching on `{%`, `{#`, `#}`, and `%}`. For an example of this, see the source for the [Handlebars demo](http://codemirror.net/mode/handlebars/)

PR #3303, opened 2015-06-03
Create factor.js

kerabromsmu:
Factor language simple mode.


marijnh:
Wonderful! Could you add an index.html demo file for the mode (copy-paste one from another mode and adjust)

kerabromsmu:
I've created another pull request -- with factor.js and index.html. This one turned out to be glitchy.

PR #3305, opened 2015-06-03
Add Factor mode to CodeMirror

kerabromsmu:
Second try


marijnh:
Thank you! Merged as 63155b7.

(You can, by the way, simply push to the branch used for a pull request to change the patches in the request. That way, you don't have to create a new PR, but simply ask for another review on the old one.)

kerabromsmu:
Thanks!! I was trying to do that, but didn't find how. ;]

PR #3312, opened 2015-06-06
Update factor.js

kerabromsmu:
defineMode instead of defineSimpleMode - sort of simple mode simulation; slight optimization; fixed a few glitches.

### DIFF
--- a/mode/factor/factor.js
+++ b/mode/factor/factor.js
@@ -88,9 +88,9 @@
         token: function(stream, state) {
           var stDef = mD[state.state];
           var stMore = state.more;
-          console.log(stMore);
+          //console.log(stMore);
           if (stMore && stMore.length>0) { 
-            console.log(state.words[0] + " -" + state.more[0] + "- " + state.state);
+            //console.log(state.words[0] + " -" + state.more[0] + "- " + state.state);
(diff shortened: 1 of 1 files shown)


marijnh:
Woah, you basically reimplemented the simple mode addon. What limitation did you run into, precisely? And would it make sense to extend the simple mode addon instead?

kerabromsmu:
Yep, you're right. I had to do it to make it work in adobe brackets. Turns
out they don't load the simple mode, and the way they let you load the
modules, I cannot use it. So I had to emulate it.

Actually, I made an extension for brackets that uses the same code and just
updated it here to match that one for consistency. Not only consistency,
but that too.

Not sure what you can do about it.

marijnh:
Have you tried opening an issue for Brackets to load the simple-mode addon by default? It is tiny, and a bunch of modes are using it, so they may be open to that.

kerabromsmu:
Actually, I don't remember. Probably I did, but still went for a quicker
solution, because otherwise I would have to wait for the next version of
brackets to release. I'll check anyway

marijnh:
In any case, I'm not merging this, since it is simply duplicating code that's already in the distribution.

Issue #2569, opened 2014-05-22
why remove bower.json's version?


laurelnaiad:
How is it helping to omit the version from bower.json?

It renders bower incapable of knowing which version you have locally and thus incapable of following your bower.json file's instructions to update the component... thus every app that references codemirror through bower gets locked to the version that a particular user happened to get when they first ran bower install on one's repo.... 

https://github.com/marijnh/CodeMirror/commit/82b62f3bce55d081f9a7c2f9bc5ed0f11f26d7b3#diff-0a08a7565aba4405282251491979bb6bL3

This seems to be altogether counterproductive.

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #3322, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "codemirror",
-  "version":"5.3.1",
   "main": ["lib/codemirror.js", "lib/codemirror.css"],
   "ignore": [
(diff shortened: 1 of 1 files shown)


marijnh:
Please see #2569 and reach some kind of consensus with @laurelnaiad

kkirsche:
Na. But consider that Bower may be under NPM anyway which is still something being discussed bower/bower#1520

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

marijnh:
All right. Attached patch removes it along with the line in the release script that was responsible for updating it.

PR #3323, opened 2015-06-11
Initialize docs before creating tern server instance

angelozerr:
This PR initialize docs before creating tern server instance, because in some case, tern server creation call getFile which use docs.

I have this case by using requirejs tern plugin + custom plugin which defines requirejs modules : 

```json
{
    "!name": "delite",
    "!define": {
      "!requirejs": {
        "delite/register": {
          "!type": "fn(tag: string, superclasses: [], props: ?) -> fn()",
      }
    }
  }
```


marijnh:
Good point. Merged as 61e27f0

angelozerr:
Thanks!

PR #3324, opened 2015-06-11
Kotlin mode minor fixes

satamas:
Add interface keyword and info about comments.


marijnh:
Thanks. Merged as ee62a07

PR #3327, opened 2015-06-12
Add line commenting to VB.NET mode

vincentwoo:
in VB the comment char is just a single quote: `'`


marijnh:
Thanks. Merged as a08e3ec

PR #3335, opened 2015-06-17
Update runmode.node to match core

ForbesLindesay:
This fixes problems in jade where:

```jade
doctype html
html(lang="en")
  head
    title= pageTitle
    script(type='text/javascript').
      if (foo) {
         bar(1 + 5)
      }
  body
    h1 Jade - node template engine
    #container.col
      if youAreUsingJade
        p You are amazing
      else
        p Get on it!
      p.
        Jade is a terse and simple
        templating language with a
        strong focus on performance
        and powerful features.
```

was correctly highlighted in the browser but incorrectly highlighted in node.js

### DIFF
--- a/addon/runmode/runmode.node.js
+++ b/addon/runmode/runmode.node.js
@@ -5,17 +5,38 @@
 
 // declare global: StringStream
 
-function splitLines(string){ return string.split(/\r?\n|\r/); };
+function splitLines(string){return string.split(/\r\n?|\n/);};
 
-function StringStream(string) {
+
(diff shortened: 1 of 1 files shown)


marijnh:
Merged as 0bc820bf583, and copied a helper function you missed in 59d78210d6d.

(This is all awful and at some point in the future I'll turn CodeMirror in a bunch of ES6 modules so that you can load parts in Node without duplicating code.)

ForbesLindesay:
:+1:

Issue #3337, opened 2015-06-18
Enhance search addons custom keybinding support.


jeff-blaisdell:
At my place of work we've had several requests to change the editor's search functionality to mimic the browser's search.  Specifically:
* Search bar should remain displayed while user cycles through matches.
* When closing search, the highlighted searches should disappear.
* Use same shortcuts Enter (Next) / Shift-Enter (Previous) as browser.

I do not believe this is possible to accomplish in the current state due to the search addons current usage of the dialog addon.  It would be useful to expose the dialog onClose and onKeyDown extension options through the search addon to allow for further customization of the search key bindings.

marijnh:
Done via #3338

PR #3338, opened 2015-06-18
3337: Enhance search addons custom keybinding support.

jeff-blaisdell:
I want to resolve issue #3337 

After a review(s), I plan to add tests.

Demos
Utilizing Enhancement:
[Incremental Search](http://www.googledrive.com/host/0BzIoeqmeuBlnWFVpNFNNY3FnUVk/demo/search-incremental.html)

Current Functionality is still Default:
[Existing Search](http://www.googledrive.com/host/0BzIoeqmeuBlnWFVpNFNNY3FnUVk/demo/search.html)

Visually, the change looks like this:
screen shot 2015-06-18 at 2 05 20 pm


.editorconfig
```diff
@@ -0,0 +1,13 @@
...
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
jeff-blaisdell: EditorConfig works across IDEs / Text Editors to maintain some basic code styles.  I believe the above matches the style requirements articulated in CodeMirror's contribution style guide.

lib/codemirror.js
```diff
@@ -8220,6 +8220,24 @@
...
+    return !!value && (type == 'object' || type == 'function');
+  }
+
+  var _isFunction = CodeMirror.isFunction = function(value) {
+    // The use of `Object#toString` avoids issues with the `typeof` operator
+    // in older versions of Chrome and Safari which return 'function' for regexes
+    // and Safari 8 equivalents which return 'object' for typed array constructors.
+    return isObject(value) && objToString.call(value) == funcTag;
+  }
+
```
jeff-blaisdell: I wanted to do some tests for functions, and didn't see any existing utils provided to accomplish.  This is a copy/paste job from [Lodash](https://github.com/lodash/lodash/blob/master/lodash.js#L8669)

marijnh:
I can't make much of your description of this functionality, but the demo seems to show some kind of incremental search. The patch seems to formulate the functionality in a strange way and doesn't work very well (it gets confused if you backspace or paste). I think it would make more sense to define a new command for starting an incremental search. Also, use the `"input"` event, not key events, to notice when the dialog's content changes.

jeff-blaisdell:
@marijnh As for the functionality, the request I'm being asked to do is make the search like Google Chrome's search.

As for the strange implementation, I wasn't particularly happy with it myself - this is my first real look at CodeMirror source and getting an understanding of what's going on.

I'll take another look and see if I can encapsulate in a single command.  I'm not sure if the "input" event would work for my use case, without additional changes.  The request was to leave the dialog open through out the search - and the dialog internals auto close the dialog on the "keydown" event.

Here's the relevant code from dialog.js:
```
      if (options.onInput)
        CodeMirror.on(inp, "input", function(e) { options.onInput(e, inp.value, close);});
      if (options.onKeyUp)
        CodeMirror.on(inp, "keyup", function(e) {options.onKeyUp(e, inp.value, close);});

      CodeMirror.on(inp, "keydown", function(e) {
        if (options && options.onKeyDown && options.onKeyDown(e, inp.value, close)) { return; }
        if (e.keyCode == 27 || (options.closeOnEnter !== false && e.keyCode == 13)) {
          inp.blur();
          CodeMirror.e_stop(e);
          close();
        }
        if (e.keyCode == 13) callback(inp.value, e);
      });
```

jeff-blaisdell:
@marijnh I completed a second pass based on your comments.  I removed the usage of keydown events, in favor of the input event.  Also - simplified the usage to just an incrementalSearch flag.  

I'm struggling a bit to find the best place to override the dialog's keybindings.  As mentioned above, I'm trying to implement a Google Chrome like search - which means using Enter / Shift-Enter for search navigation.  Those bindings collide with existing CodeMirror default bindings, and really only make sense in the context of the search dialog.  

I managed to make things work based on the dialog input event, but this implementation depends on the fact that the dialog internally binds Enter to executing the callback.  Maybe this is fine, but to me it feels more appropriate to override the dialog's keydown handler.

Thoughts?

addon/search/search.js
```diff
@@ -76,19 +116,18 @@
...
-        if (cm.showMatchesOnScrollbar) {
-          if (state.annotate) { state.annotate.clear(); state.annotate = null; }
-          state.annotate = cm.showMatchesOnScrollbar(state.query, queryCaseInsensitive(state.query));
+        initQuery(cm, queryStr);
+
+        if (cm.state.incrementalSearch) {
+          // In incremental search we need to prevent default behavior
+          // from pre-maturely closing the dialog box.
+          CodeMirror.e_stop(keyEvent);
+          findNext(cm, (keyEvent.shiftKey === true));
```
jeff-blaisdell: Rather than providing custom keybindings for our search dialog, rely on the fact the dialog binds Enter to the execution of this callback.  I don't like this.

Another option is to override the dialog's keydown event and specify search specific bindings.

Other thoughts or opinions?

addon/search/search.js
```diff
@@ -38,9 +38,35 @@
...
+          state.annotate = null;
+        }
+        state.annotate = cm.showMatchesOnScrollbar(state.query, queryCaseInsensitive(state.query));
+      }
+      state.posFrom = state.posTo = cm.getCursor();
+    }
+  }
   function SearchState() {
-    this.posFrom = this.posTo = this.lastQuery = this.query = null;
+    this.posFrom = null;
```
marijnh: Definitely don't make random style changes to projects you didn't write.

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# editorconfig.org
```
marijnh: This has nothing to do with the pull request as a whole. Split such things into different pull requests so that discussion and review can stay focused.

In this case, I am not really interested in including this file, as none of the editors I use recognize it, and I don't want to accumulate tooling cruft that I have no relation to in my projects.

.gitignore
```diff
@@ -6,3 +6,4 @@
 *.swp
 .idea
 *.iml
+.DS_Store
```
marijnh: Again, different pull request. No matter how trivial, it does not belong to this feature.

addon/search/search.js
```diff
@@ -158,6 +201,10 @@
     });
   }
 
+  CodeMirror.defineOption("incrementalSearch", false, function(cm, val) {
+    cm.state.incrementalSearch = (val === true);
```
marijnh: I'd prefer this as a separate command that passes around its own state. It is conceivable to want both incremental and non-incremental search on the same instance, so having it as an editor-global flag is undesirable.

addon/search/search.js
marijnh: So do I understand correctly that with this change you're no longer updating the search on every change to the query field, but only on enter presses?

addon/search/search.js
jeff-blaisdell: reverted. done.

.editorconfig
jeff-blaisdell: removed.

.gitignore
jeff-blaisdell: removed.

addon/search/search.js
jeff-blaisdell: That's not quite right.

So in both the original and this PR version, the dialog executes the search callback once it detects an Enter keydown event.  This is where they start to diverge:

In original:
The search callback executes and parses the query into SearchState.  Then the global keybindings, i.e. Cmd-G / Shift-Cmd-G are bound to the search commands to move next / prev.

In PR:
The search callback executes and will only parse the query into SearchState if the query string has changed from the previous execution.  This was necessary because the dialog remains on screen, and the user may/may not choose to modify the search.  Because the dialog is still on screen, the dialog's local keybindings are still active and every keydown of Enter will trigger the search callback which will move to next Search term.

Both versions are pretty much the same thing.  The new one just uses Enter/Shift-Enter to search next/previous and it also keeps dialog visible.

About packaging this into a command...  How would you see this command used?  I see the search's findNext / findPrev commands being bound to the global keybindings Cmd-G / Shift-Cmd-G.  Because Enter is an already bound key in the global context, I don't see how we could leverage it for search.  It works today, because the Dialog provides it's own keybindings that override those defined at global level.

addon/search/search.js
jeff-blaisdell: See above.  I can retry a command method implementation if you still think it makes sense.  I'm not sure if this is really an incremental vs non-incremental search thing.  It is pretty much the same search with different keybindings and the dialog stays open rather than closing.

marijnh:
Could you take a look at patch 404e308 and see if it meets your requirements?

jeff-blaisdell:
It looks good.  Looks like your patch had a couple of bugs that I had accounted for in this pr.  I made a PR against that patch w/ fixes.
https://github.com/codemirror/CodeMirror/pull/3389/files

marijnh:
Great. I've merged your pull request.

PR #3341, opened 2015-06-19
Master

dperetti:
New option to disable the scroll wheel in the editor so one can scroll
the holding page without being stuck in the editor.


marijnh:
You can already scroll the outer page without being stuck in the editor. Our wheel event handler only preventDefaults the event in rare circumstances, so I am not sure this patch is addressing a real problem.

PR #3354, opened 2015-06-30
Fix ASN.1 and TTCN-CFG mode names in mode.js

oswcab:
These two modes were using unsupported characters in the mode
name which prevented integration into Gerrit.


marijnh:
Who exactly determined that these characters are unsupported? It seems to be Gerrit, and if that's so, I think submitting a patch to Gerrit would be more appropriate.

PR #3358, opened 2015-07-03
new themes, minimap, new mode(brainfuck), and line numbers in runmode

mkaminsky11:
#### minimap
+ adds a basic minimap to the right side of the editor. Activated with the `minimap` option
+ pulls colors from editor
+ no "click to scroll" functionality yet like [Sublime Text's](http://programmers.stackexchange.com/questions/156362/does-sublimes-minimap-improve-productivity)
+ no indication of currently visible section (similar to above)
+ add documentation for minimap
+ ~~**note**: can slow down editor, needs to be optimized~~  **fixed**
+ ~~**note**: `minimap.js` does not shrink the actual editor, so text will pass behind the minimap (see below)~~ **fixed**
+ **note**: makes the horizontal scrollbar act kind of weird. Setting the width of the scrollbar to `calc(100% - 200px)` does not fix the issue (see below).

minimap

#### themes
+ fix a typo in `solarized.css`
+ added 4 new themes: `seti` `dracula` `yeti` `material`
+ added `themes/meta.js`, to allow for lazy loading themes, and so that new themes can easily be integrated into applications by simply referencing `meta.js`
+ added `mode` dropdown in the "themes" demo to show how the themes highlight tags, qualifiers, etc.

**dracula**
dracula
**seti**
seti
**yeti**
yeti
**material**
material

#### new mode (brainfuck)
brainfuck

#### line numbers in runmode
Allows for the `options` parameter to have a `lineNumbers` attribute.

line numbers


vincentwoo:
Pretty interesting! How performant is the minimap on larger documents?

mkaminsky11:
since it pulls all of the information from CodeMirror's viewport, which only shows a snippet of larger documents for performance reasons, it's pretty good. 

The only disadvantage of this is that it can't show large parts of the file (only what is rendered by CodeMirror), but the performance gains are worth it.

marijnh:
I've pushed the themes (f52e64d095db6a8bbb) and the Brainfuck mode (84c8c92968174df). I don't consider the linenumbers-for-runmode or the minimap implementations to be production-ready. Next time, _please_ create one pull request per feature, clean up your patch history, and only submit code that is done. This saves me a lot of time.

PR #3359, opened 2015-07-03
ICEcoder's default theme

mattpass:
Spent quite a while perfecting this (IMO), but a lot of people seem to
really like it. Hope you do too.

Note: need to add to the themeing demo on the site, I didn't cover that.

### DIFF
--- a/theme/icecoder.css
+++ b/theme/icecoder.css
@@ -21,7 +21,6 @@ ICEcoder default theme by Matt Pass, used in code editor available at https://ic
 .cm-s-icecoder span.cm-string-2 {color: #6cb5d9}                /* blue */
 
 .cm-s-icecoder span.cm-meta {color: #555}                       /* grey */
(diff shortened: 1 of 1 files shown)


mattpass:
This is a sample of what the theme looks like BTW:

untitled-1

vincentwoo:
Looks nice. Reminds me of a higher-contrast monokai

mattpass:
Thanks. Was trying to use what I thought were the best parts of Seti, Monokai and Flatland.

marijnh:
Merged as 8d79749. Have you considered making `variable-2` a little brighter? In JavaScript code, you'll often have a lot of these (http://codemirror.net/demo/theme.html#icecoder) and the current color makes for a rather low-contrast look.

mattpass:
Good feedback, thanks. Looking at it, maybe is a bit subdued under JS mode. Put in a new pull request with a slight colour boost:

new-hex

Issue #3361, opened 2015-07-06
Auto resize demo scrolling issue


maniax89:
Couldn't find an exact issue already opened for this. Things that looked similar included https://github.com/codemirror/CodeMirror/issues/2368 and potentially when fixing https://github.com/codemirror/CodeMirror/issues/1514 

Basically, if you go to the resize demo http://codemirror.net/demo/resize.html and delete one line and try to scroll using a touchpad or mousewheel inside the codemirror, the numbers and textarea will scroll.

Expected: It shouldn't scroll since no scrollbars are present.

Steps to reproduce:
1) Go to http://codemirror.net/demo/resize.html (has the codemirror height set to auto)
2) Delete a line so that the total number of lines is less than what you started with
3) While focused on the code mirror text area, hover the mouse over that area.
4) Scroll (either using a touch pad or mouse wheel) - content will scroll beyond the visible area

I believe it is due to the height of the gutters being set to be larger than the box itself - but those who have been working on the issue may have better insight to the issue.

I have reproduced this on Mac OSX 10.9.5 on the following browsers:
1) Chrome 43.0.2357.130 (64-bit)
2) Firefox 36.0.1
3) Safari 7.1.3 

I also ran Windows 7 on a VM and was able to reproduce it on:
1) IE 9
2) Firefox 38.0.5

PR #3362, opened 2015-07-06
Fix scrolling issue when height: auto

maniax89:
After deleting a line with height: auto set,
the codemirror would scroll even though it shouldn't

fixes https://github.com/codemirror/CodeMirror/issues/3361


marijnh:
That fix introduces a new layout round-trip into the redraw logic, which is definitely not acceptable. See my alternative patch 31d89d6e4 . Test merged as 50aba2033fc

maniax89:
Looks great! Thanks for fixing this. Great plugin too.

Any idea when the next release is that includes this patch?

marijnh:
About one week from now (I release monthly, around the 20th)

PR #3370, opened 2015-07-09
Update tern.css

Oreoluwao:
I noticed that the image that accompanies auto-completion suggestions when the tern addon is active gets cut slightly because the default line height is slightly smaller than the image height.
This change ensures that that the image is fully shown.


marijnh:
Thank you. Merged as efca209

PR #3373, opened 2015-07-10
Emacs Keymap: Fix handling of non-integer line numbers in goto-line.

xiemaisi:
This fixes cases such as `Alt-G`, `G`, "2.2", which would previously lead to a `TypeError`.


marijnh:
Ah, the legacy of C's operator precedence choices. Thanks for catching that! Merged as 127a6e0

PR #3374, opened 2015-07-10
Erlang Mode: Remove duplicate if branches.

xiemaisi:
Unless I'm missing something, the two branches in question are duplicates of branches further up and hence unreachable. Not a bug per se, but a potential readability/maintenance problem.

### DIFF
--- a/mode/erlang/erlang.js
+++ b/mode/erlang/erlang.js
@@ -220,16 +220,12 @@ CodeMirror.defineMode("erlang", function(cmCfg) {
         }else{
           return rval(state,stream,"function");
         }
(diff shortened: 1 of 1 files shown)


marijnh:
Indeed. I unfortunately don't review all code in depth, and nonsense like this is probably common in the mode files. Thank you for doing a more serious review. Merged as 8cc9eb8

marijnh:
(Actually, looks like you used an automated tool to find these. Nice!)

xiemaisi:
> (Actually, looks like you used an automated tool to find these. Nice!)

Indeed :smile_cat:

PR #3375, opened 2015-07-10
Remove a few useless expressions in vim_test.js.

xiemaisi:


marijnh:
Merged as c2350fd. Not sure how those ended up there and why the tests passed without them, but so it goes.

PR #3376, opened 2015-07-10
VIM Mode: Fix notification about lacking pattern support in sort command.

xiemaisi:


keymap/vim.js
```diff
@@ -4308,7 +4308,7 @@
               if (decimal + hex + octal > 1) { return 'Invalid arguments'; }
               number = decimal && 'decimal' || hex && 'hex' || octal && 'octal';
             }
-            if (args.eatSpace() && args.match(/\/.*\//)) { 'patterns not supported'; }
```
xiemaisi: Note that `args.eatSpace()` can never return `true` here, since there is another `eatSpace()` just a few lines above.

marijnh:
Thank you. Merged as 9a532d3

PR #3377, opened 2015-07-10
Use displayName if it exists for completion display text

angelozerr:
Tern completion for required module returns name (module wuith quote) and displayName (module without quote). This PR gives the capability to display completion for required module without quote.


marijnh:
Thanks! Merged as b8c4ea1

PR #3378, opened 2015-07-11
Add a .npmignore file

belhaj:
The .npmignore file will keep certain files/directories out of the package when installing CodeMirror with npm. I reused the same ignore list found in bower.json.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,11 @@
+**/.*
+node_modules
+components
+bin
+demo
+doc
+test
+index.html
+package.json
+mode/*/*test.js
+mode/*/*.html


marijnh:
NPM will use `.gitignore` when `.npmignore` isn't present. Is there any specific thing that wasn't being handled by `.gitignore` that you are addressing here?

belhaj:
The list of files/directories is in the PR. Directories like `bin`, `demo`, `doc`, `test`, ... etc are usually deemed as unnecessary when installing CodeMirror as a dependency through npm. But that doesn't mean they should be excluded from a Git repository either.

marijnh:
`bin` should definitely be included, since it contains utilities that users might want, and there isn't actually a `components` directory in the repository. Stripping `package.json` also seems like a bad idea. But I guess stripping the docs and examples from the release is probably good. Merged as 4c4c0c6

PR #3379, opened 2015-07-11
[sublime keymap] Fix scrollLine[Up/Down] bindings

vincentwoo:
In current CodeMirror on OSX, hitting cmd + up or down only scrolls the viewport (using scrollLineUp or Down) instead of jumping to the start or end of the entire buffer.

In Sublime OSX, the default key for scrolling one line up or down is Ctrl-Alt-Up/Down. In Windows/Linux, it is just Ctrl-Up/Down.


marijnh:
Thank you. Merged as baad003

PR #3380, opened 2015-07-11
[travis] Use stable node and container infrastructure

vincentwoo:
Should improve test robustness as well as speed - specifying `sudo: false` will let Travis use its new container-based infrastructure.


marijnh:
Sounds good. Merged as 0026883

PR #3381, opened 2015-07-13
Color boost on variable-2 to be brighter

mattpass:
Was a little subdued


marijnh:
Thanks. Merged as e585ae8

PR #3383, opened 2015-07-13
Added name properties to pre defined keymaps

mkarajohn:
`cm.removeKeyMap()`, as per documentation, depends on the `name` property of the active keymap in order to remove it if a string is passed as an argument. Most predefined keymaps did not have a `name` property, leading to unexpected behaviour.

>cm.removeKeyMap(map: object)
>Disable a keymap added with addKeyMap. Either pass in the key map object itself, or a string, which will be compared against *the name property of the active key maps*.

### DIFF
--- a/keymap/emacs.js
+++ b/keymap/emacs.js
@@ -398,7 +398,8 @@
     "Ctrl-X H": "selectAll",
 
     "Ctrl-Q Tab": repeated("insertTab"),
-    "Ctrl-U": addPrefixMap
+    "Ctrl-U": addPrefixMap,
+    "name": "emacs"
   });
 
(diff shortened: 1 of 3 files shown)


marijnh:
The keymaps in the distribution are meant to be used as base keymaps (the `keyMap` option). Using them with `addKeyMap` is probably not what you want.

mkarajohn:
Ok, I see. 

So,  what would be the best way to enable and disable the sublime keymap, for example?

marijnh:
You probably want a select, not a checkbox, for choosing a base keymap. You can do `editor.setOption("keymap", "default")` to go back to the default map.

PR #3388, opened 2015-07-14
[squirrel mode] Add

M1cha:
http://squirrel-lang.org


marijnh:
Looks great. Pushed as 0c2264b. I've removed out the example editor on the demo page, since that is already getting too crowded.

PR #3389, opened 2015-07-14
Issue 3337: A couple fixes for new persistent search feature.

jeff-blaisdell:
* Fix a javascript error when invoking persisting search multiple times.
* Fix issue with search cursor not being visible on screen.

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -132,7 +132,7 @@
       if (!cursor.find(rev)) return;
     }
     cm.setSelection(cursor.from(), cursor.to());
-    cm.scrollIntoView({from: cursor.from(), to: cursor.to()});
+    cm.scrollIntoView({from: cursor.from(), to: cursor.to()}, 20);
     state.posFrom = cursor.from(); state.posTo = cursor.to();
   });}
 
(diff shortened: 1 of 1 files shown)


addon/search/search.js
```diff
@@ -141,6 +141,7 @@
     state.lastQuery = state.query;
     if (!state.query) return;
     state.query = null;
+    state.queryText = null;
```
jeff-blaisdell: queryText needs to be reset as well.  otherwise on a second search by user the following JS error is thrown

Uncaught TypeError: Cannot read property 'global' of nullSearchCursor 
@ searchcursor.js:27(anonymous function) 
@ searchcursor.js:173getSearchCursor 
@ search.js:57(anonymous function) 
@ search.js:129runInOp 
@ codemirror.js:3126CodeMirror.operation 
@ codemirror.js:5209findNext 
@ search.js:127(anonymous function) 
@ search.js:114(anonymous function) 
@ dialog.js:78

addon/search/search.js
```diff
@@ -132,7 +132,7 @@
       if (!cursor.find(rev)) return;
     }
     cm.setSelection(cursor.from(), cursor.to());
-    cm.scrollIntoView({from: cursor.from(), to: cursor.to()});
+    cm.scrollIntoView({from: cursor.from(), to: cursor.to()}, 20);
```
jeff-blaisdell: Without the margin, if a search term is on an early row i.e. Line 1 / 2, after the search cycles back up to the top from the bottom, the highlighted term is off the screen.  Here's a screenshot:

screen shot 2015-07-14 at 10 40 33 am

marijnh:
Thanks! Merged as 01f2a1b

PR #3393, opened 2015-07-16
update rust primitive types

rrandom:
add new types usize and isize.
delete deprecated types int, uint and float.
https://doc.rust-lang.org/nightly/std/index.html#primitives

### DIFF
--- a/mode/rust/rust.js
+++ b/mode/rust/rust.js
@@ -26,7 +26,7 @@ CodeMirror.defineMode("rust", function() {
   };
   var typeKeywords = function() {
     var keywords = {"fn": "fn", "block": "fn", "obj": "obj"};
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as da42989. But note that the Rust mode in general reflects a very ancient version of Rust (early 2012). The current grammar is much more regular and can probably be expressed with a much simpler mode. I'm still waiting for the team to publish a formal grammar before I rewrite it, since I haven't touched it much since then, but if someone who's familiar with the language wants to write a [simple mode](http://codemirror.net/demo/simplemode.html) for the current language, that'd be very welcome.

rrandom:
Oh! Really? I am just reading your source code of rust.js and trying to modify it to be adapted to current version of Rust. since It can rewrite with simple mode, it may be simpler. I will try to figure it out. btw, the grammar doc is at:  https://doc.rust-lang.org/grammar.html  ,but it's incomplete.

Issue #1823, opened 2013-09-12
Korean input problem - Nothing shows until space or enter key is given.


minsu:
The problem can be reproducible on Safari/Chrome using any codemirror demo pages. I am using a mac book and I am not sure about other platforms.

 When searching for the problem, I saw one google group posting possibly regarding this same problem quite a long time ago but the english expression wasn't clear enough for you to understand the issue.
### problem

When typing in Korean, nothing will be reflected until space key or enter key is pressed, at that time all the accumulated inputs are reflected. Even the cursor won't move until space/enter key.

For example, if I key in `abc 강민수` in the editor, the editor will display:

```
abc
```

and then, nothing will happen until the last space key is pressed, when it will finally display:

```
abc 강민수 
```

You can reproduce the problem by selecting korean keyboard and try to input random keys(except numbers and punctuations) and then press space(or enter). Hope my description is clear enough. Thanks.

marijnh:
(If IME is important, consider setting the `inputStyle` option `"contenteditable"` in CodeMirror 5.0+. This should help, though it might introduce other issues since the contentEditable backend isn't all that well tested yet.)

flyskyko:
I have a problem with korean also. in my case is on IE.

contenteditable is still have a problem. it drops a character.
for example, I try input a '한국'.
composition start with 'ㅎ', and then '하', and then '한' and composition end.
'한' is show in editor. but next letter 'ㄱ' is dropped.

I found a solution when using a textarea.

```javascript
if (!cm.state.focused || /*(hasSelection(input) && !prevInput) ||*/
    isReadOnly(cm) || cm.options.disableInput || cm.state.keySeq)
  return false;
```

when korean is composition, IE make a selection on composition character.
so, I remove a one of condition. `(hasSelection(input) && !prevInput)`.

![screencast 2015-03-30 16-45-43](https://cloud.githubusercontent.com/assets/1835204/6892161/5a874768-d6fc-11e4-9398-b0aa1846b30a.gif)

adrianheine:
This is a issue that's difficult if not impossible to solve with the fundamental approach currently taken by CodeMirror.

We are working on a rewrite (CodeMirror 6) that will address this issue, and we are currently raising money for this work: See [the announcement](https://codemirror.net/6/) for more information about the rewrite and a demo.

Note that CodeMirror 6 is by no means stable or usable in production, yet. It's highly unlikely that we pick up this issue for CodeMirror 5, though.

PR #3394, opened 2015-07-17
dont exit poll() when ime composing

shiren:
related #1823 

we have problem when composing korean because poll() method keep exit when korean first letter composing in IE11

IE11 make selection when korean letter is composing(we need several key type to make one letter) 

I think this maybe solution

how about this?

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1398,7 +1398,7 @@
       // will be the case when there is a lot of text in the textarea,
       // in which case reading its value would be expensive.
       if (this.contextMenuPending || !cm.state.focused ||
-          (hasSelection(input) && !prevInput) ||
+          (hasSelection(input) && !prevInput && !this.composing) ||
           isReadOnly(cm) || cm.options.disableInput || cm.state.keySeq)
         return false;
 


marijnh:
Could you describe the problem that is occurring a bit more precisely?

shiren:
ok sorry about my english 

I'll show you with gif animation 

below is right situation 

good

when i type keyboard button "ㅇ"

we need poll() "ㅇ" and applyTextInput so editor should render "ㅇ"

then i type keyboard button "ㅏ"   

we need poll() "아" and applyTextinput so editor should render "아"

then i type keyboard button "ㄴ"   

we need poll() "안" and applyTextinput so editor should render "안"

so we can get complete one letter "안"

but **IE 11 / window7** dont show first letter so when we want to input "안녕"

cm dont render "안" until "녕" composition start

show below

bad

this because 

``` javascript
poll: function() {
      var cm = this.cm, input = this.textarea, prevInput = this.prevInput;
      // Since this is called a *lot*, try to bail out as cheaply as
      // possible when it is clear that nothing happened. hasSelection
      // will be the case when there is a lot of text in the textarea,
      // in which case reading its value would be expensive.
      if (this.contextMenuPending || !cm.state.focused ||
          (hasSelection(input) && !prevInput) ||
          isReadOnly(cm) || cm.options.disableInput || cm.state.keySeq)
        return false;
```

`(hasSelection(input) && !prevInput)` 

that condition cancel polling "ㅇ" -> "아" -> "안"

so we can't see "ㅇ" -> "아" -> "안"

i hope that this comment can help

marijnh:
That makes sense. Merged as 76582ee

Issue #3395, opened 2015-07-18
Preserve newline LF/CRLF for each line


mihailik:
Currently CodeMirror has a single newline character for the whole document.

That means `getValue` may return different text to what `setValue` was called with, depending on the newlines used.

Can the code preserve the ending newline for each line of the document, and use exactly that when concatenating line text for ranges, copy/paste, getValue/setValue please?

Ideally pressing 'Enter' would be made smarter too, so it insert the right kind of newline based on the one at the end of the current line.

marijnh:
Though I agree with your premise (it _would_ be nice if you could round-trip through CodeMirror without losing information), I don't think your solution is the right way to go.

The reason why this hasn't been much of a problem for people so far is probably that there are very few good reasons for mixing different line ending conventions in a single file. I am not really interested in supporting a system where CodeMirror magically keeps track of each line's ending and restores that.

Instead, I've implemented a different solution, which is much less likely to cause problems and is fully backwards compatible. Attached patches add a `lineSeparator` option, which defaults to null to give the current behavior (normalize to `\n`). You can set it to a string to have the editor break lines on exactly that string, and leave other characters as part of the line text. That way, your documents won't be damaged, but you also keep the invariant that an editor instance has a single line separator.

The intended use would be for those who care about line endings to implement some kind of heuristic that determines the line endings used in a given file, and pass that to the editor in the `lineSeparator` option.

mihailik:
Thanks a lot @marijnh.

Though I do **not** think `lineSeparator` would add value. It certainly does not help the use case I have.

My suggestion is to revert this change and leave it as is for now. It complicates code without solving any practical problem.

The use case I have is that user edits a file from external source, and upon save the only changes are those the user made.

I totally see it is not a trivial one to implement, so don't disagree with letting it rest until better opportunity (or until a friendly fork). But let's help that future work by not muddling the code with options almost nobody uses.

marijnh:
The patch is not going to be reverted. This is how I think being able to preserve mixed line endings should work.

mihailik:
No worries for not reverting, that's why I said 'suggestion' :)

But sorry, I might be missing the point. All the original line endings are removed from the document. How exactly do you intend to preserve *mixed line endings* with an extra parameter?

marijnh:
[...]

PR #3396, opened 2015-07-18
Preserve newline LF/CRLF for each line

mihailik:
Implements #3395 

The idea is to store EOL characters inside of `CodeMirror.Line.text`. And four broad areas needed careful adjustments:

1. Operations like getValue, getRange and similar will no longer `join("\n")`, instead they `join("")`. Of course, it's more complicated -- when `lineSep` optional argument is provided, the native EOLs need to be replaced with a given `lineSep`. That logic is in the new function `joinSep(lines, lineSep)`.

2. Cursor movement left-right need to skip over now-explicitly-present EOLs when moving in either direction. Given that EOLs can be either of `\n`, `\r\n` and `\r`, a new function `getLineSep(lineText)` is introduced. It's a cheap call, not allocating any objects and simply checking the last two characters. Several places where lineObj.text.length is used have been adjusted to account for the EOL length.

3. Highlight modes may not be prepared for EOLs, so StringStream is adjusted to skip over those. Again, care has been taken to avoid creating string objects unnecessarily.

4. Position/index conversion functions previously calculated offsets by aggregating over `line.text.length+1` where `+1` is for the EOL character. The new logic should just aggregate over `line.text.length`.

I've tested it with `/codemirror/demo/preview.html` -- it works exactly as before (at least in Chrome on desktop). Will try with more use cases.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -331,7 +331,7 @@
   // other lines onto it.
   function lineLength(line) {
     if (line.height == 0) return 0;
-    var len = line.text.length, merged, cur = line;
+    var len = line.text.length - getLineSep(line.text).length, merged, cur = line;
     while (merged = collapsedSpanAtStart(cur)) {
       var found = merged.find(0, true);
       cur = found.from.line;
@@ -340,9 +340,9 @@
     cur = line;
     while (merged = collapsedSpanAtEnd(cur)) {
       var found = merged.find(0, true);
-      len -= cur.text.length - found.from.ch;
+      len -= cur.text.length - getLineSep(cur.text).length - found.from.ch;
       cur = found.to.line;
-      len += cur.text.length - found.to.ch;
+      len += cur.text.length - getLineSep(cur.text).length - found.to.ch;
     }
     return len;
   }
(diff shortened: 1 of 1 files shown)


mihailik:
The build in Travis stailled, no idea why.

It did produce 3 small warnings, which I've fixed now.

vincentwoo:
This seems like a fairly dangerous change. I think CM runs off the assumption that newlines are `\n` pretty widely throughout the codebase, addons and such included.

Would it be easier to achieve what you want by filtering the input to `setText` on your end before passing it to CM?

mihailik:
It is indeed a significant change, but handling EOL correctly is vital in many scenarios.

Imagine you open a big file and change one line -- you don't want the whole file suddenly be reformatted with tabs replaced with spaces, CRLFs with LFs and so on. Imagine how difficult it would be to read this very pull request, if my editor decided to swap one whitespace convention for another :)

That said, I don't understand what you mean by filtering. Would you elaborate please?

vincentwoo:
> Imagine you open a big file and change one line -- you don't want the whole file suddenly be reformatted with tabs replaced with spaces, CRLFs with LFs and so on.

I might and I might not - it's more likely that users of CM have gotten used to a particular convention already, which is what I meant by dangerous.

> That said, I don't understand what you mean by filtering. Would you elaborate please?

In your linked issue, you said:

> That means getValue may return different text to what setValue was called with, depending on the newlines used.

I meant that, if this is in fact your core issue, you can standardize the newline situation in userland via something like

`editor.setText(text.replace(...))`

mihailik:
I don't see what you're saying at all, I'm sorry. Existing users already using LF convention would have their LFs exactly the same way. The whole point is to *preserve* the EOLs as is, whatever they are.

Liking your suggestion on 'userland' solution a lot, but sadly neither Bill Gates nor Linus Torvalds seem to be over-excited on the idea. Unless you manage to convince them, let's assume the input files may come with arbitrary convention.

marijnh:
I agree with Vincent that this isn't a good direction to go in. See my comment in #3395

PR #3397, opened 2015-07-19
update line numbers width

AnatoliyLitinskiy:
update line numbers width if editor was initialized in shadow (was not visible)

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -589,7 +589,8 @@
   function maybeUpdateLineNumberWidth(cm) {
     if (!cm.options.lineNumbers) return false;
     var doc = cm.doc, last = lineNumberFor(cm.options, doc.first + doc.size - 1), display = cm.display;
-    if (last.length != display.lineNumChars) {
+    var lineGutterWidth = parseInt(display.lineGutter.style.width);
+    if (last.length != display.lineNumChars || lineGutterWidth === 1) {
       var test = display.measure.appendChild(elt("div", [elt("div", last)],
                                                  "CodeMirror-linenumber CodeMirror-gutter-elt"));
       var innerW = test.firstChild.offsetWidth, padding = test.offsetWidth - innerW;


marijnh:
You can't depend on the gutter being 1px wide by default. In any case, if the editor wasn't visible, doesn't calling `refresh` on it after making it visible already take care of this?

AnatolLitnskiy:
Hi, Marijn.

You shouldn't believe to value:
> var innerW = test.firstChild.offsetWidth

if it has default 1px val

> Math.max(innerW, display.lineGutter.offsetWidth - padding) + 1;

in fact 0, but you have added 1;

I can't found a method to recount linesWidth any more in your code.

So it looks like the single method that can update lineNumberWidth value finish his work before recounting this value, and doesn't check that lineNumberWidth has (bad) default value.

The condition to recount lineNumberWidth looks like:

> last.length != display.lineNumChars

And it will not occurs if we just enter new line by pressing 'Enter' key.
And I can't found situation when this condition can occur at all.

I just try to solve problem that is described on huge of sites in internet, when Codemirror have used default value 1px for lineWidth if editor was not visible and browser can't count width of <div>
> test = display.measure.appendChild(elt("div", [elt("div", last)]

My solution works, yes it is not the best architecture design solution, but architecture design have to take into account errors, that can occur, and it skips error with width in method 'maybeUpdateLineNumberWidth'

> var test = display.measure.appendChild(elt("div", [elt("div", last)];
> var innerW = test.firstChild.offsetWidth;

This fix allow CodeMirror to updates lineNumberWidth by calling refresh, it looks like solution of problem. You can use it as hint.

f.e. You can mark, that width was count in bad way and add it to condition:

> last.length != display.lineNumChars || lastLineWIdthCountError

Thanks, Anatoliy

marijnh:
So _are_ you calling refresh after making the editor visible? If so, can you create a simple HTML page that shows the problem (the line number width not being accurate after refresh)

AnatoliyLitinskiy:
Hm, in simple HTML refresh() solves this problem.
http://jsfiddle.net/alitinskiy/3Lew35t0/
Even more in my project now it work too! (may be after bower update) Sorry for that.

PR #3398, opened 2015-07-20
[vim] Make it easier to override color on blinking cursor

vincentwoo:
Right now, if you want to change the color of the cursor in vim mode, you need to both

1. Modify the background color of `.CodeMirror.cm-fat-cursor div.CodeMirror-cursor`
2. Reassign the vendored animation properties of `.cm-animate-fat-cursor` to an animation with a new custom color.

This change makes the latter easier, by just animating to transparent and using whatever the default background color of `.cm-animate-fat-cursor` is, which is much easier to override.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -63,21 +63,22 @@
   -webkit-animation: blink 1.06s steps(1) infinite;
   -moz-animation: blink 1.06s steps(1) infinite;
   animation: blink 1.06s steps(1) infinite;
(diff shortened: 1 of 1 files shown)


vincentwoo:
cc @mightyguava, who I think is the "vim guy"

marijnh:
When it's just CSS, I'm also happy to review vim stuff. Looks great. Merged as a84b756

PR #3399, opened 2015-07-20
Create CHANGELOG

WesCossick:
GitHub is an excellent and expected place to store the changes in each release. But if not, you should at least link to where it can be found on your website.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,1 @@
+Located at: http://codemirror.net/doc/releases.html


marijnh:
The release list _is_ on github -- that html file is part of the repository. I don't really see the value of adding another file just to link to it. What practical problem would it solve?

WesCossick:
The problem is that GitHub is where users expect to find a change log for each version; after all, GitHub is a version control system that tracks changes. The best way to do that is looking at the 'Releases' tab, a handy feature GitHub has built for this exact reason. This tab is used when new releases are alerted via services like Sibbell. Unfortunately, not all repositories use this feature, but that means they use a CHANGELOG file. This is where users can go to view the changes for each release and is a bit less useful than the releases tab. The problem is, CodeMirror uses neither. So now as a user, I'm wanting to find the changes for each release, and I've exhausted all my standard options on GitHub and have found nothing. I will probably not dig around in the repository until I find that HTML file you just mentioned, and even if I do, reading HTML code on GitHub is not anywhere near as friendly as reading a well formatted markdown document.

So now I'm a user trying to find the change log, and I've hit a dead end. If I'm determined enough, I'll probably search Google for "CodeMirror change log" and I finally find the releases on your own website, not GitHub the version control system.

My pull request solves this scenario and saves wasted time for users. A user searching for the change log will go to GitHub, see the CHANGELOG file, open it, and know exactly where to go.

marijnh:
I don't think this problem (having to look at the project page to find the change log) is big enough to warrant a file like that, sorry.

vincentwoo:
FWIW I think this is actually a pretty reasonable request. A one-time tiny file addition that can alleviate the confusion of maybe 10% new users seems like a very friendly, worthwhile change to have. I have definitely googled "codemirror changelog" a fair number of times.

Minopolis:
I would love to see Codemirror use Github's releases feature as I like to keep track of my project's dependencies' updates through RSS feeds which codemirror's project page doesn't feature. I also agree that it would be more user-friendly.

marijnh:
Could you link a repository that you think uses github's release feature properly? It looks to me like they just list tags under 'releases' -- I'm not sure what feature you are referring to.

vincentwoo:
In the Firepad repo we have straightforward changes + download links in the releases tab: https://github.com/firebase/firepad/releases

marijnh:
Ah, there's an 'add release notes' control if you switch to the 'tags' page from the 'releases' page. I hadn't seen that yet. I'll try to make some time this week to upload notes for all halfway recent releases, and try to automate pushing them out for future releases.

WesCossick:
Thanks, I really think using the releases feature will be a great improvement. Especially for those of us using Sibbell and/or RSS feeds for keeping up-to-date with new releases on the dozens of libraries we rely on.

Also, for reference, I think [SimpleMDE](https://github.com/NextStepWebs/simplemde-markdown-editor/releases) does a great job of explaining each release in terms of New, Improved, and Fixed. But then again, I may be a bit biased since I'm the one who writes the releases.

marijnh:
Okay, all releases have their notes attached in github's database now

vincentwoo:
Beautiful, thank you.

PR #3401, opened 2015-07-20
Avoid passing array as 'this' argument to operation callbacks

mihailik:
Small suggestion to avoid possible bugs related to unintended array value captured in 'this' argument.


marijnh:
Thanks. I've pushed a slightly different patch (8bb0736) which avoids the comma trick and uses `call` instead.

mihailik:
:+1:

PR #3402, opened 2015-07-20
Allow users to define their own lint event instead of just 'change' event.

amshali:
This patch changes the codemirror's lint addon to allow users to define an event on which lint should be done. Instead of the hard-coded change event, a user can define their own event and trigger that in order to get lint messages.


marijnh:
I'd prefer a different approach -- allow users to pass an option to the lint addon to disable automatically triggering linting on change, and export (via defineExtension) a method to explicitly trigger linting from user code.

amshali:
OK, I will send a patch.

PR #3405, opened 2015-07-22
Remove duplicate color hex sign

sheabunge:
The CSS for the Yeti theme is invalid - there is a duplicate # sign for a hex color on line 62.


marijnh:
Thanks. Merged as d002b03

PR #3406, opened 2015-07-23
rewrite rust.js with defineSimpleMode.

rrandom:
I am not very sure which token should be assigned to macro, so I choose variable-3.


marijnh:
Hi. Thanks for working on this. You seem to be missing some things though (attributes jump out). Might want to go over the whole grammar spec as far as it exists, and make sure no other constructs are missing. `variable-3` is a good choice for macros -- most themes give it a distinctive bright color.

One thing I was hoping to do when a full grammar is available is to write a detailed enough state machine to distinguish types from values, and highlight them accordingly. But even without that, this is a step up from the old obsolete mode. Could you add attributes and double check the grammar to see if you missed anything else? (And add a comment when you have, so that I can review again.) Thanks!

rrandom:
OK! I will do it later

marijnh:
(Be sure to leave a comment when you want me to review this again.)

rrandom:
Yes, I think I can finish it within the next week

rrandom:
Hi, @marijnh, you may review this file now. But I did not find any code folding option in defineSimpleMode. Is there one?

marijnh:
Thank you. Merged as d7dc8c31b99 and followed up with a few tweaks in d56b13a17336

PR #3407, opened 2015-07-24
New CodeMirror option for tern

mg80:
The tern tooltip hides after 1700 milliseconds.  Many tooltips display a lot of content.  Adding a CodeMirror option for controlling overriding the 1700 milliseconds.

### DIFF
--- a/addon/tern/tern.js
+++ b/addon/tern/tern.js
@@ -616,7 +616,7 @@
         else mouseOnTip = false;
       }
     });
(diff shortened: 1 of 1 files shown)


marijnh:
It is preferable to pass addon options to the addons themselves, rather than putting them on the editor themselves. The Tern addon already takes an options object, which seems like the ideal place to put this.

mg80:
Created a new pull request

PR #3411, opened 2015-07-24
[Dockerfile mode] Added lineComment

TDaglis:
Added lineComment in Dockerfile mode.


marijnh:
Thanks. Merged as ea3277b

PR #3413, opened 2015-07-25
Adding tern option for the hint delay

mg80:
The tern tooltip hides after 1700 milliseconds. Many tooltips display a lot of content. Adding a tern option for controlling overriding the 1700 milliseconds.


marijnh:
Looks great. Merged as b6f9fd3

PR #3415, opened 2015-07-27
This patch allows users explicitly trigger the linting.

amshali:
This patch allows users to pass an option to the lint addon to disable automatically triggering linting on change, and export (via defineExtension) a method to explicitly trigger linting from user code.


marijnh:
I've merged this as 536bba1 and modified it in 3686816 to:

 - Rename the option to `lintOnChange` (to avoid double negatives) -- must be `false` to disable.
 - Make the extension method use its `this` binding, rather than also expecting the editor to be passed as an argument.

amshali:
Thanks a lot!

Issue #3251, opened 2015-05-08
Allow CodeMirror to execute in a different window than the GUI is drawing in


bedney:
I have a use case whereby CM's code is executing in the top-level window, but it is being handed an Element to create the editor and that Element is in an iframe.

There are hardcoded references to 'document.' and 'window.' in CM's codebase that cause this approach to have problems. I went into the base file for CM and introduced a new variable ('placeWin'), set it to the window of the element that gets handed in to create the editor (placeWin = <elem>.ownerDocument.defaultView) and replaced all of the 'window.' references with 'placeWin.' and 'document.' references with 'placeWin.document.'.

This seems to solve the problem - at least for the base lib (I haven't looked through any of the plugins).

@marijnh - if I wrapped this up into a formal PR and submitted it, is it something you would be interested in? I can't move forward without it, but I'd rather not maintain my own fork of CM.

Thanks!

marijnh:
This has come up before. I think a better approach is to have a window associated with each editor instance, so that you can have editors open in different frames at the same time.

bedney:
Yea, this is how I used CM for years and it worked well. Multiple editors, each in their own iframe, no problem.

Unfortunately, the app I'm currently building is more sophisticated and requires me to be able to move the editor around in the DOM (not visually move it - reparent it within the hierarchy). Whereupon I ran smack into these bugs:

Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=254144
Webkit: https://bugs.webkit.org/show_bug.cgi?id=13574

As you can see, these are old bugs and are marked as invalid/wontfix. So I don't forsee a future where the browser vendors are going to support moving an \<iframe> around in the DOM.

This does bring up an interesting point, though: does CM have other shared/global state that it is counting on such that it only anticipates having one instance per window? I've actually have multiple instances running in the same window with this patch and it seems to work ok - I haven't run into any issues so far, but maybe I'm getting myself into trouble here and don't know it yet :-).

Thanks for taking a look at this and thanks so much for CodeMirror. I have worked in browsers for 15+ years and I can tell you that it is a most impressive piece of software.

marijnh:
So the question is why are you putting the editor in an iframe in the first place? Of course you have multiple CodeMirror instances in a single window.

bedney:
Marjin -

Sorry, I actually am conflating two scenarios here. Before I explain, let me state that both of these are Single Page App scenarios.

[...]

PR #3419, opened 2015-07-28
Added patches to make window and document indirect

bedney:
@marijnh -

I created this PR for #3251 as I needed it right away. It's patched against 5.5.0.

Let me know what you think.

Cheers,

- Bill

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -51,6 +51,8 @@
   // Optimize some code when these features are not used.
   var sawReadOnlySpans = false, sawCollapsedSpans = false;
 
+  var placeWin;
+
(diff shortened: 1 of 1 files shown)


marijnh:
As I mentioned in the other discussion, I'd want windows to be associated with editors, not global. Also, the tests are failing. So this may be useful to someone, but it's not going into the distribution the way it is.

bedney:
@marijnh -

It meets my needs for now while awaiting the official patch and communicates my needs.

It's curious that that the tests seem fail on the statement of obtaining the window reference via `place.ownerDocument.defaultView`, which is a standard DOM way of obtaining the window from a supplied Element.

Cheers,

- Bill

PR #3423, opened 2015-07-30
Fixed the bug with testing lintOnChange

amshali:
Fixed the bug with testing lintOnChange. The current way of testing this parameter will result in lintting on change even when one sets it to false.

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -161,7 +161,7 @@
     if (options.onUpdateLinting) options.onUpdateLinting(annotationsNotSorted, annotations, cm);
   }
 
-  function onLintEvent(cm) {
+  function onChange(cm) {
     var state = cm.state.lint;
     if (!state) return;
     clearTimeout(state.timeout);
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -161,7 +161,7 @@
     if (options.onUpdateLinting) options.onUpdateLinting(annotationsNotSorted, annotations, cm);
   }
 
-  function onChange(cm) {
+  function onLintEvent(cm) {
     var state = cm.state.lint;
     if (!state) return;
     clearTimeout(state.timeout);
@@ -187,7 +187,7 @@
   CodeMirror.defineOption("lint", false, function(cm, val, old) {
     if (old && old != CodeMirror.Init) {
       clearMarks(cm);
-      if (!state.options.lintOnChange !== false)
+      if (cm.state.lint.options.lintOnChange === undefined || cm.state.lint.options.lintOnChange === true)
         cm.off("change", onChange);
       CodeMirror.off(cm.getWrapperElement(), "mouseover", cm.state.lint.onMouseOver);
       clearTimeout(cm.state.lint.timeout);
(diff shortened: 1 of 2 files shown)


addon/lint/lint.js
```diff
@@ -187,7 +187,7 @@
   CodeMirror.defineOption("lint", false, function(cm, val, old) {
     if (old && old != CodeMirror.Init) {
       clearMarks(cm);
-      if (!state.options.lintOnChange !== false)
+      if (cm.state.lint.options.lintOnChange === undefined || cm.state.lint.options.lintOnChange === true)
```
marijnh: Why are you complicating the test here, though?

demo/lint.html
```diff
@@ -165,7 +165,7 @@
     mode: "css",
     gutters: ["CodeMirror-lint-markers"],
     lint: {
-      disableLintOnChange: true
+      lintOnChange: false
```
marijnh: I would think we want to keep the automatic linting in the demo.

addon/lint/lint.js
amshali: COND = !state.options.lintOnChange !== false
truth table:
lintOnChange: undefined. Should lint on change? YES. COND is true. OK.
lintOnChange: true. Should lint on change? YES. COND is false. ERR.
lintOnChange: false. Should lint on change? NO. COND is true. ERR.

Am I right?

demo/lint.html
amshali: OK, I'll fix that. Although this is meant to present the use of performLint in the following lines which performs lint every 5 seconds. Do you still think I should remove it?

addon/lint/lint.js
marijnh: No, I don't think you are right. `true !== false` is true. With the `!== false` condition you only turn of the change event handler when `lintOnChange` is explicitly set to `false`, which is what we want.

amshali: I was very confused as a programmer so I wrote this program:
```javascript
a = {};
// lintOnChange undefined
console.log(!a.lintOnChange !== false);
a.lintOnChange = true;
console.log(!a.lintOnChange !== false);
a.lintOnChange = false;
console.log(!a.lintOnChange !== false);
```
which generates this:
```
true
false
true
```

Am I missing anything?

marijnh: > Am I missing anything?

Yes, you are. The precedence of `!` is higher than that of `!==`, so you're comparing `(!a.lintOnChange)` to `false`.

amshali: I am not making this up. The (current) following code does not do what it is supposed to do. You can try it yourself. 
```
      if (!state.options.lintOnChange !== false)
        cm.off("change", onChange);
```
I know about precedence of operators. The correct condition is really:
```
cm.state.lint.options.lintOnChange === undefined || cm.state.lint.options.lintOnChange === true
```

marijnh: No, it is `state.options.lintOnChange !== false`. See 38c3e9c for the correct patch.

amshali: Or that. Thank you.

PR #3424, opened 2015-07-31
[stylus mode] Fix highlight class after a $var

dmitrykiselyov:
For example:

body
  color $red

.content
  color #000

The .content was highlighted as the variable-2


marijnh:
Thanks. Merged as 7523531

dmitrykiselyov:
Thanks you!

PR #3427, opened 2015-08-02
Standardize theme CSS

vincentwoo:
The formatting/design of the CSS in the themes directory is a bit all over the place, so I tried to tidy it up. I broke this change up into 4 commits, the first two being less controversial than the second two.

1. Four themes (seti, yeti, dracula, and material) use three-line css definitions where every other theme uses one-liners. I switched those four to use one-line definitions as well.
2. A fair amount of themes use `!important` in their `.CodeMirror-selected` definitions. Since https://github.com/codemirror/CodeMirror/commit/e1e9aca0c7c90695d600b0cf38e2798f2de9f83d got checked in, this is unnecessary (and problematic for users to override). Now, all themes that specify a rule for `.CodeMirror-selected` do so with `.themename div.CodeMirror-selected` and no `!important` clause.
3. Some themes have definitions like `.cm-s-material span.cm-meta{ color: #80CBC4; }` where there is no space between `meta` and the opening brace. This commit adds spaces there.
4. This is probably the most contentious change. The majority of one-line css definitions look like `.selector { attr: value; }`, that is, with one space after the selector, one after the opening brace, and one before the closing brace. However, there are a lot of nonstandard one-liners, the most common form being something like `.selector {attr:value}`. This commit ensures that all the definitions have the same spacing and that a semicolon is present.

Feel free to pick and choose if you don't want all of this! The last commit definitely touches a lot of lines, so I can understand not wanting to dirty up the blame log for essentially a cosmetic change.

### DIFF
--- a/theme/3024-day.css
+++ b/theme/3024-day.css
@@ -8,34 +8,34 @@
 
 */
 
-.cm-s-3024-day.CodeMirror {background: #f7f7f7; color: #3a3432;}
-.cm-s-3024-day div.CodeMirror-selected {background: #d6d5d4;}
+.cm-s-3024-day.CodeMirror { background: #f7f7f7; color: #3a3432; }
+.cm-s-3024-day div.CodeMirror-selected { background: #d6d5d4; }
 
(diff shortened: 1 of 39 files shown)


marijnh:
Thanks for doing this boring work! I had plans to do something about all the `!important` annotations, but hadn't gotten around to it. Very glad you did it. Whitespace inconsistency doesn't bother me much, but since you already went through the work of fixing it, I've merged those patches too.

PR #3433, opened 2015-08-04
Fixed the issue with undefined state.

amshali:


marijnh:
Thanks. Merged as 6143625

PR #3434, opened 2015-08-04
Reduce css specificity for cursor and activeline classes

vincentwoo:
Also some general improvements to material.css and solarized.css

Lines like https://github.com/codemirror/CodeMirror/compare/master...vincentwoo:master#diff-33646e9ee4d04eedb98a991a636c2302L26 can get pretty dangerous - I think cases like this are a good reason to advocate for using SASS or any kind of preprocessor step to standardize theme CSS.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -41,14 +41,17 @@
 
 /* CURSOR */
 
-.CodeMirror div.CodeMirror-cursor {
+.CodeMirror-cursor {
+  position: absolute;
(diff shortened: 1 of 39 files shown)


marijnh:
Thanks. Merged as a7d443ec82c and updated in f06c53f (the CSS file intentionally splits user-overridable styles from internal styles, so the fact that a rule was duplicated and that there was an empty rule were not bugs)

PR #3439, opened 2015-08-07
add values for property flex-direction

rrandom:
http://www.w3.org/TR/css-flexbox-1/#property-index
 I found some other values in flex properties are missing. Should I add them?


marijnh:
Wonderful. Merged as 9f1b3a6

PR #3441, opened 2015-08-07
Allow initial custom list through hintOptions

allensarkisyan:
- Provide custom list through CodeMirror hintOptions to kickoff autocompletion for domain specific usage.


marijnh:
Looks like a good idea. Merged as 5d782cb

PR #3442, opened 2015-08-08
add all missing values for properties in flex

rrandom:
http://www.w3.org/TR/css-flexbox-1/#property-index


marijnh:
Thanks! Merged as 341a850

PR #3459, opened 2015-08-18
HTML Lint Integration

mikethedj4:
Codemirror comes with css-lint, js-lint, and json-lint, and even coffeescript-lint. So I decided to integrate html-lint. 

I tested this myself and added it on Github - https://github.com/mikethedj4/html-lint-for-codemirror 

This however is my first commit to another git repository.


marijnh:
Thanks. I've merged this as f1ffc95f43a0b2 and fixed it up a little in 644803cd61b008 (style issues, and allowing the user to pass in the configuration using the `rules` property of the `lint` option, rather than global).

PR #3462, opened 2015-08-20
Added vue.js single-file component support

karevn:
It looks like Vue.js is cool enough to deserve some attention. I added a first iteration of .vue templates support.

### DIFF
--- a/mode/htmlmixed/htmlmixed.js
+++ b/mode/htmlmixed/htmlmixed.js
@@ -10,112 +10,203 @@
     mod(CodeMirror);
 })(function(CodeMirror) {
 "use strict";
+var nestedModes = {
+    script: {
+      attributes: {
+        lang: {
+          javascript: /(javascript|babel)/i
+        },
+        type: {
+          javascript: /^(?:text|application)\/(?:x-)?(?:java|ecma)script$|^$/i
+        }
+      },
+      defaultMode: 'javascript'
+    },
+    style:  {
+      attributes: {
+        lang: {
+          css: /^css$/i
+        },
+        type: {
+          css: /^(text\/)?(x-)?stylesheet$/i
+        }
+      },
+      defaultMode: 'css'
+    }
+  }, attrRegexpCache = {}, tagRegexpCache = {};
 
-CodeMirror.defineMode("htmlmixed", function(config, parserConfig) {
-  var htmlMode = CodeMirror.getMode(config, {name: "xml",
-                                             htmlMode: true,
-                                             multilineTagIndentFactor: parserConfig.multilineTagIndentFactor,
-                                             multilineTagIndentPastTag: parserConfig.multilineTagIndentPastTag});
-  var cssMode = CodeMirror.getMode(config, "css");
+  function deepmerge(target, src) {
+    var array = Array.isArray(src);
+    var dst = array && [] || {};
 
-  var scriptTypes = [], scriptTypesConf = parserConfig && parserConfig.scriptTypes;
-  scriptTypes.push({matches: /^(?:text|application)\/(?:x-)?(?:java|ecma)script$|^$/i,
-                    mode: CodeMirror.getMode(config, "javascript")});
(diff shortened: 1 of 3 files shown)


mode/vue/vue.js
```diff
@@ -0,0 +1,193 @@
...
+            "../sass/sass",
+            "../stylus/stylus"], mod);
+  } else { // Plain browser env
+    mod(CodeMirror);
+  }
+})(function (CodeMirror) {
+
+
+  CodeMirror.defineMode("vue-template", function (config, parserConfig) {
+    var mustacheOverlay = {
```
marijnh: Isn't Mustache compatible with Handlebars? Would it make sense to use the more advanced Handlebars mode as a dependency instead?

mode/vue/vue.js
```diff
@@ -0,0 +1,193 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function (mod) {
+  "use strict";
```
marijnh: You want to put the `"use strict"` in the main body — here it'll only affect the module header.

mode/vue/index.html
```diff
@@ -0,0 +1,88 @@
+<!doctype html>
```
marijnh: This appears to be a copy-pasted versin of the HTML mixed mode demo. That doesn't really show what this mode can do, does it?

mode/vue/vue.js
karevn: Actually, base Vue templates are in just its own format, and the closest one is handlebars (that's why I copy-pasted it from the example). But it's still a different format.

mode/vue/index.html
karevn: You're right, there's some space for improvement :)

marijnh: So the examples of single-file vue apps I found online seem to use `<script>` tags to delimit parts of the document. Since the `htmlmixed` mode already allows you to customize the way the content of different types of script tags get parsed, I'm not convinced that this mode adds a lot.

karevn: @marijnh the question is just if .vue files will be highlighted correctly. But... After your comment I started to think that vue mode should be just some wrapper over a refactored htmlmixed. What do you think?

karevn:
I think this code is ready to replace htmlmixed (after some adjustments). Ideas?

karevn:
@marijnh can you please review my merge for Vue and htmlmixed? Even if you don't like an idea of .vue,  I think new htmlmixed mode code is much better structured and flexible (and may be a little bit faster).

mode/htmlmixed/htmlmixed.js
```diff
@@ -10,112 +10,203 @@
...
+    var cur = stream.current(), close = cur.search(pat);
+    if (close > -1) {
+      stream.backUp(cur.length - close);
+    } else if (cur.match(/<\/?$/)) {
       stream.backUp(cur.length);
-      if (!stream.match(pat, false)) stream.match(cur);
+      if (!stream.match(pat, false)) {
+        stream.match(cur);
+      }
     }
```
marijnh: Reviewing this now. Just a general point: I really have better things to do than review randomly inserted unneeded braces in already existing perfectly fine code. Don't muddle up your patches with that stuff.

marijnh:
I've merged this as bfbb9c716d1b3 and largely rewrote it in eb56e911191.

Things that were broken:

 - Detecting close tags inside tokens didn't work anymore because you were passing a regexp that started with `^`.

 - People could no longer pass non-string mode specs in your object format.

 - The dependency on `addon/mode/overlay` wasn't declared.

 - The demo html page was still a mess with the text from the htmlmixed demo in it.

 - Attributes other than `lang` weren't recognized, since you passed the hard-coded string `"lang"` to `getAttrValue`.

My patch changes the format to arrays-of-arrays, removing the need for the huge `deepmerge` function, and fixes these bugs.

karevn:
@marijnh whatever, man, if it will make this software better :) I'm a newbie in CodeMirror world.

PR #3464, opened 2015-08-21
[sublime] Make smartBackspace delete and not deindent

### LINKED ISSUE #3449: indent error with Tab
when I input 2 tabs here, and hit del, the second tab will be removed but the 1st one will not be removed.

vincentwoo:
Fixes #3449 

Basically, what's currently happening in sublime mode is that if you hit backspace and we think we can shave off a full indent, we tell CM to deindent.

This is fine if your cursor is right before the text that's getting shifted left (in this case, your cursor moves with the text), but if your cursor is surrounded by whitespace to the left and right, it stays in place.

This change makes `smartBackspace` behave more like backspace and less like a "deindenter". It figures out what text should be deleted to the immediate left of the cursor and then deletes it.

I also opened up the helper function `findColumn` which is super helpful for doing any kind of tab math. Is that acceptable?

### DIFF
--- a/demo/sublime.html
+++ b/demo/sublime.html
@@ -69,7 +69,8 @@ option to <code>"sublime"</code>.</p>
     autoCloseBrackets: true,
     matchBrackets: true,
     showCursorWhenSelecting: true,
-    theme: "monokai"
+    theme: "monokai",
+    tabSize: 2
   });
 </script>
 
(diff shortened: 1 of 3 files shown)


marijnh:
Thank you. Merged as 7f51f13

PR #3467, opened 2015-08-22
Adjust CSS indent to align with parent context when closing blocks

dbarnett:
Fixes the CSS indentation for closing braces to align with the outer block scope instead of just subtracting an indent level. The difference is only noticeable on especially complex formatting like the "css_document" test.

I noticed this issue while working on a mode for Closure Templates a.k.a. GSS where `@component` provides some unique indentation cases.


marijnh:
Thanks. Merged as 3d18f89

PR #3468, opened 2015-08-23
Advertise gfm mode from markdown's demo page.

cben:
More than once people have asked questions expecting markdown mode to support GFisms and were simply not aware of gfm mode existance. Hopefully this will make it more discoverable.

Page after the patch: https://rawgit.com/cben/CodeMirror/cben-patch-1/mode/markdown/

### DIFF
--- a/mode/markdown/index.html
+++ b/mode/markdown/index.html
@@ -350,8 +350,10 @@ Output:
       });
     </script>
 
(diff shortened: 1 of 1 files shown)


marijnh:
That's a very good idea. Merged as 13eadcd

PR #3471, opened 2015-08-23
Add mode for "Closure Stylesheets (GSS)"

### LINKED ISSUE #3470: Implement mode for Closure Stylesheets (GSS)
CodeMirror should have a mode for Closure Stylesheets syntax, an extension of CSS syntax (https://code.google.com/p/closure-stylesheets/).

The most obvious difference from standard CSS is that adding an `@component` block around the whole file breaks most syntax features.

dbarnett:
The only difference from standard CSS so far is that it parses syntax properly inside `@component` blocks and also doesn't indent them.

Fixes #3470.

### DIFF
--- a/mode/css/css.js
+++ b/mode/css/css.js
@@ -28,7 +28,9 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
       counterDescriptors = parserConfig.counterDescriptors || {},
       colorKeywords = parserConfig.colorKeywords || {},
       valueKeywords = parserConfig.valueKeywords || {},
-      allowNested = parserConfig.allowNested;
+      allowNested = parserConfig.allowNested,
+      supportsAtComponent = (typeof parserConfig.supportsAtComponent != "undefined" ?
+          parserConfig.supportsAtComponent : false);
 
   var type, override;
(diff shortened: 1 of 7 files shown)


marijnh:
Merged as e79d4ffafb3300 . Though I'm somewhat confused why you went to the effort of adding a test without making sure it actually passed. Fixed that in 5d853959ca8a

dbarnett:
Weird. I did some last-minute refactors, but thought I verified tests pass after those. Which ones failed? (Travis results look broken for a different reason.)

marijnh:
For one thing, the `getMode` call didn't actually name the mode correctly.

dbarnett:
Yeah, I moved the test file and changed the name and at least thought I re-ran tests. No idea about the other change preventing negative indent, though. Didn't see any failures at all and I think that one would have shown up before my last-minute changes.

Anyway, sorry for test failures. I'll do better next time. =)

dbarnett:
How often do you do version bumps? It looks like there hasn't been a new version since 5.6.1, before this change.

marijnh:
Once per month. 5.7 is coming out early next week.

PR #3479, opened 2015-08-26
experimental es7 async/await support

zaggino:
Hi guys,
we're using es7 experimental async/await features ( http://babeljs.io/docs/advanced/transformers/#es7-experimental- ) and it breaks syntax highlighting when using CodeMirror. Atom and VS Code handle our files quite well, so I've put together this naive fix for CodeMirror (used by Brackets). I'd appreciate any guidance, how to do this properly so it could be potentially merged upstream.


marijnh:
This looks good. Merged as 94fc3b7

PR #3484, opened 2015-08-26
Activeline background added

mattpass:
Didn't have this previously


marijnh:
Thanks. Merged as 756ddc5

PR #3487, opened 2015-08-28
Allow setting raw HTML as placeholder option

vincentwoo:
If you want to have more complicated or highlighted text in a CodeMirror placeholder, it can be difficult with the current addon.

This patch lets you at least pass HTML as the placeholder, which I think is a relatively safe change that gives a fair amount of power without additional complexity.

### DIFF
--- a/addon/display/placeholder.js
+++ b/addon/display/placeholder.js
@@ -37,7 +37,7 @@
     var elt = cm.state.placeholder = document.createElement("pre");
     elt.style.cssText = "height: 0; overflow: visible";
     elt.className = "CodeMirror-placeholder";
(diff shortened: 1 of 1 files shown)


marijnh:
Uh oh, you've just created an XSS attack vector (interpreting content that was previously raw text as HTML, and making it possible, on sites that allow users to influence the placeholder text somehow, to insert arbitrary HTML into the document).

So we can't change the meaning of string data here. In general, the CodeMirror API tends to allow passing in of DOM nodes, not raw HTML, to be passed in. Attached patch implements an alternative solution where `placeholder` may be a DOM node.

vincentwoo:
Thanks, works for me!

PR #3490, opened 2015-08-28
[oz mode] Add

anthonygego:
Hi,

Here's a mode for the Oz multi-paradigm programming language. Some info can be found at :
- https://en.wikipedia.org/wiki/Oz_(programming_language)
- http://mozart.github.io/

This language is essentially used for research and education. The mode will also be used from this September for the grading system of *Louv1.1x* and *Louv1.2x* MOOCs on edX.

Feedback is welcome :)

### DIFF
--- a/mode/oz/oz.js
+++ b/mode/oz/oz.js
@@ -1,0 +1,242 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"));
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror"], mod);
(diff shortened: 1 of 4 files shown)


mode/oz/oz.js
```diff
@@ -0,0 +1,242 @@
...
+      if (trueText.match(endKeywords) || trueText.match(middleKeywords) || trueText.match(/(\[])/))
+        return conf.indentUnit * (state.currentIndent - 1);
+
+      if (state.currentIndent < 0)
+        return 0;
+
+      return state.currentIndent * conf.indentUnit;
+    },
+    fold: "indent",
+    electricChars: "eEdDtTnNiIoOfFhHyY[]",
```
marijnh: Looks like you're trying to express something non-trivial here. You might want to use `electricInput` instead, which allows a regexp (which is matched starting from the start of the line) to reduce false positives.

marijnh:
Looks very good. Merged as 0cd2432. It'd be great if you could follow up with a PR to clean up the electric indentation test.

PR #3491, opened 2015-08-31
null check

wuliupo:
if the keywords are not resisted, there will be exception: Cannot read property 'length' of undefined.


marijnh:
Is there any hint source that returns objects without `list` array property? If so, I think we should fix that instead.

PR #3493, opened 2015-09-01
Update haxe.js

### LINKED ISSUE #3492: Haxe mode multi-line string indentation incorrect
double-checking via http://codemirror.net/mode/haxe/

haxe supports multi-line strings

    var s :String = "hello
        world";

but codemirror's haxe mode gets confused

    "
    	";
    		{
    		}

or, with single quotes

	'
		';
			{
			}

with both of these, the preferred result is something like

    "
    ";
    {
    }

or

    '
    ';
    {
    }

respectively.

increpare:
fixes #3492 

returns true only once it sees an unescaped quote of matching type, even if it has to skip newlines to do it.

I can't find formal specs saying that haxe supports them, but it does. (see http://try.haxe.org/#5Eb5E for instance (notice the formatting is broken there as well ) ).

### DIFF
--- a/mode/haxe/haxe.js
+++ b/mode/haxe/haxe.js
@@ -48,7 +48,7 @@ CodeMirror.defineMode("haxe", function(config, parserConfig) {
         return false;
       escaped = !escaped && next == "\\";
     }
-    return escaped;
+    return true;
   }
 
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Merged as a2cc9c0a55

PR #3496, opened 2015-09-03
[markdown] Support an option for tildes as code fences.  

brondsem:
And also allow more than 3 \`s or \~s

These are part of the [CommonMark spec for fenced code blocks](http://spec.commonmark.org/0.22/#fenced-code-blocks)

### DIFF
--- a/mode/gfm/gfm.js
+++ b/mode/gfm/gfm.js
@@ -112,7 +112,7 @@ CodeMirror.defineMode("gfm", function(config, modeConfig) {
   var markdownConfig = {
     underscoresBreakWords: false,
     taskLists: true,
-    fencedCodeBlocks: true,
+    fencedCodeBlocks: '```',
     strikethrough: true
   };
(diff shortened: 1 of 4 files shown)


marijnh:
CommonMark supports both styles interchangeably. I think that would be preferable behavior, and I'd be in favor of making that the default behavior, rather than requiring configuration. If that sounds good to you, could you update the pull request with an implementation?

brondsem:
That does make sense and I'll see what I can do.  I wonder if its ok for the gfm mode to support ~~~?

marijnh:
That's a good question. It does conflict with their strikethrough syntax -- https://help.github.com/articles/github-flavored-markdown/#strikethrough

So maybe do expose it an option, make it on by default, and have the GFM mode disable it.

brondsem:
Updated with a new commit.

marijnh:
Thank you. Merged as 5e34e117d44bd7e3

PR #3497, opened 2015-09-03
[markdown] Add option to allow ATX headers without a space.  Fixes #3343

### LINKED ISSUE #3343: CodeMirror fails to recognize Markdown headers if they lack a space after the final '#'
CodeMirror fails to add the class `cm-header` to headers that lack a space between the final '#' and the title. For example:

```
# H1 ... has class cm-header
## H2 ... has class cm-header

#H1  ... does not have class cm-header
##H2  ... does not have class cm-header
```

Whether or not there technically should be a space, nearly all Markdown parsers will process `#H1` without a space identically to `# H1` with a space. This leads to inconsistencies when formatting is applied to the headers while editing using the `cm-header` class and when previews are generated using other parsers.

This issue can be tested at http://codemirror.net/mode/markdown/

brondsem:


marijnh:
With tests. Looks great! Merged as 5cd327f

PR #3499, opened 2015-09-06
text/coffeescript for coffeescript's mime type

matthewbauer:
Node's 'mime' package gives me text/coffeescript for this.


marijnh:
Thanks. Merged as 812d779

PR #3501, opened 2015-09-06
a not-so-great fix to an indentation bug

### LINKED ISSUE #3500: haxe mode - indentation breaks when function name declared with capital letter
declaring a function name to begin with a capital letter results in incorrect indentation

    function a()
    {}
    function A()
      {}

(this is because haxe mode recognises all things starting with capital letters as type names, while in this case it's a function name, and things break down)

increpare:
fixes #3500 

I don't know if 

1) there's a way to get haxe to, while in functiondef, for instance, be able to tell the compiler that what it thought was a type name is actually a variable name, and to indent/highlight correctly

or

2) there's a way that, assuming it's a type name, to still indent correctly

but it *is* possible to loosen the lexer to accept things marked as types as function names (in functiondef), and this fixes the indentation problem.

### DIFF
--- a/mode/haxe/haxe.js
+++ b/mode/haxe/haxe.js
@@ -358,7 +358,8 @@ CodeMirror.defineMode("haxe", function(config, parserConfig) {
     if (value == "in") return cont();
   }
   function functiondef(type, value) {
-    if (type == "variable") {register(value); return cont(functiondef);}
+    //function names starting with upper-case letters are recognised as types, so cludging them together here.
+    if (type == "variable" || type == "type") {register(value); return cont(functiondef);}
     if (value == "new") return cont(functiondef);
     if (type == "(") return cont(pushlex(")"), pushcontext, commasep(funarg, ")"), poplex, typeuse, statement, popcontext);
   }


marijnh:
So, I don't know Haxe, and the Haxe mode (which was contributed) appears to be based on the JavaScript mode, which takes an approach to mode writing that does a more or less full parse. This is nice, because it gives use a lot of information about the code, but it is also very demanding and fragile. I'm going to just go ahead and merge your patches without too much scrutiny, because I can't afford to spend a lot of time debugging and improving this mode. One option would be to create a `clike`-derived mode for Haxe, which would be a lot easier to maintain, but you'd lose things like solid type and variable-definition highlighting.

I've merged this as c9fb71c

increpare:
Thanks :) I understand it's a bit hairy keeping mental track of things like this.

I'll keep a close eye on it, and report bugs back (with fixes where possible).

increpare:
(also, until I say otherwise, feel free to @ mention me if anyone finds 'orrible bugs haxe.js )

PR #3502, opened 2015-09-06
haxe mode globalVars

increpare:
porting javascript globalVars code to haxe.js to allow tracking of (previously declared in file) global variables/functions.

As with javasript mode, to enable you call fromAreaText with

    mode: {name:"haxe",globalVars: true},

see the javascript mode file for reference - 
https://github.com/codemirror/CodeMirror/blob/master/mode/javascript/javascript.js

### DIFF
--- a/mode/haxe/haxe.js
+++ b/mode/haxe/haxe.js
@@ -386,11 +386,10 @@ CodeMirror.defineMode("haxe", function(config, parserConfig) {
   }
 
   // Interface
-
   return {
     startState: function(basecolumn) {
       var defaulttypes = ["Int", "Float", "String", "Void", "Std", "Bool", "Dynamic", "Array"];
-      return {
+      var state = {
         tokenize: haxeTokenBase,
         reAllowed: true,
         kwAllowed: true,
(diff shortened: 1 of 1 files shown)


increpare:
Not sure what to do about the Travis CI error. Haven't had a CI error before, not sure how to read the logs, but this looks like the error:

"mode/haxe/haxe.js: Access to global variable state. (405:8)"

I don't believe the code I added is any different from the javascript one, and the code is precisely to modify the global variable state.  I don't know if there's somewhere I was supposed to add an exception to this or not, or if I've messed up in some other way.

marijnh:
> Not sure what to do about the Travis CI error.

It says `mode/haxe/haxe.js: Access to global variable state. (405:8)`, and a casual glance at that line shows that you A) put an unreachable statement directly after a `return`, and B) are using the `state` variable, which doesn't exist in that scope.

marijnh:
Thanks. Merged as 897586c0ad1bec7, slightly tweaked in a039316.

Next time you add to a pull request, make sure to also add a comment. I don't get a notification when a new patch is added.

PR #3505, opened 2015-09-06
Update haxe.js

### LINKED ISSUE #3504: haxe mode - array with trailing commas breaks indentation
according to http://try.haxe.org/#47888 it's valid to have a trailing comma, but it upsets the indentation in CodeMirror's haxe mode :

    var a = [1,];
             Debug.log("boop");

increpare:
fixes #3504 

(not *totally* confident about its correctness, but it looks like a plausible fix, and seems to work ok)


marijnh:
I agree this looks good. Merged as 698c8e5

PR #3509, opened 2015-09-08
Update manual.html

ramwin1:
Hello. When I create my first codemirror script, I found it can't load the file lib/codemirror.css because the filepath is incorrect. the file 'codemirror.css' and 'codemirror.js' is in the same directory, the the html should be <link rel="stylesheet" href="lib/codemirror.css">. Thank you for your watching.


marijnh:
Thank you! Merged as 1ded3f6

PR #3511, opened 2015-09-08
runmode.node.js example

increpare:
There were no examples provided for this, and given the generally pretty great coverage that the documentation has, I thought this simple example that generates some code might be of interest.


marijnh:
There's already an example in `bin/source-highlight`. So I've just pointed towards that in the manual.

increpare:
ah, cool! missed that. thanks! :)

PR #3514, opened 2015-09-08
fixes #3513

### LINKED ISSUE #3513: haxe mode - indentation getting confused that expressions are types
on the current safari/osx, using the current head, it uses "first letter upper case" to guess at what is a type. Which is normally satisfactory, but results in occasionally strange behaviour.

    f(R.a(),
      1,
        1);

vs

    f(r.a(),
      1,
      1);

the second one is desirable (with more arguments, the functions cascade outwards further still). 

(likely) fix incoming...

(also, I only just read the submission guidelines now, apologies for the bad-format bug-reports and ill-tested fixes...)

increpare:
Not 100% sure that this is the best way to deal with this, but pretty sure - 

If you're in an expression, and encounter what you think is a type, it's probably better to change your assumption and treat it as you would a variable/constant (Because that's what it probably is) rather than to forget everything.

I ran the test suite, and it didn't break anything, and the linter, which was quiet.

### DIFF
--- a/mode/haxe/haxe.js
+++ b/mode/haxe/haxe.js
@@ -264,6 +264,7 @@ CodeMirror.defineMode("haxe", function(config, parserConfig) {
   }
   function expression(type) {
     if (atomicTypes.hasOwnProperty(type)) return cont(maybeoperator);
+    if (type == "type" ) return cont(maybeoperator);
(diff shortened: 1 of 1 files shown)


marijnh:
Looks harmless enough. Merged as 45adc4d

Issue #2521, opened 2014-04-30
[gfm] colon in string triggers link highlighting.


shindakun:
When using markdown/gfm the text ab:xyz in the editor is formatted as a 'link'. 
![gfm01](https://cloud.githubusercontent.com/assets/319128/2837051/5ba81344-d00d-11e3-98d2-3396e51dbd67.gif)

PR #3516, opened 2015-09-08
[gfm] limit the url regex to certain schemes

brondsem:
Since foo:bar was matching as a regex (as reported in #2521 but still occurring for me), limit URI schemes to those in the CommonMark spec for autolinks http://spec.commonmark.org/0.22/#autolinks

The spec's list of uri schemes seems pretty long, so it might be better to limit it even more.  Not sure exactly how to pick the schemes that should be supported though.

### DIFF
--- a/mode/gfm/gfm.js
+++ b/mode/gfm/gfm.js
@@ -92,11 +92,12 @@ CodeMirror.defineMode("gfm", function(config, modeConfig) {
           return "link";
         }
       }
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks. That is an awfully ugly regexp, but I guess if its based on a specification I can live with that. Merged as 1a56b954e0ab and followed up with 3c41bf145bd26 for cleanness.

PR #3518, opened 2015-09-09
closebrackets  - selection functionality change

increpare:
Right now if you have a selection, and you enter a bracket/quote, surrounds it with quotation marks, and selects the entire quote:

closebrackets_current

This change should change the behaviour to selecting the interior of the quote instead:

closebrackets_proposed

The motivating example for me was autocomplete, where after autocompleting a function the arguments are selected, so if you want to insert a string literal right now you have to delete the selection before you can enter the quotes. With the proposed fix, it'll be rather more fluid.

closebrackets_timeline

I checked it works ok with multiple selections.
The test suite doesn't have any new things other than the scroll ones that are already broken.
I've passed it through the linter and it checks out.

### DIFF
--- a/addon/edit/closebrackets.js
+++ b/addon/edit/closebrackets.js
@@ -90,6 +90,16 @@
     });
   }
 
+  function contractSelection(sel){
+    if (sel.anchor.line<sel.head.line || (sel.anchor.line===sel.head.line&&sel.anchor.ch<sel.head.ch)){
+      sel.anchor.ch++;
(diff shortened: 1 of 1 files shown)


marijnh:
This is a good idea. Sublime Text does the same thing, so I'm going to assume it is common.

Your patch mutates data that it shouldn't be mutating though (the ranges and positions you get from CodeMirror should not be changed). So I've added df8afed to fix that.

PR #3519, opened 2015-09-09
[oz mode] clean up electric indentation test

anthonygego:
This follows PR #3490. It ensures reindentation is done only when matching block-closing keywords. (Upper-case letters were erroneously put in the list, case-sensitive check is correct, as for highlighting.)


marijnh:
Thanks! Merged as 0c9a2ff

PR #3521, opened 2015-09-10
Option for GitHub spice in GFM mode

WesCossick:
Does not affect default behavior, but does allow users to disable GitHub spice if needed using the `modeConfig` parameter.

By GitHub spice, I mean:

> User/Project@SHA
> User@SHA
> SHA
> User/Project#Num
> User#Num
> #Num


marijnh:
Thanks. Merged as 658ef7809e3 and slightly changed in ece5c232a43

PR #3522, opened 2015-09-10
New addon: Tab list

WesCossick:
Tab list is a new addon in `/addon/edit/` that allows users to map the shift and shift-tab keys to these functions that will indent or unindent a Markdown list element.

### DIFF
--- a/addon/edit/tablist.js
+++ b/addon/edit/tablist.js
@@ -41,7 +41,7 @@
       cm.execCommand('indentLess');
       return;
     }
-        
+
     if(cm.options.indentWithTabs){
       cm.execCommand('insertTab');
     }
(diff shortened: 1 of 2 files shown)


WesCossick:
@marijnh, thoughts on this PR?

marijnh:
Hi Wes. This is too specific for including in the distribution. You'll have to keep it in your local code, or put it somewhere as an [external module](https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons).

WesCossick:
Do you believe this addon is any more specifc than [continue list](https://github.com/codemirror/CodeMirror/blob/master/addon/edit/continuelist.js), which is already included in the repository? I would argue that each fulfils similarly useful and specific roles.

marijnh:
Yes, it is, and also easier to implement locally. The list-continuing addon is also borderline too specific.

PR #3523, opened 2015-09-10
Mode for MscGen and two similar languages

sverweij:
Adds a simple mode for three sequence chart DSLs:
- [MscGen](http://www.mcternan.me.uk/mscgen),
- [Xù](https://github.com/sverweij/mscgen_js/blob/master/wikum/xu.md) - a strict superset of MscGen
- [msgenny](https://github.com/sverweij/mscgen_js/blob/master/wikum/msgenny.md) - a simplified version of MscGen/ Xù

An even cruder version of this mode has been in use in the
[mscgen_js](https://sverweij.github.io/mscgen_js) interpreter for some time.
I've cleaned it up and added some tests - I hope it is CodeMirror worthy.


marijnh:
Thanks! I've merged this as d47c918fd7452 and moved things around a little in 9016ae38f5e208158 to make the modes conform to the expectation that mode X is defined in `mode/X/X.js`, so that it can be dynamically loaded. That means you can no longer load the `xu` and `msgenny` modes by simply specifying that string, but have to use their MIME strings or an object like `{name: "mscgen", language: "msgenny"}`.

PR #3525, opened 2015-09-14
Provide option "somethingSelected" to open completion even if something is selcted

angelozerr:
I'm developping a CodeMirror addon for https://github.com/angelozerr/tern-guess-types

For instance if you open tern completion for `document` and select `addEventListener` :

When you apply completion, tern guess types gives the capability to retrieve variables, functions for each function parameters. Here a screenshot which shows a list of variable with string type for the `addEventListener` `type` argument :

If you see this screenshot, the **type parameter is selected** and with CodeMirror when I call `cm.showHint(),` this.somethingSelected() return true and **completion cannot be opened**.

With this PR, I can call the showHint like this

```javascript
cm.showHint({hint: templateHint, somethingSelected: false});
```

and it works with my case.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -42,8 +42,9 @@
     this.widget = null;
     this.debounce = 0;
     this.tick = 0;
-    this.startPos = this.cm.getCursor();
+    this.startPos = this.cm.getCursor("start");
     this.startLen = this.cm.getLine(this.startPos.line).length;
+    if (options && options.somethingSelected == false ) this.startLen = this.startLen - this.cm.getSelection().length;
 
     var self = this;
     cm.on("cursorActivity", this.activityFunc = function() { self.cursorActivity(); });
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -26,7 +26,7 @@
 
   CodeMirror.defineExtension("showHint", function(options) {
     // We want a single cursor position.
-    if (this.listSelections().length > 1 || this.somethingSelected()) return;
+    if (this.listSelections().length > 1 || (this.somethingSelected() && !(options && options.somethingSelected == false ))) return;
 
     if (this.state.completionActive) this.state.completionActive.close();
     var completion = this.state.completionActive = new Completion(this, options);


angelozerr:
@marijnh I have created a demo with Completion guess:

test

You can play with the online demo at  http://demo-angelozerr.rhcloud.com/CodeMirror-Java/guess-types.html

Hope you will accept this PR, in order to Completion Guess will work. Thanks!

addon/hint/show-hint.js
```diff
@@ -42,8 +42,9 @@
     this.widget = null;
     this.debounce = 0;
     this.tick = 0;
-    this.startPos = this.cm.getCursor();
+    this.startPos = this.cm.getCursor("start");
     this.startLen = this.cm.getLine(this.startPos.line).length;
+    if (options && options.somethingSelected == false ) this.startLen = this.startLen - this.cm.getSelection().length;
```
marijnh: Why is this only done when `somethingSelected` is false? Isn't that the situation in which it should _not_ be done.

angelozerr: By default options.somethingSelected is the same thing than true.

In my case (when I set options.somethingSelected to false), I must remove this.cm.getSelection().length to startLen (like today)

marijnh: Ah, `somethingSelected: false` is how you indicate that completing with selected text _is_ allowed? That doesn't seem like the obvious choice.

angelozerr: > Ah, somethingSelected: false is how you indicate that completing with selected text is allowed?

Yes that it is! Perhaps my code is bad. My need is to open completion with selected text and adjust startLen according the selection.

marijnh:
Attached patch allows you to set a `supportsSelection` property on your _hint function_ (not on options) to indicate that it can safely be called when there is a selection. It also reorganizes the way the automatic hinter works to allow this property (as well as `async`) to work even for automatic hinting.

angelozerr:
@marijnh I have tried to test your patch, it works well with [complete.html](https://github.com/codemirror/CodeMirror/blob/master/demo/complete.html) demo but with [tern.html](https://github.com/codemirror/CodeMirror/blob/master/demo/tern.html) demo the completion doesn't open (it seems that completions.options.hint is null https://github.com/codemirror/CodeMirror/blob/master/addon/hint/show-hint.js#L43)

Just for your information, I have tried too the [html5complete.html](https://github.com/codemirror/CodeMirror/blob/master/demo/html5complete.html] and I have the error "Uncaught TypeError: Cannot read property 'line' of undefined" when I try to open the completion.

marijnh:
You are right, I broke some things. Try again with attached patch.

PR #3528, opened 2015-09-15
Add missing file_put_contents function

bocharsky-bw:
In order to proper syntax highlighting


weaverryan:
The only addition was `file_put_contents` - it's a valid PHP function, so it looks like this might have just been an oversight. :+1:

marijnh:
Thanks. Merged as ed6651b

PR #3531, opened 2015-09-16
Update dialog.css

osake:
Input area feels too short even at 20em, 80% feels a bit better to me across the board and seems solid for various widths of window.


marijnh:
There are dialogs that use multiple text inputs or put a bunch of text and buttons next to them, so I don't think pinning them to 80% is a good idea. You can override this in your local style sheet if it is helpful for your application.

PR #3539, opened 2015-09-18
[ruby] Add `until` to indentWords list

vincentwoo:
In ruby, `until` is just `while !...`


marijnh:
Thank you. Merged as 3d1a056

PR #3540, opened 2015-09-19
Syntax highliting for C++14 numeric literals.

melpon:
In C++14, the single-quote character may be used arbitrarily
as a digit separator in numeric literals.

https://en.wikipedia.org/wiki/C%2B%2B14#Digit_separators


marijnh:
Thanks. Merged as 08df099

PR #3541, opened 2015-09-20
haxe mode - adding in semicolons to satisfy linter

increpare:


marijnh:
I'm going to move CodeMirror to semicolon-less style somewhere in the future. I suggest you adjust your linter to not complain about semicolons (or, if it can't do that, to get a new linter).

PR #3544, opened 2015-09-21
Move Kotlin mode to clike.

satamas:
Kotlin mode is just a old copy of clike mode. The only difference is soft keywords, that are not supported by codemirror. So i think it is a good idea to use clike mode for kotlin.


marijnh:
Thanks for cleaning this up. Merged as ed2e611 .

@Leonya CC-ing you since you wrote (/copied) the original mode

PR #3549, opened 2015-09-25
修正merge插件的ie8兼容性

trumanliu:
addEventListener() doesn't work in IE8.I use attachEvent() instead when addEventListener() is not being supported. All tests paassed.


marijnh:
There's a `CodeMirror.on` utility function for this that should have been used instead. Attached patch fixes this.

PR #3550, opened 2015-09-25
New option: allowDroppedFileTypes

WesCossick:
Dropping an image, PDF, or other special file type results in unreadable characters being added to the editor. Plus, if the file is large enough (like images typically are), it'll freeze the browser. This fixes the issue by giving an option to specify which file types are allowed to be dropped. All others will be ignored.

This does not affect the default behavior however.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -3823,6 +3823,9 @@
       var loadFile = function(file, i) {
         var reader = new FileReader;
         reader.onload = operation(cm, function() {
+          if(this.options.allowDroppedFileTypes !== undefined && this.options.allowDroppedFileTypes.indexOf(file.type) === -1)
+            return;
+          
           text[i] = reader.result;
           if (++read == n) {
             pos = clipPos(cm.doc, pos);
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -3821,11 +3821,11 @@
     if (files && files.length && window.FileReader && window.File) {
       var n = files.length, text = Array(n), read = 0;
       var loadFile = function(file, i) {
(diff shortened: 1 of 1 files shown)


marijnh:
Would you also be okay with the following alternative solutin?
 
 - Detect binary files, and simply refuse to insert them

 - Add an option to set a max file size, and ignore files bigger than this

WesCossick:
That sounds like it would the accomplish the same thing, so I wouldn't be opposed to this solution either. Maybe even having both in conjunction with one another. One question though, would Word Documents and/or PDFs be considered binary files?

marijnh:
PDFs would turn into a lot of postscript soup (they are textual, but not necessarily readable). Word Docs are binary data, and unless we implement a binary format parser for them in JavaScript, we can't get at any sane textual representation of them.

WesCossick:
I'm thinking that only plain text would result in something legible when dropped in the editor. Non-binary files can still result in a mess of characters.

marijnh:
Could it be that you never even tested this? Because `this`, inside of that function, does not refer to a CodeMirror instance, and the code simply errored out as it is. I've fixed it up in 031afde.

I've also added a patch that will prevent users from dragging binary files into the editor.

WesCossick:
The PR worked in my local repository, though I've made a number of other changes that could have affected the way it behaved there versus when I copied and pasted just this relevant change into a new branch.

With that said, thanks for cleaning it up and committing the improvement.

marijnh:
(I forgot to mention I renamed the option to `allowDroppedFileTypes` for naming consistency with similar options)

PR #3553, opened 2015-09-28
Update Cypher keywords etc

nawroth:
* Removes like, ilike.
* Adds starts, ends, contains, toString, size, reverse.


marijnh:
Thanks. Merged as 80ba426

PR #3559, opened 2015-10-01
Update coffeescript.js

Kaworru:
Removed possible testing and unreachable code to avoid JavaScript compiler warnings:
...
codemirror/mode/coffeescript/coffeescript.js:271: WARNING - unreachable code
    if (false && current === ".") {
...


marijnh:
Thanks. I forgot to delete that after disabling it. Merged as 825ca00

PR #3560, opened 2015-10-01
Support for Ceylon language

quintesse:
This PR adds support for the [Ceylon programming language](http://ceylon-lang.org) to the existing "clike" mode.

I added a couple of minor features to the "clike" configuration, mostly to be able to pass a function to define keywords instead of a fixed list of words (this because in Ceylon all "types" start with an uppercase letter, so this is a much better test for us than trying to list a large number of often-used type names).

I also made the `isOperatorChar` configurable because we have a couple more symbols that are used as operators.

Anyway, if you have any remarks, just let me know.

PS: Very nice job on CodeMirror! It's a great piece of software. We're using it here in our online "Web IDE" (so people can try out bits of code): http://trybeta.ceylon-lang.org/

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -449,6 +449,7 @@ stoskov
 Sungho Kim
 sverweij
 Taha Jahangir
+Tako Schotanus
 Takuji Shimokawa
 Tarmil
 tel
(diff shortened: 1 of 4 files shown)


marijnh:
Thanks, looks great. Merged as 4255f78

quintesse:
That's quick, thanks!

PR #3562, opened 2015-10-01
Allow markdown styles to be overridden

brennie:
The `markdown` mode now supports styling overrides. The configuration
object passed to the class now accepts a `classOverrides` object, which
allows the user to sepcify different CSS classes for elements. This is
useful for the case where the default CSS classes are reused for
multiple elements (e.g. `list1` maps to `variable-2` by default) and
more customized styling is desired.

Unit tests have been added to cover all types of styling overrides.

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -51,24 +51,36 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
   if (modeCfg.strikethrough === undefined)
     modeCfg.strikethrough = false;
 
+  // Allow token types to be overridden by user-provided token types.
+  if (modeCfg.tokenTypeOverrides === undefined)
+    modeCfg.tokenTypeOverrides = {};
+
   var codeDepth = 0;
 
-  var header   = 'header'
-  ,   code     = 'comment'
-  ,   quote    = 'quote'
-  ,   list1    = 'variable-2'
-  ,   list2    = 'variable-3'
-  ,   list3    = 'keyword'
-  ,   hr       = 'hr'
-  ,   image    = 'tag'
-  ,   formatting = 'formatting'
-  ,   linkinline = 'link'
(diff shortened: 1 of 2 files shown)


mode/markdown/markdown.js
```diff
@@ -51,24 +51,28 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
...
+  ,   hr       = modeCfg.classOverrides.hr || "hr"
+  ,   image    = modeCfg.classOverrides.image || "tag"
+  ,   formatting = modeCfg.classOverrides.formatting || "formatting"
+  ,   linkinline = modeCfg.classOverrides.linkinline || "link"
+  ,   linkemail = modeCfg.classOverrides.linkemail || "link"
+  ,   linktext = modeCfg.classOverrides.linktext || "link"
+  ,   linkhref = modeCfg.classOverrides.linkhref || "string"
+  ,   em       = modeCfg.classOverrides.em || "em"
+  ,   strong   = modeCfg.classOverrides.strong || "strong"
+  ,   strikethrough = modeCfg.classOverrides.strikethrough || "strikethrough";
```
marijnh: How about moving the styles into an object, say `styles`, and using a for/in loop to do the overriding? That's less noisy and easier to extend in the future (it does make the patch bigger, requiring you to rename all references to these, but that should be safe and improve clarity -- I've found myself wondering what, say, `code` refers to in a given context).

Also, since these aren't direct CSS classes (they will be prefixed with `cm-`), we tend to call them styles or token types, not classes. It'd be good if the option name reflected that.

marijnh:
Thanks. Merged as 770a771

When you change a pull request branch, it is a good idea to follow up with a comment, because I don't get email otherwise, and won't notice your update until the next time I look over the list of open pull requests (which might be a while).

brennie:
Thanks for merging this, @marijnh! I know now for next time :)

PR #3567, opened 2015-10-03
Multiplex addon: Add open and close tokens when highlighting delimiters

doryphores:
This simple change adds extra `open` and `close` token styles to multiplexing delimiters when `delimStyle` is set. This makes it easy to style open and close delimiter tokens as in this example:

screen shot 2015-10-03 at 16 13 23

For example, if `delimStyle` is set to `delim`, then the opening delimiter gets the following token styles: `cm-delim` and `cm-delim-open` and the closing delimiter gets `cm-delim` and `cm-delim-close`.


marijnh:
Looks useful. Merged.

PR #3569, opened 2015-10-04
[tern demo] Use protocol relative URLs

aviraldg:
https://codemirror.net/demo/tern.html is broken at the moment; this fixes it

### DIFF
--- a/demo/tern.html
+++ b/demo/tern.html
@@ -13,16 +13,16 @@
 <script src="../addon/dialog/dialog.js"></script>
 <script src="../addon/hint/show-hint.js"></script>
 <script src="../addon/tern/tern.js"></script>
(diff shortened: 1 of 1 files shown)


marijnh:
How does this fix anything? `ternjs.net` doesn't support HTTPS, so it'll just break things in a different way.

aviraldg:
That's something only you can fix :sweat_smile:

marijnh:
Yes, I'll put it on my list (though messing around with SSL certs isn't my favorite hobby)

idleberg:
@aviraldg Protocol-relative URLs are now an [anti-pattern](http://www.paulirish.com/2010/the-protocol-relative-url/)

PR #3571, opened 2015-10-05
Updated Ceylon mode

quintesse:
Affecting all modes:

 - Made punctuation and number characters configurable.
 - Undid a change I made in my previous PR because I noticed similar functionality was already supported with the existing `styleDefs` configuration option.

Affecting only the Ceylon mode:

 - Strings now support interpolation (code within strings)
 - Not using `def` style anymore
 - Style members using `variable-2`

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -678,7 +678,7 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
       return "string";
     }
   }
-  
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as d8237c8f1d3 and 84d5f64006

quintesse:
Thank *you*!

PR #3575, opened 2015-10-07
Changed the link color in abcdef theme

amshali:
Changed the link color in abcdef theme to blueviolet, because the original blue on black background is difficult to read.


marijnh:
Thanks. Merged as b629242

PR #3576, opened 2015-10-07
Fixed the liquibyte theme css file to have more complete selectors

amshali:
Fixed the liquibyte theme css file to have more complete selectors. Including liquibyte theme with other themes causes other themes(when in use) to not look right. For example there will be underline text decorations when a text is selected in default theme.


marijnh:
Thanks for noticing that! Merged as 61e9389

PR #3579, opened 2015-10-07
fix vertical scroll jitter on OSX trackpads

sccolbert:
This fixes an issue (jupyter/notebook#513) we see when using a trackpad to vertically scroll a page with several CodeMirror instances, none of which have vertical overflow. A full description/explanation of the cause can be found in the [issue comment](https://github.com/jupyter/notebook/issues/513#issuecomment-146073284). 

Basically, preventing the default action on a small `deltaX` wheel event prevents the page from scrolling vertically. Since humans are imprecise, and OSX track pads support simultaneous `deltaX` and `deltaY`, scrolling such pages is jittery because CodeMirror squashes some percentage of the events where `deltaX` is not zero.

This works around the issue by only preventing the default if there was either no vertical scroll delta, or if there is actually content which could be vertically scrolled.

This PR assumes you have prevented the default in order to avoid horizontally scrolling the page *in addition* to horizontally scrolling the CodeMirror content (which seems reasonable to me). If you think it's not worth the bother, we can just remove the `preventDefault`.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -4004,7 +4004,7 @@
       // actually possible. Otherwise, it causes vertical scroll
       // jitter on OSX trackpads when deltaX is small and deltaY
       // is large:
-      // https://github.com/jupyter/notebook/issues/513#issuecomment-146073284
+      // https://github.com/codemirror/CodeMirror/pull/3579
       if (!dy || (dy && canScrollY))
         e_preventDefault(e);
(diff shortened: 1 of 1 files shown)


marijnh:
Looks like a good solution. Thanks for figuring out the problem. Merged as 5fafcf8

sccolbert:
Cheers!

PR #3580, opened 2015-10-08
Restore true as a Scala keyword

jengler:
Removed in 3d5d9a4af38facd8c2d79abbe74ead61470ca97b as it had been
mispelled when introduced in eb495233e3813d04c20e495dcf88217e093261fc.

After talking with @marijnh, will remove false as Scala keyword in favor of true/false being defined as atoms.

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -478,9 +478,9 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
     keywords: words(
 
       /* scala */
-      "abstract case catch class def do else extends false final finally for forSome if " +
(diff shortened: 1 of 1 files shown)


marijnh:
`true` and `false` are actually defined as atoms in the Scala mode (and in most other clike modes). `false` accidentally also appears as a keyword. Would it be okay to remove `false` from keywords and just have both of them be atoms?

jengler:
Sounds good to me :) I will update PR. Thank you

marijnh:
Thanks. Merged as 872bb73

PR #3583, opened 2015-10-09
refactor local variable name to save a little bit file size when build

zhenyong:


marijnh:
Merged (with trailing whitespace fix) as de7b15f

PR #3586, opened 2015-10-09
first commit

idleberg:


marijnh:
I'd like the filenames to match the theme names whenever possible. Would you prefer to add the `-dark` prefix to the filenames, or to remove it from the css classes?

idleberg:
Oh, right. That's a remainder from base16-builder, removed it from the file classes!

marijnh:
Your commit only modifies hopscotch.css. Should the same be done for the other files?

idleberg:
👍

marijnh:
All right. Merged as bc008c070f

PR #3587, opened 2015-10-10
Allow passing options to toggleComment

vincentwoo:
The `toggleComment` helper is very useful, and the sublime keymap uses it by default. However, the actual sublime behavior for that keystroke is to comment with indentation.

This change makes `toggleComment` a first class extension method, and leaves the command (of the same name) to just call the extension. The new extension also takes an options object to allow more customization.

### DIFF
--- a/addon/comment/comment.js
+++ b/addon/comment/comment.js
@@ -21,22 +21,27 @@
   }
 
   CodeMirror.commands.toggleComment = function(cm) {
-    var minLine = Infinity, ranges = cm.listSelections(), mode = null;
+    cm.toggleComment();
+  };
+
+  CodeMirror.defineExtension("toggleComment", function(options) {
+    if (!options) options = noOptions;
+    var minLine = Infinity, ranges = this.listSelections(), mode = null;
     for (var i = ranges.length - 1; i >= 0; i--) {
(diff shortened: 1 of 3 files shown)


addon/comment/comment.js
```diff
@@ -21,22 +21,22 @@
...
+
+  CodeMirror.defineExtension("toggleComment", function(options) {
+    if (!options) options = noOptions;
+    var minLine = Infinity, ranges = this.listSelections();
     for (var i = ranges.length - 1; i >= 0; i--) {
       var from = ranges[i].from(), to = ranges[i].to();
       if (from.line >= minLine) continue;
       if (to.line >= minLine) to = Pos(minLine, 0);
       minLine = from.line;
-      if (mode == null) {
```
marijnh: Why did you remove the mode-preserving logic here?

vincentwoo:
@marijnh I deleted the mode stuff because I am a big idiot who codes tired. I restored it - I thought the mode code could never do anything because `mode` was always `null`... hah.

marijnh:
Merged as 8a012c0ae93654, with the change of binding `cm` to `this` in the extension method, so that it actually works.

vincentwoo:
oh god, good catch, sorry

PR #3588, opened 2015-10-10
Smarter indenting in multiline linecommenting

vincentwoo:
When indenting multiple lines with `lineComment` and `indent: true`, you currently get behavior like this:

```ruby
def test
  p ‘comment me’
end

# after commenting the last two lines…

def test
  # p ‘comment me’
  # end
```

Which messes up the indentation of the block. Preferably, it would look like (and does in sublime, at least):

```ruby
def test
#   p ‘comment me’
# end
```

This change makes `lineComment` smarter by looking for the minimum indent level of the selected range, and only commenting at that depth. It’s not 100% smart in that it doesn’t use column positioning to account for mixed tab usage, but its better than what we have right now.

### DIFF
--- a/addon/comment/comment.js
+++ b/addon/comment/comment.js
@@ -57,7 +57,14 @@
 
     self.operation(function() {
       if (options.indent) {
-        var baseString = firstLine.slice(0, firstNonWS(firstLine));
+        var baseString = null;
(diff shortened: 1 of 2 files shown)


marijnh:
Looks like an improvement. Merged as 8fd466f

PR #3589, opened 2015-10-10
Fix TypeScript mode to highlight "boolean" not "bool"

elisee:
Hey there. This PR fixes the type name for boolean variables in TypeScript, which is spelled "boolean", not "bool" (which I guess was infered from C++ - EDIT: actually looks like it was initially named bool and then they fixed it - https://typescript.codeplex.com/workitem/135).

Thanks!


marijnh:
Thanks! Merged as 0ca9fc8

PR #3593, opened 2015-10-14
Merge addon: Use npm's copy of diff-match-patch for CommonJS requires

wbinnssmith:
This lists `diff-match-patch` as a formal requirement in package json and updates the UMD wrapper in the merge addon to properly use it (afaict it currently just requires it and throws it away). We could probably use the [`diff_match_patch` module on npm](https://www.npmjs.com/package/diff_match_patch) (note the underscores), but it has no README and I simply trust @ForbesLindesay more as a maintainer.

Since these packages (responsibly :stuck_out_tongue_closed_eyes:) attach the `DIFF_INSERT`, etc. enum values to the module and not the global scope, we also check the module to see if they exist there before falling back to the global versions (which don't exist in the CJS versions).

### DIFF
--- a/addon/merge/merge.js
+++ b/addon/merge/merge.js
@@ -3,17 +3,21 @@
 
 // declare global: diff_match_patch, DIFF_INSERT, DIFF_DELETE, DIFF_EQUAL
 
-(function(mod) {
+(function(mod, root) {
   if (typeof exports == "object" && typeof module == "object") // CommonJS
-    mod(require("../../lib/codemirror"), require("diff_match_patch"));
+    mod(require("../../lib/codemirror"), require("diff-match-patch"), root);
   else if (typeof define == "function" && define.amd) // AMD
-    define(["../../lib/codemirror", "diff_match_patch"], mod);
+    define(["../../lib/codemirror", "diff_match_patch"], mod, root);
   else // Plain browser env
-    mod(CodeMirror);
-})(function(CodeMirror) {
+    mod(CodeMirror, diff_match_patch, root);
+})(function(CodeMirror, diff_match_patch, root) {
   "use strict";
   var Pos = CodeMirror.Pos;
   var svgNS = "http://www.w3.org/2000/svg";
+
+  var DIFF_INSERT = 'DIFF_INSERT' in diff_match_patch ? diff_match_patch.DIFF_INSERT : root.DIFF_INSERT;
+  var DIFF_DELETE = 'DIFF_DELETE' in diff_match_patch ? diff_match_patch.DIFF_DELETE : root.DIFF_DELETE;
+  var DIFF_EQUAL = 'DIFF_EQUAL' in diff_match_patch ? diff_match_patch.DIFF_EQUAL : root.DIFF_EQUAL;
 
   function DiffView(mv, type) {
     this.mv = mv;
@@ -772,4 +776,4 @@
   CodeMirror.commands.goPrevDiff = function(cm) {
     return goNearbyDiff(cm, -1);
   };
-});
+}, this);
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "license": "MIT",
     "directories": {"lib": "./lib"},
     "scripts": {"test": "node ./test/run.js"},
+    "dependencies": {"diff-match-patch": "^1.0.0"},
     "devDependencies": {"node-static": "0.6.0",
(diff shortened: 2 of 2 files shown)


addon/merge/merge.js
```diff
@@ -5,15 +5,18 @@
...
-    mod(CodeMirror);
-})(function(CodeMirror) {
+    mod(CodeMirror, diff_match_patch);
+})(function(CodeMirror, diff_match_patch) {
   "use strict";
   var Pos = CodeMirror.Pos;
   var svgNS = "http://www.w3.org/2000/svg";
+  var DIFF_INSERT = diff_match_patch.DIFF_INSERT || DIFF_INSERT;
+  var DIFF_DELETE = diff_match_patch.DIFF_DELETE || DIFF_DELETE;
+  var DIFF_EQUAL = diff_match_patch.DIFF_EQUAL || DIFF_EQUAL;
```
ForbesLindesay: The `|| DIFF_EQUAL` here won't work because you just declared `DIFF_EQUAL` so it will always be undefined, You either need to make your variable have a different name, or get a reference to the global object first and then do something like `|| window.DIFF_EQUAL`.

ForbesLindesay:
For what it's worth, I haven't touched that project since I first published it, so don't trust me too much on that one.  However, if someone else is keen to update it I'm happy to hand it over :smile:

addon/merge/merge.js
wbinnssmith: ugh yes. thanks.

Also noticed that these enum values are simple ints, where `DIFF_EQUAL` is
0 in the original implementation, so the `||` shortcircuiting will fail in
the last case, where `diff_match_patch.DIFF_EQUAL` is 0 and `DIFF_EQUAL` is
undefined.

wbinnssmith:
@marijnh do you have any thoughts on this? Much appreciated.

marijnh:
The problem with this is that it's going to require everybody who installs CodeMirror to download this library, even though 99% of them don't need it. I don't think that's acceptable. CodeMirror 6 is going to be more modular, and big addons like this will become their own NPM modules. Until then (and it might take a while for me to get around to working on this), I'm afraid you're just going to have to live with little inconveniences like having to manually specify sub-dependencies in your own project file.

ForbesLindesay:
Downloading the module is cheap, practically free.  What does matter is whether end users need to download it to their browser as part of the bundle, but this doesn't change that.

wbinnssmith:
@ForbesLindesay is right. diff-match-patch is dependency-free and quite
small. But that size only matters in "npm install". Whether or not this
gets downloaded to the browser is up to the bundler, and browserify and
webpack will only do so if the require call is in the AST, which it would
only be if the merge addon is itself required.

The merge addon simply doesn't work in its current state with CJS bundlers
like browserify since it requires a module that isn't listed as a
dependency. Even then, the result of that module require is unused and a
global (many global a in this case) is assumed to exist anyway.

wbinnssmith:
@maijnh will you please reconsider?

If you'd rather leave the dependency implicit (which I would still strongly
recommend against), those who do use the module from npm will receive one
that doesn't leak globals.

Either that or the require call should be removed and users will be
expected to actively leak these globals themselves.

I feel both of these options are less ideal though, and feel like a step
back for browserify and webpack users. Of course, users who don't require
the merge addon will not receive this library in their bundle, nor will
users of script tags or the like.

marijnh:
I've pushed a patch that removes the `require` call for `diff_match_patch`. I am really not interested in listing it as a dependency. There's some other modules (such as the linters) that depend on external modules, and I do not want to require people to download all of these every time they install CodeMirror (that counts as a cost, even if you don't experience it as such).

So again, this isn't a permanent state of affairs -- the merge addon will become its own repository/module eventually, and will also probably move to another diffing implementation because diff_match_patch isn't so great. But for now, you'll have to work with it.

wbinnssmith:
Thanks for your understanding. I'd love to help with the modularization of CodeMirror. Is that work somewhere at the moment?

marijnh:
There is no work on it yet. Basically, I want to move the core library into a bunch of ES6 modules, and split a lot of stuff out of the repository into smaller npm modules. But I want to do it in one swoop, when releasing version 6, so I'm not sure it'd make much sense to start on it now -- we'd have to keep the code in the main repo as well until the next major version, and duplicating code rarely ends well.

PR #3596, opened 2015-10-14
Handle <constructor> tags in htmlmixed mode (fixes mozilla bug 1214663)

bgrins:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1214663.  The HTML mixed syntax mode is breaking the editor when encountering a `<constructor>` tag, so instead of checking that tagInfo exists, we need to make sure it's actually an array before continuing into findMatchingMode.

You can see the problem if you update the modes index.html demo to include this tag:

    <constructor>
      foobar
    </constructor>

### DIFF
--- a/mode/htmlmixed/htmlmixed.js
+++ b/mode/htmlmixed/htmlmixed.js
@@ -91,10 +91,11 @@
     function html(stream, state) {
       var tagName = state.htmlState.tagName;
       var tagInfo = tagName && tags[tagName.toLowerCase()];
+      var isTagInfoArray = Object.prototype.toString.call(tagInfo) === "[object Array]";
 
       var style = htmlMode.token(stream, state.htmlState), modeSpec;
 
-      if (tagInfo && /\btag\b/.test(style) && stream.current() === ">" &&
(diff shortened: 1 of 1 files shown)


mode/htmlmixed/htmlmixed.js
```diff
@@ -91,10 +91,11 @@
     function html(stream, state) {
       var tagName = state.htmlState.tagName;
       var tagInfo = tagName && tags[tagName.toLowerCase()];
+      var isTagInfoArray = Object.prototype.toString.call(tagInfo) === "[object Array]";
```
bgrins: I also considered using Array.isArray, but IE8 doesn't support that.  I took this from the polyfill at https://developer.mozilla.org/ro/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray.

marijnh:
Woops how shoddy. I've pushed an alternative patch that solves this by checking out whether the property exists in the object itself, rather than checking for an array. That seems more idiomatic and robust.

bgrins:
Alright, thanks.  Looks good to me.

PR #3597, opened 2015-10-14
[shell mode] add `PKGBUILD' as a file name

darealshinji:
`A PKGBUILD is a shell script containing the build information required by Arch Linux packages.`
https://wiki.archlinux.org/index.php/PKGBUILD


marijnh:
Thanks. Merged as 4096f2d

PR #3598, opened 2015-10-14
[dart mode] add support for triple-quoted multi-line strings

goderbauer:
Previously, triple-quoted multi-line strings could be closed with a single quote, e.g. CodeMirror thought the following was a legal string literal: '''This string literal is not terminated correctly'.

Also reported here: https://github.com/dart-lang/dart-pad/issues/667


goderbauer:
Closing this in favor of the more advanced PR #3601.

PR #3601, opened 2015-10-15
[dart mode] support for triple-quoted strings and string interpolation

goderbauer:
- Previously, triple-quoted multi-line strings could be closed with a single quote, e.g. CodeMirror thought the following was a legal string literal: ```'''This string literal is not terminated correctly'```. Also reported here: dart-lang/dart-pad#667.
- String interpolation with ```$identifier``` and ```${expression}``` now works

### DIFF
--- a/mode/dart/dart.js
+++ b/mode/dart/dart.js
@@ -26,10 +26,21 @@
     return obj;
   }
 
+  function pushInterpolationStack(state) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Looks great. Merged as caf1323d81cf

Issue #3122, opened 2015-03-09
JSX support?


chuyik:
Hi devs, how about adding JSX support into [javascript.js](https://github.com/codemirror/CodeMirror/blob/master/mode/javascript/javascript.js)?

marijnh:
If you want to work on this, the way to do it would be to make the JavaScript mode extensible, so that you can override the way it parses expressions. And then write an extension that, in case an expression starts with a `<`, switches to the JSX mode.

eranimo:
What about this?
https://code.google.com/p/lite/source/browse/trunk/Lite/web/doc/codemirror/jsx.js?spec=svn1482&r=1482

I've tried using it, but I get an error: `Uncaught RangeError: Maximum call stack size exceeded`

weblogixx:
:+1: Would also love to see this feature.

baconface:
+1

marijnh:
The link @eranimo posted looks a lot like the approach I would take if I were to implement it. So instead of posting +1's here, you could try debugging and fixing it (though note that it's LGPL code -- you might want to contact the author and ask if it can be moved under an MIT license first).

Ore4444:
+1

pcostanz:
:+1:

williamstein:
Hi,

I was using CJSX (the coffeescript analogue of JSX) recently in codemirror,
and needed syntax highlighting and indent.  I found a mode for Adobe
Brackets and rewrote it to work in codemirror, and put it here:


https://github.com/sagemathinc/smc-public/blob/master/static/codemirror-5.3/mode/coffeescript2/coffeescript2.js

Anybody is welcome to it (under the Brackets's MIT license)...

William

On Fri, May 29, 2015 at 12:16 PM, Patrick Costanzo <notifications@github.com
> wrote:

> [image: :+1:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/codemirror/CodeMirror/issues/3122#issuecomment-106906712>
> .
>



-- 
William (http://wstein.org)

abdelouahabb:
for the moment I use this trick:
https://github.com/abdelouahabb/CodeMirror/commit/521c476de9b8a961a3205e7284863de88d012a96

foo123:
[...]

PR #3605, opened 2015-10-17
add JSX mode in javascript.js 

kidwm:
I borrowed some code from `xml.js`, and made simple implement to handle JSX tag.
It does not touch `lexical` part, maybe not a perfect solution, but it works.
This should solve #3122.


marijnh:
Modifying javascript.js and copying code from xml.js isn't the right approach here. Rather, you'd want a third mode that wraps the two existing modes, so that we don't entangle these concepts and make the individual modes more complex.

PR #3606, opened 2015-10-19
:memo: `+input` origin documentation

basarat:
Just good to have :rose:


marijnh:
Good call. Merged as 4df95b1

PR #3607, opened 2015-10-20
[javascript mode] style metaproperty as keyword

caitp:
Add a mechanism for styling metaproperties, which should be simple to extend
if new ones are added in the future.


caitp:
Eh, there's a bit more work needed on this, as seen in https://twitter.com/caitp88/status/656351079358054400 (and some more work needed on template literals, too). I guess I'll make an attempt to fix it.

marijnh:
I've merged your test as 71ac921c4385 and provided my own implementation in 457468315e817. Let me know whether it works for you.

PR #3608, opened 2015-10-20
SVG tag

Tymek:
Maybe SVG is not "official" HTML5 tag, but I use it a lot. Consider HTML Design Principles - 3.2. Priority of Constituencies.

### DIFF
--- a/mode/stylus/stylus.js
+++ b/mode/stylus/stylus.js
@@ -725,8 +725,8 @@
     };
   });
 
-  // developer.mozilla.org/en-US/docs/Web/HTML/Element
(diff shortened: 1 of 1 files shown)


marijnh:
But then, shouldn't we also include all other SVG tags (`<defs>`, `<ellipse>`, etc)

idleberg:
> But then, shouldn't we also include all other SVG tags (`<defs>`, `<ellipse>`, etc)?

There are so many [SVG elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Element), it's probably worth creating a separate mode.

marijnh:
The CSS mode actually gets by fine without defining a set of tags, and HTML5 is 'open-ended' anyway, in that you can use your own tags, so I think what we should be doing is fixing the stylus mode to no longer rely on this list.

Tymek:
ok

PR #3611, opened 2015-10-20
Add a mustache mode

### LINKED ISSUE #3602: Support a Mustache mode that highlights the templated language
**Preface:** I'm opening this issue mostly for the technical discussion. If this feature seems possible and would be accepted into the project, I'd be happy to do the work.

I think that most often, Mustache template files have two extensions: the "wrapped" language extension, followed by ".mustache". So an HTML template file has the extension ".html.mustache" and a Markdown template might have the extension ".md.mustache".

It would be awesome if CodeMirror supported a mixed mode Mustache mode that highlighted the "wrapped" language as normal, and caught Mustache tags, highlighting them as Mustache tags.

One way that mustache makes this crazy simple is it's [crazy simple interface](https://mustache.github.io/mustache.5.html): Mustache tags are exactly the two following syntaxes: `{{tag}}` and `{{{tag}}}`. The mustache highlighter could distinguish between variables `{{name}}` and sections `{{#list}}`, etc., but honestly, if they were just plain text, that'd be fine.

The ["Set Delimiter"](https://mustache.github.io/mustache.5.html#Set-Delimiter) functionality would probably be hard to support, but I don't think it would be needed in v1. It's not a commonly used feature.

The only tricky thing I can think of is the Mustache mode knowing the file extensions, and dynamically using the proper mode for the "wrapped" language.

srawlins:
Fixes #3602 

In this PR:

* add a mustache mode
* replace `vue.js`'s implementation of mustache with the new one (essentially the same)
* delete `demo/mustache.html` and redirect the overlay example to `mode/mustache/index.html`.


marijnh:
Are Mustache and Handlebars really different enough to warrant two separate modes? Couldn't we do something like we do with the clike and css modes, where different similar languages are parsed by passing different configuration parameters to a single mode implementation?

srawlins:
Good idea. In fact, most of the templating languages are incredibly similar to Mustache (Django, vue?).

I don't know what the base should be. Mustache is the simplest, so maybe Mustachelike?

In any case, I think this PR with the `backdrop:` functionality makes it the most powerful, and correct for Mustache. In the next PR, should I change Handlebars to be a configuration of Mustachelike?

marijnh:
I think using the multiplexing addon for wrapping the underlying language is fine. Both the Django and Handlebars modes already allow that. Your mode simply styles everything between {{ and }} as strong. The other modes actually understand things like comments, strings, and keywords inside of the template elements, which seems useful.

Maybe take the Handlebars mode, make it configurable, and add an easy way to wrap it around another mode, as in your `backdrop` option?

srawlins:
Thanks for the suggestions, @marijnh. I've created an alternative PR, #3624, to compare with this one. What do you think?

srawlins:
@marijnh implemented a different feature, and commented about it over at #3624. Thanks!

PR #3612, opened 2015-10-21
[django-mode] Added template tag keywords

TDaglis:
Added the following forgotten tags:
`filter`, `endfilter`, `elif`, `endblocktrans`, `endspaceless`

[Built-in template tags reference](https://docs.djangoproject.com/en/1.8/ref/templates/builtins/)


marijnh:
Thanks! Merged as d05dc6e

PR #3617, opened 2015-10-27
Julia mode: Indent using indentUnit option instead of hardcoded 4

JobJob:


marijnh:
Thanks! Merged as de1f27d

JobJob:
Thank *you*! :+1:

PR #3622, opened 2015-11-02
[dart mode] fix escaping in raw strings

goderbauer:
In a raw string, a character following a "\" is not escaped.


marijnh:
Thanks! Merged as a155d55

PR #3624, opened 2015-11-04
Add "mustachelike"; "mustache", "handlebars" and vue.js use it

### LINKED ISSUE #3602: Support a Mustache mode that highlights the templated language

srawlins:
An alternative PR to #3611, fixes #3602, with @marijnh's suggestions.

* This creates a `mustachelike.js` which allows for easily writing new simple syntaxes that can be overlayed.
* Two modes are defined in the file: "mustache" and "handlebars".
* I chose "mustachelike" as [Mustache](https://en.wikipedia.org/wiki/Mustache_(template_system)) appears to be the father of [Handlebars](https://en.wikipedia.org/wiki/Handlebars_(template_system)).
* Django, Soy, and Tornado, are quite different from Mustache and Handlebars (they have `{% %}` syntax, for example, so that it wouldn't make sense to include them.


marijnh:
Nope, still not really what I mean. So searching around the web, I have several websites telling me that Handlebars is a superset of Mustache syntax. So why, precisely, aren't you simply using the existing Handlebars mode and leaving it at that?

srawlins:
Handlebars does appear to be essentially a superset of Mustache (Handlebars does not feature Mustache's [Set Delimiter syntax](https://mustache.github.io/mustache.5.html#Set-Delimiter), which my PR also does not include).

> So why, precisely, aren't you simply using the existing Handlebars mode and leaving it at that?

The handlebars mode does not support overlaying highlighting onto another language. Instead, every user who wants to overlay handlebars (or mustache) onto another language must copy and paste [7 lines](https://github.com/codemirror/CodeMirror/blob/master/mode/handlebars/index.html#L64..L70) and then make configurable per backdrop language. The goal was to have a mode that highlights Mustache syntax and accepts a backdrop language, the concept taken from vue.js.

srawlins:
Should we instead apply the overlay configurability to the Handlebars syntax and be done with the Mustache parser?

marijnh:
Patch 88a0257 does this. Give it a try.

srawlins:
@marijnh Thanks much this is perfect! Thanks for your work!

niftylettuce:
@marijnh @srawlins Can we please update documentation here <https://github.com/codemirror/CodeMirror/blob/master/demo/mustache.html>? 

```
mode: {name: "handlebars", base: "text/html"}
```

Issue #3625, opened 2015-11-04
[css mode] Support CSS4 #rrggbbaa and #rgba colors


JosephPecoraro:
Support CSS4 `#rrggbbaa` and `#rgba` colors. Hex with alpha.

Currently the css mode treats these as errors.

```css
body {
  color: #F00;
  color: #F00F;
  color: #FF0000;
  color: #FF0000FF;
}
```

[WebKit supports](http://trac.webkit.org/changeset/192023) them. I don't think Firefox or Chrome do at the moment. [Specification](https://drafts.csswg.org/css-color/#hex-notation) is still a draft, but I don't think there is anything controversial about this.

marijnh:
Closed via #3626

PR #3626, opened 2015-11-04
[css mode] Support CSS4 #rrggbbaa and #rgba colors

JosephPecoraro:
Issue #3625


marijnh:
I guess this is minor enough to include even before widespread browser support shows up. Merged as 292a12a

PR #3627, opened 2015-11-05
Fix material.css cm-error state

cjolly:
it seems `.cm-error` lost some specificity in [this commit](https://github.com/codemirror/CodeMirror/commit/a7d443ec82c1806d24e25ad8014a4d8651409320#diff-33646e9ee4d04eedb98a991a636c2302L35) that causes the error state to have same color as the tag.

c22ad46c-7656-11e5-808f-b3136875428a

EDIT: As I look at that commit again, I don't think that was actually the problem - I think it's always been an issue. And `cm-tag` seems to have to rules as well.


marijnh:
Thanks! Merged as fec63b9

PR #3631, opened 2015-11-09
added even/odd classes to generated lines

toni07:
Hi, I added even/odd class to generated lines, so I'm not forced to wirte tricky CSS for zebras in lines (such as: .CodeMirror-code > div:nth-child(even) > .CodeMirror-line{background:#dddddd;}, that works well, but won't work if you change the DOM elements hierarchy, and this CSS code is ugly).
Sorry I have no time to test (and don't know how to do it).

Thanks for this amazing library!

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -867,7 +867,7 @@
       var lineView = view[i];
       if (lineView.hidden) {
       } else if (!lineView.node || lineView.node.parentNode != container) { // Not drawn yet
-        var node = buildLineElement(cm, lineView, lineN, dims);
+        var node = buildLineElement(cm, lineView, lineN, dims, i);
         container.insertBefore(node, cur);
       } else { // Already drawn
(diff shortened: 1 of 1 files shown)


marijnh:
This is broken (see test runner results). But even if it weren't, it is not a good idea. If you press enter inside the editor, the lines below the cursor will shift down, without having their DOM redrawn, and thus the classes will be out of sync. I think you should stick to using CSS tricks.

toni07:
Hi, how can I see the test runner results? You may be right for CSS, actually I just use this lib as a readonly component, so I don't need to support the redraw.
Maybe I should keep my own changes, sorry for disturbing.
Thanks

marijnh:
Install node and npm, and then do `npm install` inside of your CodeMirror checkout to install the dependencies, and `npm test` to run the tests.

toni07:
Great, thanks!

PR #3632, opened 2015-11-09
Bug correction

toni07:
Sorry I created a bug in my last commit.


marijnh:
The tests still aren't passing. See my comment on #3632

Issue #3633, opened 2015-11-11
IE11 - Stale reference to variable "measureText" can cause permission denied errors from deleted iframes


kevjames3:
This can be a pretty obscure case, but because of IE11's security setup (or bug depending on your preference), the variable measureText is holding a stale reference that can make IE11 throw a permission denied error.

Platform: IE11.0.9600
Version 5.8.1

To replicate, do the following:
1. Create an iframe with a text box inside it.
2. Initialize CodeMirror on the text box
3. Programmatically delete the iframe
4. Try to re-initialize CodeMirror on another item (text box somewhere on the page, does not matter where)

Error will be "Permission denied" on CodeMirror/lib/codemirror.js: Line 2937.

Issue is the following code... (starting at line 2933)
```javascript
  var measureText;
  // Compute the default text height.
  function textHeight(display) {
    if (display.cachedTextHeight != null) return display.cachedTextHeight;
    if (measureText == null) {
     //.....
```
`var measureText` can be holding a stale reference to a child of a deleted element, which FF and Chrome can handle, but IE sees that as a DOM element that no longer be referenced and is a security risk(?).  

The change to fix this is simple.  Simply scope `var measureText` into the `function textHeight(display)` and remove the conditional check `if (measureText == null) {` as it is no longer needed.  Code is below:

```javascript
  function textHeight(display) {
    if (display.cachedTextHeight != null) return display.cachedTextHeight;

    var measureText = elt("pre");
    // Measure a bunch of lines, for browsers that compute
    // fractional heights.
    for (var i = 0; i < 49; ++i) {
      measureText.appendChild(document.createTextNode("x"));
      measureText.appendChild(elt("br"));
    }
    measureText.appendChild(document.createTextNode("x"));

    removeChildrenAndAdd(display.measure, measureText);
    var height = measureText.offsetHeight / 50;
    if (height > 3) display.cachedTextHeight = height;
    removeChildren(display.measure);
    return height || 1;
  }
```

marijnh:
Using CodeMirror in a window other than the one in which it was loaded isn't supported -- I'm highly surprised it worked at all (it'll fail completely on older versions of Firefox). Until I get around to having a different window per editor (stored as option, used everywhere instead of the global window), that's just the way things are. I don't think your patch is a full solution, and it definitely is a performance liability.

kevjames3:
[...]

PR #3635, opened 2015-11-11
Scoping measureText to be inside the function textHeight to eliminate stale references (IE11)

kevjames3:
… textHeight to eliminate stale references

See issue #3633


kevjames3:
This pull request addresses #3633

PR #3636, opened 2015-11-11
change selected item in hint options on mousemove

continuata:
### Issue
currently only cursor keys change the highlighted option and allow hitting Enter to select it.
If the user moves the mouse over the select box, there is no visual feedback (unless hover styling is applied) to show the option that will be selected when clicked. If hover styling is applied, 2 items in the list will be highlighted; the active one and the hovered one.

### Solution
this PR allows the user to switch between keyboard and mouse navigation of the hint options by using the mouse move event to change the selected option.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -295,6 +295,13 @@
       setTimeout(function(){cm.focus();}, 20);
     });
 
+    CodeMirror.on(hints, "mousemove", function(e) {
+      var t = getHintElement(hints, e.target || e.srcElement);
+      if (t && t.hintId != null) {
+        widget.changeActive(t.hintId);
+      }
+    });
+
     CodeMirror.signal(data, "select", completions[0], hints.firstChild);
     return true;
   }


marijnh:
How about a pure-css solution instead, where you style `.CodeMirror-hints:hover .CodeMirror-hint` to not have a background and style `.CodeMirror-hint:hover` (with some other stuff to give it precedence) as having a highlighting background? This solution would cause things like accidentally sweeping the mouse over the widget to change the selection. That's not a huge deal, since people don't tend to move their mouse while they are selecting something with the keyboard, but it seems that the CSS approach is more lightweight and less invasive.

continuata:
We were using `.CodeMirror-hint:hover`.
The problem isn't styling the hover, it's the fact that both the hover and the active element are highlighted, which is confusing for the use.
I don't see any CSS way around that.
screen shot 2015-11-11 at 14 46 33

marijnh:
Yeah, that's what the first part of my response addressed -- you can disable the normal highlighting when the mouse is hovering over the widget using just CSS. But on closer inspection, that is not ideal. If the user tries to use the arrows while the mouse is still on the widget they wouldn't be able to see their selection.

I tried out your patch, but I think it makes things more confusing. We're expecting a double click to select a completion, and controls that show a hover behavior like this seem to suggest that a single click is enough.

So I merged this as 5a6f158eedf and followed up with patch 1de8132, which only enables this behavior when `completeOnSingleClick` is on, but also makes that the default.

continuata:
ah yes ... sorry we use single click, so I hadn't considered the double-click use-case. 
That seems like a good solution to me now.

vg6193:
Hello guys,
I am facing the same issue. Could you please advice what I should do.

autocomplete

PR #3640, opened 2015-11-12
Typo

mihailik:
Apart from affecting markdown feature, it can break whole CM in older browsers (like IE8).

In my case I've inlined CM with a bunch of addons, and got IE syntax error, rendering whole CM inlined script unparseable. That's because `s.this` uses risky `this` in place of property name.


marijnh:
Thanks! Merged as 7530191

PR #3642, opened 2015-11-14
[dart mode] add "as" to keywords

goderbauer:
`import "package:foo/foo.dart" as bar;` should highlight "as" as a keyword.


marijnh:
Merged as e477745

PR #3643, opened 2015-11-14
[dart mode] in annotations "." is legal

goderbauer:
`@ng1.Injectable` is a legal annotation


marijnh:
Thanks. Merged as a1e780b

PR #3645, opened 2015-11-16
haml + liquid: Current behavior sees the closing %} as a haml tag begin.

joshco:
The result is that the rest of the file after the %} doesnt highlight correctly.
Proposed fix: Dont go into quote mode when stream is '{%'.


marijnh:
Looks good. Merged as 2a84b59

joshco:
cool. thanks!

PR #3649, opened 2015-11-18
Fix JSDoc comment

kayur:
According to http://usejsdoc.org/about-getting-started.html, jsdoc comments must start with /**


marijnh:
Merged as 5bf7669. We're not really using JSDoc, but I agree that if there are going to be comments looking like JSDoc, we might as well make them properly formed.

PR #3657, opened 2015-11-20
[annotatescrollbar addon] Add an id attribute to annotations

alur:
I mainly want this so that I can bring the annotation into view when it's clicked. It's useful when files are large enough that finding the annotation by hand would be difficult.

### DIFF
--- a/addon/scroll/annotatescrollbar.js
+++ b/addon/scroll/annotatescrollbar.js
@@ -100,6 +100,9 @@
       elt.style.cssText = "position: absolute; right: 0px; width: " + Math.max(cm.display.barWidth - 1, 2) + "px; top: "
         + (top + this.buttonHeight) + "px; height: " + height + "px";
       elt.className = this.options.className;
+      if (ann.id) {
+        ann.setAttribute("annotation-id", ann.id);
+      }
     }
     this.div.textContent = "";
     this.div.appendChild(frag);


marijnh:
Since the surrounding div has `pointer-events: none`, I'm not sure how such a handler would ever be activated. Also, why add handlers to each individual element, rather than just registering a single handler (outside of the plugin) that notices when the event target is one of these annotations?

alur:
I just added pointer-events: all to the css class for the individual markers.

Good point about using a single handler. I would still like a way to get the original annotation object from the click though. Would just setting elt.annotation = ann; be acceptable?

Alternatively, I could set elt.annotation = ann; and move the listener to this.div, to keep the interface a little cleaner while only having one event handler per annotation type.

marijnh:
> Would just setting elt.annotation = ann; be acceptable?

That would be a bad idea -- adding JS properties to DOM nodes makes them a bunch more expensive for the browser to represent, and since this is a kind of node that'll be created a lot, it seems wasteful to do this. Would adding `id` properties to your annotation objects, and having the addon check for those and set an `annotation-id` attribute on the node, work for you?

alur:
That should work great. I updated the push request.

marijnh:
Great. Merged as 6eef0e0

PR #3659, opened 2015-11-21
[sql mode] Add `limit` to default SQL keyords

cosinequanon:


marijnh:
Thanks. Merged as 446d89d

Issue #3655, opened 2015-11-20
[mobile] Colouring not picking changes until non-word characters


mihailik:
Using [JavaScript demo](http://codemirror.net/mode/javascript/) go in the method and type 'this'.
The text is using basic black colour and 'this' word is underlined.
Now type '.'.
The word 'this' is no longer underlined, and it's dark-purple (correct keyword colour).

This happens in:
* Chrome mobile (46.0.2490.76 on KitKat),
* Android Browser (33.0.0.0 on KitKat),
* FireFox mobile (42.0 on KitKat).

It also is dependent on the order of the edits. For example, deleting characters with backspace never causes this quirk.

Also note that when this happens in the middle of the line, the characters on the right of the line do get  pushed right. Which means this typing experience is not completely virtual, it does change the DOM state somehow. But it looks like CM fails to notice the change.

Also see related:
#3653 *Arrow keys hide keyboard - Chrome mobile 46*
#3654 *Newline/editing bugs in Android Browser*

marijnh:
Mobile platforms tend to input words using composition events. Highlighting tends to involve redrawing, and redrawing during a composition will reset that composition, which would be bad, so CodeMirror has to leave the input field alone until the composition ends (in this case, when the space was typed).

As such, I don't think this is something we can really help.

mihailik:
Reasonable, thank you!

Might be worth mentioning in the manual, also clarifying if CodeMirror merely backs off highlighting during composition, or if it doesn't fire 'change' events either.

marijnh:
It currently does not fire change events, but possibly will in the future (I'm planning to overhaul composition handling when I find time), so I don't want to set down any specific behavior yet.

mihailik:
Reasonable too.

A bigger problem is completions/hints (which some IDEs pop up intelligently listening to the 'changes').

Typing this `var div = document.crea|` will sure have keyboard app suggestions like `create` `created` `creator`, but not context-aware IDE suggestions like `createElement` `createTextNode`. And on desktop it indeed works like that.

Perhaps for code composition is harmful anyway. Typing `foreach` you get `For each` instead and such other.

Can you please add a switch to ignore composition concerns and proceed with change handling?

Again, thanks for giving it a thought.

marijnh:
> Can you please add a switch to ignore composition concerns,

Nope. Or at least, I don't know how to disable this behavior in mobile browsers. If you do, let me know.

[...]

Issue #3653, opened 2015-11-20
On Chrome mobile arrow keys hide keyboard


mihailik:
Standard Google on-screen keyboard on Android has no arrow keys, but some other keyboards do (we care for on-screen keyboards here).

When arrow keys like <-- or --> are pressed, the cursor moves, but the onscreen keyboard hides.

That happens on:
* Chrome mobile (46.0.2490.76, on KitKat)

That does not happen on:
* FireFox mobile (42.0, KitKat)
* Android Browser (KitKat), which reports itself as Chrome/33.0.0.0

From other clues it looks like FireFox uses full 'contenteditable' [inputStyle](http://codemirror.net/doc/manual.html#option_inputStyle), whilst Android Browser seems to fall back onto 'textarea'.

Generally I've never had such problems in 'textarea', but ever since 'contenteditable' was introduced, the arrow keys have always been like that on Chrome.

Haven't got an iPhone, so cannot check Safari.

marijnh:
I'm getting contenteditable as default for the `inputStyle` option on the android browser. Could you tell me the userAgent string your stock android browser produces?

mihailik:
Mozilla/5.0 (Linux; Android 4.4.4; UMI eMAX Build/KTU84P) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/33.0.0.0 Mobile Safari/537.36

marijnh:
That one's recognized as a mobile browser by CodeMirror, so barring some really weird thing happening, you should get `"contenteditable"` as default input style.

adrianheine:
From what I understand, this issue seems to be fixed. In any case, we are working on a rewrite (CodeMirror 6) that will address this and similar issues more thoroughly, and we are currently raising money for this work: See [the announcement](https://codemirror.net/6/) for more information about the rewrite and a demo.

Note that CodeMirror 6 is by no means stable or usable in production, yet. It is unlikely that we put much work into fixing this and similar problems for CodeMirror 5, though.

PR #3664, opened 2015-11-23
Introducing options.updateWhenComposing

mihailik:
This option (defaulting to `false`) makes CodeMirror ignore composing events, which switches off mobile keyboard suggestions and makes mobile experience align closely with the desktop for highlighting and change handling.

The purpose of the option is rich code editing and IDEs, which provide custom completions and rich contextual services.

These IDE features tend to conflict with mobile touch keyboard suggestions.

The key downside is IME is disabled or severely handicapped when this option is enabled. Given that most programming languages tend to stick with English and ASCII, lack of IME is often a reasonable tradeoff.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1299,21 +1299,23 @@
         if (!eventInWidget(display, e)) e_preventDefault(e);
       });
 
-      on(te, "compositionstart", function() {
-        var start = cm.getCursor("from");
-        if (input.composing) input.composing.range.clear()
-        input.composing = {
-          start: start,
-          range: cm.markText(start, cm.getCursor("to"), {className: "CodeMirror-composing"})
-        };
-      });
-      on(te, "compositionend", function() {
-        if (input.composing) {
-          input.poll();
-          input.composing.range.clear();
-          input.composing = null;
-        }
-      });
+      if (!cm.options.updateWhenComposing) {
+        on(te, "compositionstart", function() {
+          var start = cm.getCursor("from");
+          if (input.composing) input.composing.range.clear()
+          input.composing = {
+            start: start,
+            range: cm.markText(start, cm.getCursor("to"), {className: "CodeMirror-composing"})
+          };
+        });
+        on(te, "compositionend", function() {
+          if (input.composing) {
+            input.poll();
+            input.composing.range.clear();
+            input.composing = null;
+          }
+        });
+      }
     },
 
     prepareSelection: function() {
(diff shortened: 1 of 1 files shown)


mihailik:
See also discussion in #3655 *[mobile] Colouring not picking changes until non-word characters*

mihailik:
I've created a gist with a CodeMirror using this option and HTML mode:

[Gist hosted with githack](https://gist.githack.com/mihailik/a8baccbeac8eae9a1be6/raw/3fe15ab4d33c1fcfb674002e2f098a26c6098101/index.html)

[Gist itself](https://gist.github.com/mihailik/a8baccbeac8eae9a1be6)

URL shortener to quickly type on mobile:
http://tinyurl.com/updateWhenComposing

marijnh:
This seems to sort of work for the textarea [inputStyle](http://codemirror.net/doc/manual.html#option_inputStyle), but when using contenteditable it creates a mess of latin and Japanese characters when I try to use it with Japanese IME.

Also, in general, I'm rather hesistant to include such an accessibility-hostile feature in the core editor. Have you tried adding your own event handlers to somehow force composition to end as soon as it starts? That sounds like it would also be possible, and doesn't require changing the library.

mihailik:
IME is an expected downside, insignificant in the target use case: *mobile code editing.*

I am not aware of any accessibility issues. Both reading and writing should work just the same.

mihailik:
To put it into context, currently **code editing on mobile** with CodeMirror is a major pain. Because of autocorrection, lack of completion and #3653 (troubles with arrow keys).

Comparing to that, troubles with occasional ideographic in string literals/comments is very minor.

But you've mentioned composition refactoring you're planning. Would that bring back 'change' events during composition? Would you elaborate please?

Many thanks!

marijnh:
The rewrite will, if it works out the way I planned it, have the editor fire change events during composition.

In my test with your option, there we no _occasional_ problems with IME, it was just completely broken -- inserting both the composed character and the latin variants next to each other. The code is written with the expectation that compose events are handled, and just disabling those handlers doesn't seem like it will do the right thing.

mihailik:
Sure, if a better solution is in the works let's close this one.

marijnh:
Well, the thing is, that other solution isn't something that is going to happen soon, since I have much other things to do and this isn't very high-priority. So waiting for it is likely to end with frustration.

mihailik:
Only a bit: I am keeping this option in my forked code, so it's fine in my app :-)

Or shall I try to implement that other solution?

I assume the crux of the plan is to tweak the line rendering. If composition is on (plus some extra checks), rendering will update existing elements in place, instead of recreating from scratch.

marijnh:
The plan is to just disable line re-rendering for the line that is being composed (since the entered characters will already show up), and force a composition end if something else (say, a script) tries to modify that line. In addition, we have to read the actual characters that are in the composition from the DOM, since the events often contain unhelpful or even plain bogus data, and immediately update the document on each compositionupdate. I expect it's going to a bit hairy to implement, but if you want to take a stab at it, that'd be very helpful.

mihailik:
Thanks, I shall stab indeed :-)

PR #3666, opened 2015-11-24
Add missing keywords for typescript mode

elisee:
Nothing fancy, just a bunch of keywords that were missing: `enum`, `module` and `namespace`, the `abstract` modifier, the `implements` directive and the `as` operator.

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -47,13 +47,21 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
         // object-like things
         "interface": kw("interface"),
         "extends": kw("extends"),
+        "implements": kw("implements"),
         "constructor": kw("constructor"),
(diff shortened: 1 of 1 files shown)


marijnh:
So none of these (including the old TypeScript keywords) seem to be implemented in the parser. This will confuse the highlighting, since in many cases (local variables, template strings, regexps) that depends on the parser actually understanding the current context of the token.

Have you tried, instead of doing things like `kw("implements")`, which isn't something the parser is going to know what to do with, to actually use one of the existing syntactic roles (in this case `C`) that is close to how the keywords work?

marijnh:
Guess not. Merged as 026ad081b16c, followed up by 7cde7a70940e8

elisee:
Hey sorry for the lack of an answer. I don't know much about CodeMirror's internals so I thought I'd take a look later and forgot. Thanks for taking care of it!

PR #3669, opened 2015-11-24
Fix 5.9 release date in release notes

vincentwoo:


marijnh:
Oops. Thanks for catching that. Merged as f14b2f8

PR #3673, opened 2015-11-28
Fix mobile shift arrows

mihailik:
This includes #3672 *Fixing hidding keyboard on arrow keys on mobile.*
extending it with support of Shift-selection.

The crux of the problem (at least on Chrome mobile and Opera) is `selection.removeAllRanges`, which causes the keyboard to hide. Assuming that intermittent selection loss hints the keyboard is not needed.

The fix (after previous more complex logic turned into one simple method):
* IF running in gecko  OR  editor is  NOT focused,
 * use old method `removeAllRanges` + `addRange`
* ELSE
 * `collapse` selection to the beginning of new selected range
 * IF new selected range is non-empty, `addRange`

That so far fixes all the onscreen keyboard quirks for Chrome/Opera mobile.
FireFox acts weird, so the logic is left doing old `removeAllRanges`/`addRange`. It doesn't feel hopeless though, need to debug more on FF.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1692,18 +1692,13 @@
       try { var rng = range(start.node, start.offset, end.offset, end.node); }
       catch(e) {} // Our model of the DOM might be outdated, in which case the range we try to set can be impossible
       if (rng) {
-        if (gecko && this.cm.state.focused && start.offset == end.offset) {
+        if (!gecko && this.cm.state.focused) {
           sel.collapse(start.node, start.offset);
+          if (!rng.collapsed)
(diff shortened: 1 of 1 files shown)


mihailik:
I've added one more exception for FireFox. It acts too weird, hiding keyboard sporadically -- best to leave it with simple `removeAllRanges` + `addRange` as is for now.

marijnh:
Thanks, that clarifies the reasoning. Merged as 4bdf644. Mobile Firefox support is also still flaky on a number of other points, so the lack of support for that does not bother me too much.

Issue #3030, opened 2015-01-14
'Go to line' command is missing in default key map


davido:
Is there any obvious way, that i am missing, to jump to a line in default key map? Something that this addon addresses: [1], [2], that makes use of interactive dialog feature?

Any reason why similar functionality is not included in CodeMirror distribution?

* [1] http://djadmin.github.io/addon/search/goto-line.js
* [2] http://djadmin.github.io/

marijnh:
Nope, there isn't. There is no sane way to prompt for a number in the core library, so this isn't part of the core. You can write it yourself in a few lines of code, though:

```javascript
CodeMirror.commands.jumpToLine = function(cm) {
   var line = Number(prompt("Where", ""));
   if (!isNaN(line)) cm.setCursor(line, 0);
};
CodeMirror.keyMap.default["Alt-G"] = "jumpToLine";
```

(For extra points, use the dialog addon to make the prompt less ugly.)

mihailik:
Somebody's already done it, including the bonus dialog:

http://djadmin.github.io/

Specifically the addon code:

https://github.com/djadmin/djadmin.github.io/blob/master/addon/search/goto-line.js

Looks worthy, should it be added to the CM addon tree?

dpursehouse:
The reason we didn't add it in Gerrit is that the license is unclear.  See the commit message of @davido's commit that was referenced above.

davido:
Indeed, @dpursehouse meant this non answered question to clarify the license: [1].

* [1] https://github.com/djadmin/djadmin.github.io/commit/6ee5941512d73cdcede929b6b832896842e0845d#diff-5

mihailik:
OK, shell I implement it close to what @marijnh suggested, separate from @dpursehouse 's take?

* Use Alt-G as a shortcut
* Parse single number: `line`
* Parse colon-separated: `line:character` or relative to the current line `:character`
* Parse plus-minus offsets relative to the current position: `+line`, `-line`
* Parse percentages: `30%` - those take in account line counts, not character size

How does that sound?

jenspohle:
@marijnh:

at the beginning of this thread you posted some lines of code, that shall make the "goto-line-function" work.

Well, javascript is not my strong ;-)

Can you tell me, in which files i have to place it, to make it working?

Thank you very much in advance! :-)

djadmin:
Thanks guys for adding the `Goto line` command to CodeMirror :+1:

marijnh:
[...]

PR #3682, opened 2015-12-04
Jumptoline addon and added yo search demo.

mihailik:
See [#3030 for previous discussion](https://github.com/codemirror/CodeMirror/issues/3030#issuecomment-161665462)

This adds jumpToLine.js addon and jumpToLine command bound to Alt-G as @marijnh suggested.

The accepted inputs are:

* `linenumber` (one-based)
* `:linenumber` same as above, to match habitual format for 'go to line' in Chrome DevTools
* `+linenumber` and `-linenumber` for line relative to current
* `line:column`
* `%scroll` for position on scrollbar (proportional to line count, not character count neither pixel height)
* `+scroll%` and `-scroll%` for position on scrollbar relative to current

In case when character offset/column is not specified, we jump to line preserving offset/column as possible.

### DIFF
--- a/addon/search/jumpToLine.js
+++ b/addon/search/jumpToLine.js
@@ -33,7 +33,7 @@
       if (!posStr) return;
 
       var clnMatch = /^\s*([\+\-]?\d+)\s*\:\s*(\d+)\s*$/.exec(posStr);
-      var prcMatch = /^\s*([\+\-]?\d+(\.\d+))\%\s*/.exec(posStr);
+      var prcMatch = /^\s*([\+\-]?\d+(\.\d+)?)\%\s*/.exec(posStr);
       var lnMatch = /^\s*\:?\s*([\+\-]?\d+)\s*/.exec(posStr);
       if (clnMatch) {
         try {
@@ -48,7 +48,7 @@
       else if (prcMatch) {
         try {
           var prc = parseFloat(prcMatch[1]);
-          var line = cm.lineCount()*prc/100;
+          var line = Math.round(cm.lineCount()*prc/100);
           if ('+-'.indexOf(prcMatch[1].charAt(0))>=0)
             line = cur.line+line+1;
         }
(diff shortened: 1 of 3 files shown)


mihailik:
@dpursehouse @davido

Does this look OK?

marijnh:
Looks good to me. I've merged it as f1519ad5057 , and cleaned it up a little in 291b06eb8. The only externally visible change I made is that I renamed the file to `jump-to-line.js` to conform to the way other addons files are named.

dpursehouse:
Nice.  Any idea when this will be in a release?

marijnh:
I do releases around the 20th of each month, so in two weeks 5.10 will be coming out.

davido:
Nice job indeed. Really appreciated.

dpursehouse:
Then we can get it into Gerrit 2.13. I don't think we can wait 2 weeks to
release 2.12.

davido:
Ack.

mihailik:
Thanks a lot @marijnh ! I've spotted a couple of things:

1. zero-based character offsets — this is the most important, as majority of editors and tools use one-based line/columns, so copy-pasting positions will be ever so slightly off
2. remnants of search.js addon docs in the comment/header
3. semicolons — are you intentionally migrating to no-semicolon rule? it's a bit jarring when only a couple of lines skip them and others stick to old-style semicolon

Look at #3683 *Cleanup of the header, consistency tweaks* for fixes to (2) and (3).

I wasn't sure if (1) is intentional, so didn't fix it yet. But please consider it, as it might make CodeMirror feel buggy if it always lands you one-character off what you expect.

marijnh:
> zero-based character offsets

For what it's worth, Emacs uses zero-based column offsets as well, which I feel makes sense (before the first character is position zero, after the first character position 1), and which has actually never caused problems with copy-pasted positions from other tools for me -- gcc, babel, and everything else I routinely use spits out zero-based columns.

Since zero-based is what CodeMirror uses, I don't think it is a good idea to make this addon so something different.

> are you intentionally migrating to no-semicolon rule?

Yes. I don't think it makes sense to update existing code yet, though there'll be a big move to ES6 modules and ES6 code in the future, at which point I'll remove all semicolons and start enforcing semicolon-free style.

Thanks for your PR!

PR #3686, opened 2015-12-05
[mscgen mode] simplifies the keyword recognizing regexps

sverweij:
- for our purposes `(=>|<=|...)` works just as well as `((<=)|(=>)|...)`


marijnh:
Makes sense. Merged as 382f51d

Issue #3687, opened 2015-12-05
[javascript mode] Regex inside template literal


nicolo-ribaudo:
```js
`${/./}`;
```

It is parsed as string-2, operator, `.`, operator, `}`, string-2

PR #3688, opened 2015-12-05
[javascript mode] Recognize regex inside template literal

nicolo-ribaudo:
Closes gh-3687


marijnh:
Thanks! I've pushed a slightly revised patch 764022a which doesn't assume that there is no whitespace between the `{` and the `/`.

PR #3689, opened 2015-12-08
:memo: give the command a nice name

basarat:
Similar to *all other custom commands* in the same file. Good to have a `name` that users can refer to (and IDE developers can show in their command registry) :rose:

refs https://github.com/TypeScriptBuilder/tsb/issues/50

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -243,7 +243,7 @@
     });
   };
 
-  map[ctrl + "/"] = function(cm) {
+  cmds[map[ctrl + "/"] = "toggleCodeComment"] = function(cm) {
     cm.toggleComment({ indent: true });
   }
 


marijnh:
This is not the only binding in sublime.js that doesn't have a name, and there is already a command `toggleComment`, defined in `addon/comment/comment.js`, which does something very similar to this one. Thus, I don't think adding another command with a very similar name is a good idea.

basarat:
> This is not the only binding in sublime.js that doesn't have a name, and there is already a command toggleComment, defined in addon/comment/comment.js

This PR has `toggleCodeComment` which is different from `toggleComment` [comment/comment.js](https://github.com/codemirror/CodeMirror/blob/master/addon/comment/comment.js). Neverthless I can rename if further if you want, or we can leave this closed :rose:

> This is not the only binding in sublime.js that doesn't have a name,

I searched `map[` in https://github.com/basarat/CodeMirror/blob/1231ce03db3cef8d014fad9fc4b64daab96dd9d1/keymap/sublime.js but all instances other than this one point to `string`s e.g. `map["Shift-" + ctrl + "["] = "fold";` or  `cmds[map[ctrl + "F3"] = "findUnder"] = function(cm) { findAndGoTo(cm, true); };`.  Am I mistaken? :rose:

marijnh:
Fair enough. I wasn't awake enough last night to read the code it seems. Attached patch names the command `toggleCommentIndented`.

basarat:
Thanks :rose: :100:

PR #3702, opened 2015-12-14
Link text runmode script in the correct location

ZombieHippie:
There is a typo on the runmode page which refers to runmode being in the `lib` folder.


marijnh:
Hah. In some gray past the file was under `lib/`. Thanks for noticing that. Merged as b7f0785

PR #3703, opened 2015-12-14
Accept either element node, or document fragment for runmode

ZombieHippie:
Note we are using document fragments for updating selection and cursors already.

A document fragment, in short, is a parentless element.


marijnh:
Could we just check whether it has an `appendChild` property instead?

ZombieHippie:
@marijnh Yeah, that would work too.

```
if (typeof callback.appendChild === "function") {
```

PR #3706, opened 2015-12-14
Fix typo in manual

McBrainy:
It took me forever to figure out why vim mode wasn't working, and all because the manual said to set the "keymap" option and not the "keyMap" option. I also added quotes to make it clear that "vim" is a string.


marijnh:
Thanks for taking the time time submit a PR and protecting the next person from getting confused. Merged as 188ae75

PR #3707, opened 2015-12-15
[django mode] better highlighting of in/and/or/not

TDaglis:
Currently operators in most themes look like normal text because cm-operator doesn't usually get special styling.
Following the example of https://github.com/codemirror/CodeMirror/pull/3194 for python mode word operators, django mode word operators (in/and/or/not) are now highlighted as keywords to make them stand out better.

This is what the change looks like for a sample template

#### Before
before
#### After
after


marijnh:
Thanks. Merged as 0dbe0ef

PR #3708, opened 2015-12-15
Attach CodeMirror to the window explicitly

gcziprusz:
When minified the context could change breaking the global export of CodeMirror.


gcziprusz:
More specifically -> other libraries may declare strict mode and in ES5 strict `this` in an annonymus IIFE is undefined instead of referring to the global window.

marijnh:
The use of `this` here is intentional. But attached patch should solve the strict mode issue.

gcziprusz:
:+1:  works for us

PR #3712, opened 2015-12-16
[crystal mode] Add

MakeNowJust:
I implemented [Crystal](http://crystal-lang.org/) mode for CodeMirror. Crystal is one of the Ruby inspired languages. It is compiled to native binary and is faster than Ruby.


marijnh:
Thanks! Looks good. Merged as c7b64ca080df

PR #3713, opened 2015-12-16
powershell mode for codemirror

y3llowcake:


marijnh:
Thanks. But your demo highlights most of the code in the example in red, suggesting the mode doesn't quite work as intended yet. Could you look into that?

Also, the demo HTML should conform to the other demos, and the code for the mode should be indented consistently.

Dominator008:
This should be closed in favor of https://github.com/codemirror/CodeMirror/pull/3911, which has been integrated.

PR #3714, opened 2015-12-16
[markdown mode] fix escaped brackets in link def

jscissr:
Example: `[foo\[bar\]foo]: https://example.com`


marijnh:
Thanks! Looks good. Merged as c39008a

PR #3715, opened 2015-12-17
Fix options of doc.extendSelections.

just95:
The options argument of `doc.extendSelections` has currently no effect, because it is passed to `clipPosArray` instead of `extendSelections`.


marijnh:
Indeed! Merged as d9042e7

PR #3716, opened 2015-12-17
Update manual entry of extendSelectionsBy.

just95:
The manual entry of `extendSelectionsBy` currently states that the first argument is expected to be a function which maps a `{anchor, head}` range to another range. The actual implementation expects the function to return `{line, ch}` positions.


marijnh:
Thanks for noticing that! Merged as 78e3ac3

PR #3717, opened 2015-12-17
doc.extendSelectionsBy should clip extended ranges.

just95:
The mapped ranges of `doc.extendSelectionsBy` should be clipped before `extendSelections` is invoked. 
This avoids an exception to be thrown if one of the ranges falls outside of the document. Additionally this is more consistent with the behavior of `doc.extendSelections`.


marijnh:
Agreed. Merged as 420cefd

PR #3722, opened 2015-12-19
Add mode for GitHub Flavored Markdown with YAML front matter

cllu:
Markdown with YAML front matter is widely used in various Jekyll-like static website generator. This mode is developed when I create [Semantic Resume](https://semantic-resume.chunlianglyu.com/), where I cannot find existing implementations. I hope it would be useful for others.


marijnh:
This looks good, but I feel this would work better as a mode not hard-coded to Markdown (though I guess that's where YAML front matters are most commonly used). So I've merged this as a4e987079517e and modified it in d6212b9216e4a to be more general, renaming it to `yaml-frontmatter`.

PR #3724, opened 2015-12-21
[crystal mode] Fix page title

MakeNowJust:
Sorry :cry: . I forgot to fix this in copied and pasted `index.html` from Ruby's.


marijnh:
Hah. I was just about to push my own patch that fixed those. Merged yours as 6bc4c5a

PR #3727, opened 2015-12-22
yacas mode

grzegorzmazur:


marijnh:
(You can push to pull request branches -- no need to create a new one every time you change it.)

grzegorzmazur:
Right, thanks for the hint and sorry for the noise, still learning how to use git(hub).

marijnh:
This seems to just highlight every identifier as a keyword. The token `variable` would be more appropriate for that, and the language probably does have a concept of a 'keyword' -- a fixed list of words that have special syntactic meaning -- which you could recognize.

grzegorzmazur:
That's right, the language is very flexible and the notion of keyword doesn't really make sense in it. Yacas  is quite similar to mathematica and that's why I followed the mathematica mode, even though I didn't actaully like it very much. Anyway, I'd rather avoid defining a list of keywords, although such a list would actually work quite well in most cases. Still, things can be improved - in the current version I've made a difference between thing that look like function call (it's either a function call or a rule defining transformation for function calls) and all other identifiers; the former are recognized as keywords and the latter as variables.

marijnh:
Okay, that makes sense. Still, I'd use something other than "keyword" to style 90% of the text -- CodeMirror themes aren't really designed to look good when everything is a keyword.

marijnh:
I'm not notified when a new commit is added to a pull request, so it's a good idea to always follow up with a comment. Now it seems like there's a bunch of irrelevant patches in the history. Could you clean that up and let me know, so I can merge this? (You'll want to do something like `git rebase -i --root my-branch --onto master`.)

grzegorzmazur:
Ups, ok, I wasn't aware of that. Anyway, I've cleaned up the history (thanks for the instructions on how to do that), so hopefully it's in good enough shape now.

marijnh:
Thanks. Merged as be36e3e

PR #3730, opened 2015-12-22
Add mode literate haskell

martinstarman:
This is my first attempt to write CodeMirror mode. I just chose literate haskell from mode whish list. Thanks for your comments.

### DIFF
--- a/mode/haskell-literate/haskell-literate.js
+++ b/mode/haskell-literate/haskell-literate.js
@@ -1,0 +1,38 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function (mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"), require("../haskell/haskell"))
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror", "../haskell/haskell"], mod)
+  else // Plain browser env
+    mod(CodeMirror)
+})(function (CodeMirror) {
+  CodeMirror.defineMode("haskell-literate", function (config) {
+    var haskellMode = CodeMirror.getMode(config, "haskell")
+    return {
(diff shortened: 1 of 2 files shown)


mode/haskell-literate/haskell-literate.js
```diff
@@ -0,0 +1,42 @@
...
+        }
+      },
+      copyState: function (state) {
+        return {
+          haskellState: CodeMirror.copyState(haskellMode, state.haskellState)
+        }
+      },
+      token: function (stream, state) {
+        var s = stream.next()
+        if (s == '>' || state.haskellCode) {
```
marijnh: You should probably also check that `stream.sol()` (start-of-line) at the start of the function, so that you only respond to `>` characters at the start of the line. Or am I missing something?

mode/haskell-literate/haskell-literate.js
```diff
@@ -0,0 +1,42 @@
...
+          state.haskellCode = true
+          return haskellMode.token(stream, state.haskellState)
+        } else {
+          while (!stream.eol())
```
marijnh: `stream.skipToEnd()` is faster

mode/haskell-literate/haskell-literate.js
```diff
@@ -0,0 +1,42 @@
...
+            stream.next()
+          return "comment"
+        }
+      },
+      blankLine: function (state) {
+        state.haskellCode = false
+      }
+    }
+  })
```
marijnh: Modes that wrap other modes need to implement an [`innerMode`](http://codemirror.net/doc/manual.html#innerMode) method, to make things like completion and folding aware of the actual mode.

martinstarman:
@marijnh Thanks for your comments. I change the code, but i dont know if I'm using ```innerMode``` correctly.

marijnh:
Thanks! I've merged this 4955e6e018837 and integrated and generalized it a bit in 07bcf88d86

martinstarman:
@marijnh Thanks!

PR #3740, opened 2015-12-26
Add background-blend-mode to CSS

sprintr:
Adds background-blend-mode property and [associated values](https://developer.mozilla.org/en-US/docs/Web/CSS/blend-mode) to CSS mode.


marijnh:
Merged (without failing tests) as d395fec651fd31

PR #3741, opened 2015-12-27
Support Clojure Ratios.

ctford:
Clojure has Ratio literals with an integer numerator and denominator, e.g. `1/2`. Currently, the Clojure mode only recognises the numerator as a number and therefore messes the highlighting up on the rest of the Ratio.

This patch fixes that by accepting `/` in the middle of a number as an alternative to `.`.


marijnh:
Thanks! Merged as f963146

PR #3742, opened 2015-12-27
Introduce JSX mode for CodeMirror

### LINKED ISSUE #3122: JSX support?

parisk:
This PR introduces a JSX mode for CodeMirror and hopefully closes #3122.

Because of JSX' idiomatic syntax, I preferred implementing this as a completely different mode, than interfering with any of the existing JavaScript or XML modes, despite this was one of @marijnh's proposals at #3122 (https://github.com/codemirror/CodeMirror/issues/3122#issuecomment-77818075).

Most of the language features should be supported without any issues.

More info on JSX at:
  - https://facebook.github.io/jsx/
  - http://facebook.github.io/react/docs/jsx-in-depth.html

### DIFF
--- a/mode/jsx/jsx.js
+++ b/mode/jsx/jsx.js
@@ -1,0 +1,240 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"), require("../javascript/javascript"),
+        require("../xml/xml"));
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror", "../javascript/javascript", "../xml/xml"],
+           mod);
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
+  "use strict";
+
+  CodeMirror.defineMode("jsx", function(config, parserConfig) {
+    var tagOpen = /^\s*<([0-9a-z]+|(\/[0-9a-z]+))/i,
+        closingTagOpen = /^\s*<\//,
+        modes = {
+          'js': CodeMirror.getMode(config, {
+            name: "javascript"
+          }),
+          'xml': CodeMirror.getMode(config, {
+            name: "xml",
+            htmlMode: true,
+            multilineTagIndentFactor: parserConfig.multilineTagIndentFactor,
+            multilineTagIndentPastTag: parserConfig.multilineTagIndentPastTag,
+            allowMissing: true
+          })
+        };
+
+    function tokenBase(stream, state) {
+      if (state.nextMode) {
+        state.currentMode = state.nextMode;
+        state.nextMode = null;
+      }
+
+      // If current mode is not XML and an XML tag opening gets detected next,
+      // switch editor mode to XML.
+      // ATTENTION: This should not happen inside an attribute JavaScript
+      // expression, since XML structures are not valid XML attributes.
+      if (state.currentMode != "xml" && !state.context.inAttrJSExpression) {
+        var lastType = state.modeStates.js.lastType,
+            validLastTypes = [
+              'sof', 'keyword c', 'operator', '(', ':', ';', '}', ','
+            ],
+            xmlAllowed = (validLastTypes.indexOf(lastType) >= 0);
+
+        if (stream.match(tagOpen, false) && xmlAllowed) {
+          state.currentMode = "xml";
+        }
+      }
+
+      if (state.currentMode == "xml") {
+        // If the current XML tag got closed, switch to JavaScript mode if XML
+        // context is null, which means that there are no dangling open XML
+        // tags.
+        if (state.context.tagClosed) {
+          state.context.tagClosed = false; // Reset the switch anyway.
+          if (!state.modeStates.xml.context) {
+            state.nextMode = "js";
+            // Restart iteration without advancing any character, since XML
+            // expression re-evaluation has to happen, after a complete XML
+            // block closes.
+            return null;
+          }
+        }
+
+        // If a closing tag gets detected, switch to JavaScript mode after
+        // next ">" if XML context is null.
+        if (stream.match(closingTagOpen, false)) {
+          state.context.inClosingTag = true;
+        }
+
+        // If a ">" got detected while we are in a self-closing tag, if the
+        // XML mode context is null, switch to JavaScript mode in the next
+        // iteration.
+        if (state.context.inClosingTag && stream.peek() == ">") {
+          state.context.inClosingTag = false;
+          state.context.tagClosed = true;
+        }
+
+        // If a "/>" got detected this means that the tag got closed, mark this
+        // as a possible switch to JavaScript mode in the next iteration.
+        if (stream.match("/>", false)) {
+          state.context.tagClosed = true;
+        }
+
+        // Detecting a "{" in XML mode means that a JavaScript expression
+        // is about to follow. We should switch to back to JavaScript mode in
+        // next iteration.
+        if (stream.peek() == "{") {
+          state.context.inJSExpression = true;
+          state.context.jsExpressionOpenBraces = 0;
+          if (state.modeStates.xml.state.name == "attrValueState") {
+            state.context.inAttrJSExpression = true;
+          } else {
+            state.context.inAttrJSExpression = false;
+          }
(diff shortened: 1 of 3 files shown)


mode/jsx/jsx.js
```diff
@@ -0,0 +1,217 @@
...
+      if (state.currentMode != "xml" && !state.context.inAttrJSExpression) {
+        if (stream.match(tagOpen, false)) {
+          state.currentMode = "xml";
+        }
+      }
+
+      if (state.currentMode == "xml") {
+        // If a closing tag gets detected, switch to JavaScript mode after
+        // next ">".
+        if (stream.match(closingTagOpen, false)) {
```
marijnh: This seems to ignore nested tags within a single JSX expression.

mode/jsx/jsx.js
```diff
@@ -0,0 +1,217 @@
...
+        state.currentMode = state.nextMode;
+        state.nextMode = null;
+      }
+
+      // If current mode is not XML and an XML tag opening gets detected next,
+      // switch editor mode to XML.
+      // ATTENTION: This should not happen inside an attribute JavaScript
+      // expression, since XML structures are not valid XML attributes.
+      if (state.currentMode != "xml" && !state.context.inAttrJSExpression) {
+        if (stream.match(tagOpen, false)) {
```
marijnh: What about the less-than operator?

mode/jsx/jsx.js
parisk: Actually it's not. It is defaulting back to JavaScript and re-evaluates if it needs to switch to XML mode (https://github.com/codemirror/CodeMirror/pull/3742/files#diff-8fe55bb9634da263c3fa302cdbefb06cR42).

I could track the dangling open tags and wait for them to close before switching back to JavaScript, but I am not sure if it is a good practice. HTML tags are valid JSX syntax (http://facebook.github.io/react/docs/jsx-in-depth.html#html-tags-vs.-react-components) and from my understanding closing tags is optional in HTML5. I could implement that but sounds like it will introduce a lot of overhead, especially in the state object.

marijnh: The XML mode already tracks nesting (including understanding HTML close-tag-less tags, but I'm not sure JSX even allows that), and you can use that to determine when you're at the end of a tag.

mode/jsx/jsx.js
parisk: You 're right I got this initially wrong because I was trying to be more tolerant over when a tag starts (the XML mode is as well).

I took a look at the syntax of HTML tags (http://www.w3.org/TR/html-markup/syntax.html#syntax-start-tags) and added a stricter control. I am considering a valid "tag open" only when it matches the proper syntax of starting a "Start tag" (`<tagname`) or an "End tag" (`</tagname`) which allows no spaces in between `<`, `/` (if inside an "End tag") and the tag name, which should be a case insensitive alphanumeric.

I think I improved that a little bit (still has some edge cases) with https://github.com/codemirror/CodeMirror/pull/3742/files#diff-8fe55bb9634da263c3fa302cdbefb06cR17.

mode/jsx/jsx.js
parisk: It seems like the officially provided Babel syntax for Sublime does that as well (waits for tags to close before switching back to JavaScript syntax).

Can you give me a hint on how can I get that info from the XML mode state?

mode/jsx/jsx.js
marijnh: `if (bar<tagname)` is still not a JSX tag. You'll need the same check that the JS mode needs for determining whether a regexp can start at a given position, which is why I initially suggested a hook exposed by the JS mode.

mode/jsx/jsx.js
marijnh: Start each JS fragment with its own clean XML mode state, and exit it when after the end of a tag, that state's `context` property is `null` (meaning there are no remaining unclosed tags).

dabbott:
Hi guys. I had an unfortunate mid-air collision with this PR... but I'm putting my PR up as well. It's not 100% finished, but in terms of correctness I think it's very accurate. I originally attempted something like this approach, but found it drastically simpler and less error-prone to add a `jsx` flag to javascript mode and extend javascript mode directly. I think the closely-coupled-with-js approach will be easier to maintain... but would also understand if you guys don't want to go that route, since it's code that most JS users won't need.

I put more notes in the PR.

Have a look here: https://github.com/codemirror/CodeMirror/pull/3744

parisk:
@dabbott great job seems to have been done on your PR! I see that it doesn't apply XML highlighting though on the XML expressions. Do you think that this is something that we could work out? I could try out contributing that on your PR.

dabbott:
@parisk For sure! Actually, I haven't looked into how to do that stuff at all yet, so that would be great.

Also, bracket matching. And possibly tagname matching. I'm probably styling some stuff with the wrong styles too.

dabbott:
Can you clarify what you mean by XML highlighting?

mode/jsx/jsx.js
parisk: Thanks. Just updated my mode to exit XML mode only when context is `null` after matching a tag closing token.

parisk:
Hi @dabbott I took a look at your first screenshot that does not highlight the tag names, but then found out that you are highlighting them as JavaScript entities after taking a better look at your second one.

After I finish some final touches here, I could try giving some help with what's left.

mode/jsx/jsx.js
parisk: The edge cases should be even less right now. I made it stricter on when should it start considering a "tag open" according to the `lastType` detected by the JavaScript mode.

Reference: https://github.com/sourcelair/CodeMirror/commit/4f18f9c7474c490766bafa5e37d08bd85ba1d4e7#diff-8fe55bb9634da263c3fa302cdbefb06cR43

Also, it would be really helpful here if I could access also the following information:

-  the exact token recognized in the `lastType` (e.g. if the `lastType == "operator"`, knowing if it was `=` or `-` could help making highlighting even more semantic).
- if there are any new lines between the last token recognized and the current one.

dabbott:
Hmmm, yeah, I see what you mean. Logically I think they should be highlighted as JS variables/properties... but at the same time, they give the impression of not being highlighted, since it's relatively uncommon to have locally defined JSX elements or property access in element names.

petetnt:
In my opinion @parisk is correct: An JSX mode should

- Highlight all the known HTML elements
- Highlight all the custom HTML elements
- Highlight all declared JSX-elements

Consider something along these lines:

```
import FooComponent from "./foo";

const BarComponent => (props) => {
  return <div>Bar {props.bar}</div>;
}

class Biz extends React.Component {
  render() {
  return (
    <div classNames="foo">
      <h1>This is a foo component</h1>
      <x-my-component custom-attribute="foo" />
      <FooComponent>
        <BarComponent bar="is Bar" />
        <BazComponent /> 
      </FooComponent>
    </div>
   )
  }
}

```

In that example:

1. h1 and div should be highlighted as XML-tags as they are known HTML elements
2. `x-my-component` should be highlighted as it's a custom element (see [JSX gotchas @ facebook.github.io](https://facebook.github.io/react/docs/jsx-gotchas.html)
3. FooComponent and BarComponent should be highlighted as XML-tags as they:
  - Are wrapped in a `<`
  - Are declared in the file (one as a stateless component, one as imported component)
4. BazComponent should not be highlighted as it's not declared anywhere

marijnh:
HTML allows any tag names, not just the ones in the standard. As such, I don't think highlighting tag names differently is a very accurate or productive approach.

petetnt:
@marijnh Actually you are correct, I remembered that React would throw went it didn't understand the element but this doesn't seem to be the case, at least anymore.

marijnh:
Well, I suppose if three people are working on one, the need for this mode is pressing. And it looks like it's going to be less work to write one myself rather than to keep hinting and reviewing and pushing people in the right direction. You can find my version in b3f9487046e37facd64196380ebdd8639ef. Please test it and submit failing test cases. If this helps you in a financial way, a donation to cover the morning spent on this is much appreciated.

parisk:
:tada: really happy to see this implemented at all. I hope that our contributions helped you implement it faster and better.

dabbott:
Great! Glad to have this in. We sent you an email re: donation and buying a support plan.

leeoniya:
http://codemirror.net/mode/jsx/index.html

some self closed tags `/>` show up red (errors?)

PR #3743, opened 2015-12-28
Julia syntax clean-up and minor improvements

pabloferz:
I removed some unnecessary things inherited from the python mode. Also, as the syntax `{...}` is no longer used for creating arrays in Julia (currently, it is deprecated), I removed the corresponding scope handling.


marijnh:
Thanks! Merged as ef5dcef

PR #3744, opened 2015-12-28
Added jsx flag to javascript mode

dabbott:
I've added a JSX flag to javascript mode.

Originally I chose to try switching between XML and JS modes... but quickly realized I'd need a stack to keep track of the start and end of expressions... which is exactly what the JS mode already does! So rather than making a public API out of the JS mode, considering that JSX really *isn't* XML and really *is* just an extension to JS, I just added JSX directly into the javascript mode (akin to TypeScript... this is much simpler when closely coupled with JS, I think).

I notice a PR was opened for JSX yesterday (https://github.com/codemirror/CodeMirror/pull/3742)... but since this is already fully working (for both highlighting and indentation), I wanted to offer up this approach! (and I just finished this morning... it still would need tests and a little cleanup, but wanted to get it up ASAP)

Normal stuff:
screen shot 2015-12-28 at 12 02 02 pm

Edge cases:
screen shot 2015-12-28 at 11 53 43 am

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -19,6 +19,7 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
   var jsonldMode = parserConfig.jsonld;
   var jsonMode = parserConfig.json || jsonldMode;
   var isTS = parserConfig.typescript;
+  var isJsx = parserConfig.jsx;
   var wordRE = parserConfig.wordCharacters || /[\w$\xa1-\uffff]/;
 
   // Tokenizer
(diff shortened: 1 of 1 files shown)


petetnt:
This PR seems to support nested JSX/JSX as props  too right? Pretty impressive if so, especially as this is added to the JavaScript mode instead of new mode.

@lcampbell check this out

dabbott:
Yup :). By using the very powerful JS parser, most of the bookkeeping for the complicated stuff is free.

petetnt:
@dabbott Very cool, can't wait to get out of this bus I am in to test this out. We have been working on a JSX mode on and off for months but the nesting has usually been the Achilles knee in all of my implementations and @lkcampbell ran into it too just recently

petetnt:
Like discussed in #3742, I think this would need the XML-highlighting from #3742.

For this PR itself, template strings break the parser when used in props

```
class Foo extends React.Component {
  render() {
    const amount = 10;
    return (
       <Baz biz={`${amount} little piggies went to jail`} />
    )
  }
}
```

marijnh:
See my comment [here](https://github.com/codemirror/CodeMirror/pull/3742#issuecomment-167780690).

PR #3747, opened 2015-12-30
Fix ASN.1 and TTCN-CFG mode names in mode.js

oswcab:
These two modes were using unsupported characters in the mode name which
prevented integrating them into Gerrit. In fact, Gerrit uses the mode
name specified in this file as the name of a generated method. As Java
method names cannot include dots or dashes, this leads to a failed
compilation.


marijnh:
See my response to the identical issue you filed as #3354

Issue #3745, opened 2015-12-29
Failing cases in JSX mode


dabbott:
Following up on: https://github.com/codemirror/CodeMirror/pull/3742

Three from the jsx/index.html page:
1. Boolean attributes are not recognized properly (note the red self-closing `/>`)
   <img width="600" alt="screen shot 2015-12-29 at 9 41 02 am" src="https://cloud.githubusercontent.com/assets/1198882/12038976/6d3289a8-ae11-11e5-9c15-516930ddc12a.png">
2. Block comments styled incorrectly (should be the brown comment color)
3. Line comments are not recognized properly (note the red self-closing `/>`)
   <img width="599" alt="screen shot 2015-12-29 at 9 41 30 am" src="https://cloud.githubusercontent.com/assets/1198882/12038996/8eada342-ae11-11e5-9f8a-7de73b498e83.png">

Two other things I've found:

4 . Elements should be allowed as attributes
<img width="600" alt="screen shot 2015-12-29 at 9 42 54 am" src="https://cloud.githubusercontent.com/assets/1198882/12039046/f650f508-ae11-11e5-9bc9-f7bdebe97911.png">

5 . Indentation in JS expressions should be relative to the indentation level of JSX elements. I believe currently they get indented to prior JS indentation level. (this issue is highest priority, if I had to choose)

Current behavior gif here:
![js-expression-indentation](https://cloud.githubusercontent.com/assets/1198882/12039235/69cd6312-ae13-11e5-9aa8-9717118f3056.gif)

Correct position here:
<img width="213" alt="screen shot 2015-12-29 at 10 05 11 am" src="https://cloud.githubusercontent.com/assets/1198882/12039262/b9f03f9a-ae13-11e5-86fb-264053c93a83.png">

Another example of current behavior (not followed by } char):
![js-indentation-2](https://cloud.githubusercontent.com/assets/1198882/12039320/37f54d54-ae14-11e5-8e80-50b4ab480e42.gif)

Correct position here:
<img width="251" alt="screen shot 2015-12-29 at 10 08 29 am" src="https://cloud.githubusercontent.com/assets/1198882/12039316/3066bfe6-ae14-11e5-8587-36eb0f7ae92c.png">

@parisk @petetnt

marijnh:
I pushed some patches that take care of most of these (though getting indentation right across language boundaries is always tricky, and I'm not sure if the result is what you consider the right way to indent these). But...

>  Elements should be allowed as attributes

[...]

PR #3748, opened 2015-12-30
[javascript mode] Allow trailing comma in object destructure

dabbott:
Issue #3745


marijnh:
Thanks! Merged as 0d73c4b

PR #3749, opened 2016-01-01
Documentation HTML Cleanup

willdean:
We've been using CodeMirror 5.10 in a project which is being edited in Visual Studio.  For the sake of convenience, we just put the whole CodeMirror tree into the VS project.   

This has the side effect of various global error checking (mostly Resharper) running across the whole project, which then highlights a number of minor complaints about html tag errors in some of CodeMirror's documentation.

This PR contains fixes for most of them - I've wilfully breached the guidelines about one-fix-per-PR, and if you'd prefer it done like that instead, it's no problem.

Alternatively, if this PR doesn't suit the project's style in some way, I'm not bothered if it doesn't get merged at all, I just thought it might be useful.

### DIFF
--- a/demo/btree.html
+++ b/demo/btree.html
@@ -1,4 +1,4 @@
-<!doctype html>
+﻿<!doctype html>
 
 <title>CodeMirror: B-Tree visualization</title>
 <meta charset="utf-8"/>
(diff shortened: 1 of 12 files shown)


marijnh:
Thanks! On first glance, I thought this was going to be another 'I JSLint-ed your codebase for you!' pull request, but these are actually all useful and welcome cleanups. Merged (squashed) as f923adb

PR #3759, opened 2016-01-06
[jsx mode] Improve indention of js blocks

dabbott:
Issue #3745


marijnh:
Thanks. Looks good. Merged as 5b5976a6b and followed up with d40f5e270

PR #3761, opened 2016-01-08
Extended match-highlighter addon to display matches on scrollbar

hirak99:
Hi,

While editing code on Cider (a code editor used in Google, based on CodeMirror), I noticed all occurrences of the current word under cursor automatically highlighted.
However unlike how it works for a search, occurrences are not highlighted on scrollbar.

This patch request is to enable scrollbar highlights for word-under-cursor matches.

There are two proposed changes -

1. Addition of an option ``annotateScrollbar`` to the addon "match-highlighter.js". If this option is turned on, all the selection - or word under cursor - will also be highlighted in the vertical scroll bar (along with highlighting it in the document which match-highlighter already does).

2. Change in match-highlighter demo to have this option turned on by default, and added the required dependencies. Also expanded the sample text for match-highlighter demo to have a scrollbar to demonstrate this.

Regards

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -59,6 +59,7 @@ Anthony Gégo
 Anthony Grimes
 Anton Kovalyov
 AQNOUCH Mohammed
+Arnab Bose (hirak)
 areos
 as3boyan
 AtomicPages LLC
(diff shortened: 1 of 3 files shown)


hirak99:
Here's how it looks like.

Screenshot of demo

marijnh:
Looks good. Merged as 4a68108

PR #3763, opened 2016-01-12
Added HackMD in the real-world uses doc

jackycute:
Thank you CodeMirror, you made me this.
I would like to be on the list please!


marijnh:
Merged as df00bf3

PR #3771, opened 2016-01-16
bugfix: specialChar bug

ibone:
the editor value is '123456'
config specialChar = /123/
config specialCharPlaceholder = function(char){ return $("<span>7</span>")[0]; }
i think the result is '<span>7</span>456'
but the result is '<span>7</span>23456'


marijnh:
The regex provided to the `specialChars` option should match a _single_ character, not a string of them.

PR #3773, opened 2016-01-17
Update license year to 2016

pra85:


marijnh:
Thanks. Merged as 8850e26

PR #3774, opened 2016-01-17
Fix token type for beginning spaces inside string in Clojure mode

cldwalker:
Hi. 
  This is a bug one of our users [reported awhile back](https://github.com/LightTable/Clojure/pull/12) with the Clojure mode returning an incorrect token type. For a given string that starts with a space e.g. `" test"`, the token definition doesn't recognize the space as part of the string. This also means the token type for the space inside the string returns null when "string" is expected. I've demonstrated this with the following screenshots. Before the fix:

screen shot 2016-01-17 at 10 39 24 am

After the fix:

screen shot 2016-01-17 at 10 41 36 am

Let me know if there's anything else you'd like me to do. It would be great to get this fix in so LightTable can stop using a fork of the Clojure mode. Thanks for all your hard work!
Cheers, Gabriel

### DIFF
--- a/mode/clojure/clojure.js
+++ b/mode/clojure/clojure.js
@@ -142,7 +142,7 @@ CodeMirror.defineMode("clojure", function (options) {
             }
 
             // skip spaces
-            if (stream.eatSpace()) {
+            if (state.mode != "string" && stream.eatSpace()) {
                 return null;
             }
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as 17a0538

PR #3776, opened 2016-01-17
Allow namespaced forms to smartly indent in clojure mode

cldwalker:
Hi again. This is another fix a LightTable user [submitted awhile back](https://github.com/LightTable/Clojure/pull/19). As pointed in that issue, namespace qualified `with-*` calls aren't indented properly e.g.:
```clj
(namespace/with-open [a b]
                     (dothing a))
```

With this fix they are:
```clj
(namespace/with-open [a b]
  (dothing a))
```

Pulling block_indent into tests seemed reasonable so I left it as is. Let me know if there's anything you'd like tweaked here. Thanks!


marijnh:
Thanks. Merged as 75d597c

PR #3779, opened 2016-01-19
Fix anonymous fn with params indentation in Scala

vidma:
- [x] `Int` type was not defined in scala
- [x] fix indentation in scala-mode for anonymous functions with params
  * for now, loses coloring of params in multiline mode (but still much better than before IMHO), but do not affect one-line mode. this could be improved in separate PR, but need to think how to do it properly.

from:
screen shot 2016-01-19 at 13 42 41

to:
screen shot 2016-01-19 at 13 38 14

patch from https://github.com/andypetrella/spark-notebook/pull/551

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -497,7 +497,7 @@
     ),
     types: words(
       "AnyVal App Application Array BufferedIterator BigDecimal BigInt Char Console Either " +
-      "Enumeration Equiv Error Exception Fractional Function IndexedSeq Integral Iterable " +
+      "Enumeration Equiv Error Exception Fractional Function IndexedSeq Int Integral Iterable " +
       "Iterator List Map Numeric Nil NotNull Option Ordered Ordering PartialFunction PartialOrdering " +
       "Product Proxy Range Responder Seq Serializable Set Specializable Stream StringBuilder " +
       "StringContext Symbol Throwable Traversable TraversableOnce Tuple Unit Vector " +
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -11,7 +11,30 @@
 })(function(CodeMirror) {
 "use strict";
 
-CodeMirror.defineMode("clike", function(config, parserConfig) {
+  function Context(indented, column, type, align, prev) {
+    this.indented = indented;
+    this.column = column;
+    this.type = type;
+    this.align = align;
+    this.prev = prev;
+  }
+  function isStatement(type) {
+    return type == "statement" || type == "switchstatement" || type == "namespace";
+  }
+  function pushContext(state, col, type) {
+    var indent = state.indented;
+    if (state.context && isStatement(state.context.type) && !isStatement(type))
+      indent = state.context.indented;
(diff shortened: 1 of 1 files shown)


mode/clike/clike.js
```diff
@@ -519,6 +519,16 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
         stream.eatWhile(/[\w\$_]/);
         return "meta";
       },
+      "{": function(stream, state) {
+        // match a single full line with params of multi-line anonymous fn, e.g.:
+        // { a =>\n
+        // { (aa: Int, bb: Int) =>\n
+        var singleLineParams = /^(\w|[:.,() ])+=>$/;
+        if (!stream.match(singleLineParams, true)) // eat the line, if matches
```
marijnh: This'll disable all highlighting for the parameters. That seems undesirable. Also, I believe Scala allows default values for parameters, so the actual syntax for a parameter list may contain arbitrary subexpressions, which I assume your regexp won't cover. Would it make more sense to delay the context manipulation until reaching the `=>`?

mode/clike/clike.js
```diff
@@ -11,7 +11,30 @@
 })(function(CodeMirror) {
 "use strict";
 
-CodeMirror.defineMode("clike", function(config, parserConfig) {
+  function Context(indented, column, type, align, prev) {
```
marijnh: This file was retroactively wrapped in a module scope, hence the flat indentation. Please preserve it, so that indentation is at least coherent.

vidma: :ok:

mode/clike/clike.js
vidma: > This'll disable all highlighting for the parameters. That seems undesirable.

sure, I'd be more than happy to keep it. but it's much better as is, than the horrible identation we had before.

> Also, I believe Scala allows default values for parameters

not for anonymous fns, this is too weird / wouldn't compile:
```scala
scala> Option(1).map {x: Int = 1 => 1}
<console>:1: error: ';' expected but '=' found.
       Option(1).map {x: Int = 1 => 1}
                             ^
```

> Would it make more sense to delay the context manipulation until reaching the =>?

I had this in mind (but I'm not codemirror guru either). so what is needed to preserve the indentation, is to delay this context manipulation until whole line was `eaten` (into `highlighted parameters`), right?

marijnh: > but then, how to conveniently request tokening of the rest of the current line?

You don't need to, I think. You just change your hook to run on the `"="` character instead, and when it sees a `">"` ahead, it changes the current context to not try and align its body.

vidma: > when it sees a ">" ahead, it changes the current context to not try and align its body.

how would I change the `current context` ?


also, need to be rather cautious for this not to conflict with other syntax, e.g. if `{` was not present earlier on the same line before `=>`:
```scala
x match {
  case y =>
    khdjhsjdh
}
//or inside fn definition:
def callByReference(argFn: => Unit) = {
  ...
}
```

marijnh: Does patch dbb2a5c help?

Issue #3780, opened 2016-01-19
insert/overwrite cursor


vanderlee:
The cursor for insert and overwrite modes are identical.
Isn't it possible to have a more common new-character-width blinking box for overwrite mode?

vanderlee:
I've created a pull request for a simple implementation: https://github.com/codemirror/CodeMirror/pull/3782

PR #3782, opened 2016-01-19
Fix for #3780: Overwrite cursor

vanderlee:
This is a fix to have a blinking overwrite cursor using minor additions in `codemirror.js` and styling of `.CodeMirror-overwrite .CodeMirror-cursor` in `codemirror.css`.

When `state.overwrite` is true, the `CodeMirror-cursor`'s text is either the next character or, if the next character is empty, the `"\u00a0"` it has for insert mode. The css simply adjusts some things to put the inverted character in the cursor in the right place.

The result is a typical blinking overwrite cursor. It can be seen in action from the `This is CodeMirror` example in `index.html`, and presumably all other examples.

Works fine in Opera, Chrome and Edge, but there is a CSS issues in FireFox. I have not yet extensively tested all features and edge cases. I'd like to poll interest in this fix before continuing with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 *.swp
 .idea
 *.iml
+/nbproject
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -84,7 +84,14 @@
 }
 
 /* Can style cursor different in overwrite (non-insert) mode */
-.CodeMirror-overwrite .CodeMirror-cursor {}
+.CodeMirror-overwrite .CodeMirror-cursor {
+	width: auto;
+	background-color: black;
+	color: white;
+	border-left: 0;
+	line-height: normal;
+	white-space: pre;
+}
 
 .cm-tab { display: inline-block; text-decoration: inherit; }
 
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -2321,15 +2321,19 @@
   // Draws a cursor for the given range
   function drawSelectionCursor(cm, head, output) {
     var pos = cursorCoords(cm, head, "div", null, null, !cm.options.singleCursorHeightPerLine);
-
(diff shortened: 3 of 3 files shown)


marijnh:
Hi. Before I look at this, please configure your editor to not change every single newline in files that you touch, and then resubmit this as clean patches, so that I can actually see what changed.

vanderlee:
I'll run a conversion. What kind of line endings did you use?
Either way travis-ci is complaining about a license blob I didn't even touch.

vanderlee:
Done

marijnh:
Much better (though you still used tabs to indent the css, ignoring the surrounding conventions).

My take is that this is too specific and fragile (try with right-to-left text, for example) to put in the core library, but it would make a great addon. I believe the library exposes all the hooks necessary to update the cursor after a `selectionChange` event, except for an event that fires when overwrite mode is entered or left. Do you want to refactor it into an addon?

vanderlee:
Overwrite mode doesn't work at all in RTL mode, it just inserts (or atleast that's how it works in the bidi demo). This basically makes the RTL point moot at the moment. Regardless, Bidi support is quite doable; I've added basic support to show the character on the left when RTL.

No event for overwrite mode switching is a problem, Without a way to hook into that, it'll be practically impossible. Is there any way to get around this?

The difficult problem actually seems to be the CSS; `line-height` is not predictable.

Do you have any good documentation on how to make addons and/or perhaps a good example of a small one that I can use as a template?

Btw. is any of the Bidi information exposed to addons?

marijnh:
>
> No event for overwrite mode switching is a problem, Without a way to hook
> into that, it'll be practically impossible. Is there any way to get around
> this?
>
Yes, what I meant to say is that you're welcome to add such an event.

PR #3785, opened 2016-01-20
[manual] Minor fix to listed version number

vincentwoo:


marijnh:
The manual reflects the exact current version number... this is intentional.

vincentwoo:
Is this correct? Should it be 5.11.0, instead, then?

marijnh:
The manual on the website reflects the current dev version, which I give its own patch version bump to distinguish it from published versions. There's a 5.11.0 manual in the zip file for 5.11.0.

vincentwoo:
Ah, gotcha, thanks.

Issue #3781, opened 2016-01-19
CodeMirror-gutter Re-size


wissamsaikali:
i am using the merge plugin. ( https://codemirror.net/demo/merge.html )

Scenario : 
1 - Write a long commands on the same line
2- try to scroll right than left
3- gutters will move 

Please check the attached files 
i've tried to refresh the editor but it didn't worked .
how can i fix this ?

Thank you 
Regards
![screencapture-codemirror-net-demo-merge-html-1453211741453](https://cloud.githubusercontent.com/assets/7591496/12420660/9f2cddf8-bec6-11e5-9d5f-aa5c2b856eab.png)

marijnh:
Hi. This does not happen for me on Chrome. Which browser / version are you using?

wissamsaikali:
Hey, i am using Version 47.0.2526.111 (64-bit) on ubuntu

On Wed, Jan 20, 2016 at 11:54 AM, Marijn Haverbeke <notifications@github.com

marijnh:
Hm, that's exactly what I'm using. Could you describe your steps in more detail? When I go to the merge demo page, and scroll either of the editors right and then left, the gutter stays in place.

wissamsaikali:
1. Go the merge demo page
   2. Keep adding additional characters at the end of any line till the
   scrollbar on the left editor becomes clearly smaller than the one on the
   right
   3. Now when you scroll left and right on the left editor you will notice
   that the gutter will shift to the right

PS: this issue appears on Chrome Windows as well.

Thank you

On Wed, Jan 20, 2016 at 12:03 PM, Marijn Haverbeke <notifications@github.com

kepta:
+1, I was able to reproduce it on
Browser Firefox 
Mac

rtparsons:
Pull request to fix this #3788

marijnh:
The editor was clipping the horizontal scroll position wrong, leading to it believing it was setting its horizontal scroll somewhere the DOM wouldn't allow it to, confusing the gutter alignment code.

PR #3788, opened 2016-01-23
Fix for issue #3781 CodeMirror-gutter Re-size

rtparsons:
Gutter on non-editable window adding offset for changes made in
editable window.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -587,8 +587,11 @@
   function alignHorizontally(cm) {
     var display = cm.display, view = display.view;
     if (!display.alignWidgets && (!display.gutters.firstChild || !cm.options.fixedGutter)) return;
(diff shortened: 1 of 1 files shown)


lib/codemirror.js
```diff
@@ -587,8 +587,11 @@
   function alignHorizontally(cm) {
     var display = cm.display, view = display.view;
     if (!display.alignWidgets && (!display.gutters.firstChild || !cm.options.fixedGutter)) return;
-    var comp = compensateForHScroll(display) - display.scroller.scrollLeft + cm.doc.scrollLeft;
+
+    var editableOffset = cm.isReadOnly() ? display.scroller.scrollLeft : cm.doc.scrollLeft;
```
marijnh: The DOM for a read-only editor is (supposed to be) structurally the same as that for an editable one. What is the reasoning behind this conditional?

marijnh:
Superseded by attached patch.

PR #3790, opened 2016-01-27
FLC mode for CodeMirror

avatar29A:
FCL is Fuzzy control language in order to describe rules of  fuzzy logic.


marijnh:
Thanks. Merged as 6792348c4a

Issue #3794, opened 2016-01-29
In Python, multiple closing braces on a new line does not dedent multiple times.


ienzam:
Python indentation of closing braces does not trigger dedentation correctly.
Best explained with following example:
 
**Steps to reproduce:**
1. Go to any Codemirror python text area (http://codemirror.net/mode/python/index.html).
2. Write the following code.
```
# | shows the position of cursor
a = [{( | )}]
```
3. Press enter.

**What is observed:**
Code changes to:
```
# | shows the position of cursor
a = [{( 
        |)}]
```

**What is expected:**
Code may change to:
```
# | shows the position of cursor
a = [{( 
|)}]
```

marijnh:
Please take a look at attached patch, and see whether it solves your problems.

ienzam:
Thanks for a quick fix.

I tested the code. It improved the situation did not fix the problem completely.

Example:
1. Go to https://rawgit.com/codemirror/CodeMirror/master/mode/python/index.html
2. Write code:
```
a = [[
    [|]
]]
```
3a. Pressing enter turns the code to 
```
a = [[
    [
]
]]
```
3b. Expected
```
a = [[
    [
    ]
]]
```

PS: https://rawgit.com/ienzam/CodeMirror/issues/py_indentation/mode/python/index.html shows correct behaviour (at least for this case).

marijnh:
Good catch. Attached patch fixes this case.

ienzam:
Great, thanks for fixing these two issues quickly.

PR #3797, opened 2016-01-29
Improve dedentation of lines with multiple closing braces in Python.

ienzam:
Tries to address the issue #3794.

In Python, multiple closing braces on a new line does not dedent multiple times.

Can be tested the fix here - https://rawgit.com/ienzam/CodeMirror/issue/3794/mode/python/index.html


marijnh:
I've pushed an alternative patch, which makes the Python mode handle bracket scopes the way most modes do. It is probably more robust, and should also cover #3798

Issue #3795, opened 2016-01-29
In Python, add support to indent only once after multiple opening braces on previous line.


ienzam:
In Python, after a line with multiple (`n`) open braces, next line starts with `n * hangingSpace`.
But in most languages, indentation is only one unit of `hangingSpace`.

Here is coding example to make it clear what I mean.

**Steps to reproduce:**
1. Go to any CodeMirror python and clike text area (http://codemirror.net/mode/python/index.html).
2. Write the following code on both. (`|` shows the position of cursor)
```
a = [(|
```
3. Press enter.

**What is observed:**
Python cursor is like following:
```
a = [(
        |
```
Clike cursor is like following:
```
a = [(
  |
```

**What is expected:**
Both will have same clike cursor:
```
a = [(
  |
```

marijnh:
Again, I think this is addressed by 3df470adc8e7be

PR #3798, opened 2016-01-30
Python: Add support for indenting only once after a line with multiple open braces.

ienzam:
This tries to address issue #3795.


marijnh:
I think patch 3df470a has the effect you're looking for. Please test.

PR #3807, opened 2016-02-03
[active-line addon] Add active line number

sbruchmann:
This patch enables styling of active line numbers by
adding a `CodeMirror-activeline-gutter` class to the
line's gutter space.


marijnh:
Thank you! Merged as 18d3e0b

PR #3812, opened 2016-02-04
Adding jumpCharBase (defaults to 0)

mihailik:
An option `jumpCharBase` can be used to set base character offsets in Alt-G dialog to 1, so initial position would be 1:1 instead of current 1:0


marijnh:
This is not a good use of a top-level option. I'd prefer if you got the value from an optional second argument passed to `jumpToLine`.

PR #3813, opened 2016-02-05
spreadsheet mode no error on back slash

dckesler:
In spreadsheet mode with CodeMirror on a forward slash without a following character it throws stream errors. Also, if the input string ends with a forward slash it won't update the input box and throws a lot of errors.
This is a small fix that makes a single slash with no following character an atom type rather than null, and once a following character is added it becomes the string-2 type.


marijnh:
Thanks. Merged as e1a8c83

PR #3814, opened 2016-02-06
Add Vim.unmap to public api to undo Vim.map

cldwalker:
Hi. In the LightTable Vim plugin, we'd like our keybinding changes to stay in sync with the vim keymap bindings. Currently we can only map keybindings but can't undo them - https://github.com/LightTable/Vim/issues/66 . This seemed like a reasonable function to make public. Happy to discuss further if need be. Cheers

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -682,6 +682,9 @@
         // Add user defined key bindings.
         exCommandDispatcher.map(lhs, rhs, ctx);
       },
+      unmap: function(lhs, ctx) {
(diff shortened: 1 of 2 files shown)


cldwalker:
Anything I can do to expedite this? [unmap](http://vimdoc.sourceforge.net/htmldoc/map.html#:unmap) is a normal vim command so it seems reasonable to add this

marijnh:
I've merged this as 0b06a49. I'm not sure why `unmap` wasn't part of the API before, but it looks non-problematic.

Issue #3818, opened 2016-02-11
Vim mode: visual search/replace moves cursor to beginning of line rather than end


rforgione:
Visually selecting lines and then entering a command (for example, to find/replace) moves the cursor to the beginning of the command prompt rather than the end. This forces an extra keystroke to move the cursor to the end of the line. See below:

![gyf6xfdqq1](https://cloud.githubusercontent.com/assets/5572997/12969705/5cf28332-d052-11e5-9c17-e54355efce56.gif)

Using Firefox v44.0 -- first noticed the problem using Jupyter Notebook v4.1.0, but the problem is present even in CodeMirror's own demo of vim-mode, as seen [here](https://codemirror.net/demo/vim.html). To reproduce:
- choose a line in the code snippet
- press shift-v to highlight the line
- press : to enter a command

You'll notice that the cursor jumps to the beginning to the line rather than the end, which is not standard Vim behavior -- to perform any actions on these lines, the user then has to move the cursor to the end of the line. 

Thanks for taking a look!

marijnh:
The vim code is now being maintained at https://github.com/replit/codemirror-vim/ . If this is still important, check whether there is an issue open for it there.

PR #3824, opened 2016-02-14
Fix for #3818: Cursor Moves to Beginning of Dialog Lines

ggeise:
Firefox seems to handle the cursor placement differently when the input is focused.


marijnh:
Looks good. Merged as 9a80b69

PR #3826, opened 2016-02-15
adding Pharaoh to realworld.html

zacanger:
We couldn't've managed this without CodeMirror, and would be delighted if we could be included in your list!


marijnh:
Merged as 90d025e

PR #3835, opened 2016-02-19
Delete duplicate entry "MariaDB" from mode/meta.js

Dominator008:
It's already listed above as "MariaDB SQL".

Also add the ".mps" extension to the MUMPS mode.


marijnh:
Thanks! Merged as 68076a8

PR #3838, opened 2016-02-21
Add cljc, cljx, and edn to clojure.

kenrestivo:
As stated.


marijnh:
Thanks! Merged as caed330

PR #3844, opened 2016-02-22
Prepend the link to diff-match-patch.js in merge demo with "https:"

Dominator008:


marijnh:
Why?

Dominator008:
It defaults to "file://" when testing locally on Chrome.

marijnh:
Makes sense. Merged as 3e1e203

PR #3848, opened 2016-02-25
Fix lint tooltip to show multiple overlapping annotations in text editor

geowarin:
When hovering a line with multiple overlapping lint annotations, codeMirror will now show all the
 attached markers instead of just the first one.

I think this is what users expect to see since the gutter shows all the tooltips.
Previously, it depended on the order in which the annotations are sent, which is unpractical, especially in an async linter.


marijnh:
Looks good. Merged as 9e31c01

PR #3857, opened 2016-03-02
Add a mime type and separate modeInfo entry for edn

djwhitt:
This allows findModeByName to work for edn and seems more consistent with how JSON is handled.

The mime type isn't official on IANA yet, but there seems to be widespread consensus in the Clojure community on "application/edn".


marijnh:
Note that JSON actually needs to be parsed differently, which isn't the case here, right? But still, people might want to show the language name to the user somehow, in which case having a separate entry makes sense. Merged as 8799d1d4fb325

djwhitt:
Thanks!

In regard to parsing edn differently, afaik edn syntax is a strict subset of Clojure syntax, so the Clojure mode works as is. However, edn is stricter than Clojure in what it allows, so I could easily imagine adding an edn mode option to the Clojure parser at some point in the future.

By the way, the motivation behind this change (which I neglected to mention originally, but should have) was getting edn code block highlighting working in Markdown mode since it relies on findModeByName.

PR #3859, opened 2016-03-02
[twig mode] Add `base` option support for mixed mode

TDaglis:
Updated Twig mode to an multiplexingMode.
Twig mode now supports a base option and the base is properly highlighted.

##### html base
before:
before
after:
after

##### xml base
before: 
image
after:
beforexml

### DIFF
--- a/mode/twig/twig.js
+++ b/mode/twig/twig.js
@@ -133,7 +133,7 @@
     if (!parserConfig || !parserConfig.base) return twigInner;
     return CodeMirror.multiplexingMode(
       CodeMirror.getMode(config, parserConfig.base), {
-        open: "{", close: "}", mode: twigInner, parseDelimiters: true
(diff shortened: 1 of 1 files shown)


marijnh:
Would it ever make sense to use Twig with a non-HTML output language? If so, could you adjust your implementation to support a `base` option to configure the base language, the way the Handlebars mode does?

TDaglis:
Even though HTML would be the typical use case, it would make sense to use Twig with a non-HTML base. I updated the PR accordingly to support a `base` option as you suggested :+1:

marijnh:
(Is the fact that you're now defaulting to no base mode intentional?)

TDaglis:
Yes. For backwards compatibility mostly and in order to make the mode more generic.
I also just fixed the delimiters of the last commit since they did not work properly. I replaced them with regexes.

marijnh:
Thanks. Merged as a448b97

PR #3867, opened 2016-03-04
Fix indentation rules to match CommonMark (fixes #3833)

### LINKED ISSUE #3833: Markdown list highlighting only works on the top item in the second level list
Try to put below codes on Markdown mode or GFM mode:

```
- root1
- root2
    - sec1
    - sec2
        - third1
        - third2
```

The second level list only highlighting the top item (sec1) but not the others (sec2).
Is this a feature or bug maybe?
Thanks for helping!

bfrohs:
I'm pretty sure I got the behavior correct. Works for the provided test in #3833 as well as a the rather complicated one I came up with to test some of the edge cases (at least when comparing it to http://spec.commonmark.org/dingus/).

The basic idea is level is chosen based on the current item's marker (eg, `*`) indentation and the previous list item's *content* indentation levels. A second item in the same level will trash the previously stored indentation for that level's content. So, the stack will only be as deep as the list is.

```markdown
*    Level 1
    * Level 1 still
      * Level 2
     * Level 1 again
* Level 1
    * Level 2
  * Level 2 still
```

On a macro level, it can result in some confusion. But it makes sense on a micro level. Either way, a standard is a standard. Glad to have it.

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -149,10 +149,8 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
         state.list = null;
       } else if (state.indentation > 0) {
         state.list = null;
-        state.listDepth = Math.floor(state.indentation / 4);
       } else { // No longer a list
         state.list = false;
-        state.listDepth = 0;
       }
     }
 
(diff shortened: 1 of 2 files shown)


marijnh:
Many thanks! Merged as afed95a

PR #3869, opened 2016-03-07
Fix MIME type of troff

sinkuu:
`text/troff` is the standard approved by IANA. Also add some
non-standard MIME types.
https://bugs.freedesktop.org/show_bug.cgi?id=5072


marijnh:
Thanks. Merged as 12168b275d6e and followed up by 0514f28ea6776

PR #3870, opened 2016-03-07
Disable draggable in firefox desktop

SimonWoolf:
So: on Chrome, selecting text works fine. But on firefox, the `draggable="true"` attribute of `CodeMirror-scroll` (set by https://github.com/codemirror/CodeMirror/blob/24d60578042ae03ee471bf54f3e851de9684fe79/lib/codemirror.js#L179 for all but webkit and firefox mobile) prevents text from being selected by any click-and-drag operation that starts inside the draggable element, in some circumstances (which I haven't managed to pin down, but you can see an example where it's broken on any of the code snippets on https://www.ably.io/documentation/quick-start-guide). (*Edit:* that page now isn't an example as we did a hack to remove the `draggable` attribute after codemirror setup has run -- you can see the behaviour by re-adding `draggable="true"` to any of the `CodeMirror-scroll` divs)

This is probably [a firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=800050), but not one that looks likely to be fixed any time soon

This PR changes the `draggable` condition from `(!webkit && !(gecko && mobile))` to `(!webkit && !gecko)` (and puts it in a variable to stop the two uses getting out of sync).

I don't quite understand how the `dragDrop` functionality works or fits into this (since at the moment `draggable` is set whether or not `dragDrop` is true), so apologies if this is wrong.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -53,6 +53,8 @@
   // Optimize some code when these features are not used.
   var sawReadOnlySpans = false, sawCollapsedSpans = false;
 
+  var enableDraggable = (!webkit && !gecko);
+
   // EDITOR CONSTRUCTOR
 
   // A CodeMirror instance represents an editor. This is the object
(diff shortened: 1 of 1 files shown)


marijnh:
Have you been able to isolate the factor the causes this? The problem doesn't seem to happen on the demos, even when set the `readOnly` option to `"nocursor"`.

marijnh:
Not going to merge this without further information. Feel free to add comments.

PR #3873, opened 2016-03-07
addon/display/placeholder: handle swapDoc event

benmosher:
Without this, `swapDoc` from blank to non-blank (or vice-versa) results in a bad state of the placeholder.


marijnh:
Good call. Merged as d9b821a

Issue #3876, opened 2016-03-09
Velocity syntax highlighting is having hard time with variables in nested quotes


woshilapin:
First of all, great job with CodeMirror, this is really a nice piece of software.  I'll start with the formalities.

Issue observed on Linux Debian Jessie with the following webbrowsers:
* Firefox (v44.0.2)
* Chromium (v48.0.2564.116)
* Opera (v12.16)

Version of CodeMirror is the version deployed on codemirror.net and you should be able to reproduce the bug directly [here](https://codemirror.net/mode/velocity/index.html).

In this page, copy/paste the following line of Velocity code.
```
#set($something = "Hello '$me'!")
```

And you should see that all the syntax highlighting below this line is incorrect, like if the string was never closed.  Note that there is nested quotes in this line of code but that doesn't seems to be the problem.  Note also that there is a variable inside the quote but that doesn't seem to be the problem too.  For example, the 2 following lines are not causing problems.

```
#set($something = "Hello $me!")
#set($something = "Hello 'me'!")
```

It seems that the trailing single quote after the variable is causing the problem because the following line doesn't cause problem.

```
#set($something = "Hello '$me!")
```

marijnh:
@steveohara Would you be able to take a look at this?

steveohara:
Hi Marijn, I’ll take a look, might take me a couple of days to get to it

woshilapin:
Thanks, very fast :smile:

PR #3879, opened 2016-03-09
Velocity syntax highlighting is having hard time with variables in nested quotes #3876

steveohara:
Fixed: Use case added to HTML sample


marijnh:
Thanks! Merged as be7612d

PR #3882, opened 2016-03-11
Allow applications to listen for the dragleave event.

julienr:
This is a small change to allow applications to intercept the ``dragleave`` event in the same way as the other drag'n'drop events.


marijnh:
Good call. Merged as feed440

PR #3885, opened 2016-03-14
highlight the equal sign

kmsheng:
Hi, there !
Just realize that the equal sign should be able to highlighted in text/html mode, too.


marijnh:
The `equals` style isn't a standard token style, so it might be better to use `operator` for that. Also, you'll at least have to make the tests pass if you want me to merge something like this.

PR #3889, opened 2016-03-17
Avoid horizontal alignment of the gutters when the scrollbar is not…

pathob:
…moved

A reported bug in a project using codemirror for a diff view is the result
of the fact that currently the gutters may be horizontally aligned also if
the scrollbar of the affected codemirror-field is not moved (anymore).

This happend if a very long line was shortened a bit which results in a
diff view that has a left side that is a bit longer than the right
side. If now the left horizontal scroll bar is moved to the right, the
right horizontal scroll bar stops moving at the end of the line while
the gutter still moves a little bit (that looks like the left scroll bar
would push it a bit to the right).

With this pull request, it will be first verified that the scroll bar has
moved and only then the gutters will be horizontally aligned.


marijnh:
If I go to [the merge demo](http://codemirror.net/demo/merge.html) and shorten the longest line, I'm not seeing this happen. Could it be that 29555c986dd already addresses it?

pathob:
You're right... I didn't checked the newer versions precisely enough. Sorry and thank you.

PR #3897, opened 2016-03-22
jspm support

hmalphettes:
Hi!

This makes sure that jspm will not install the nodejs schims for buffer and process when installing codemirror for the browser.

I also had to override the `"directories"` to make sure that jspm was going to install codemirror in its entirety. With the original `"directories"` value, it would only install the content of the "`lib`" folder at the root of the package.

I am currently using my fork with:
```
jspm --version
0.16.31
jspm install codemirror=github:sutoiku/CodeMirror@5.13.0.n
```

I hope this help.
Many thanks for your attention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,5 +19,10 @@
                     "email": "marijnh@gmail.com",
                     "web": "http://marijnhaverbeke.nl"}],
     "repository": {"type": "git",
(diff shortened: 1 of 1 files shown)


marijnh:
Looks good. Merged as 3101742

PR #3904, opened 2016-03-23
Added yank command to yank the current line

t20:


marijnh:
Looks good! Merged as 65964cf

PR #3907, opened 2016-03-25
Check if cm.save() is available before saving in keymap/vim.js

Dominator008:
cm.save() might not exist if CodeMirror is not created from a
textarea. This adds an additional check to prevent TypeErrors.


marijnh:
Thanks. Merged as 1cc93d9

PR #3908, opened 2016-03-25
Fix the link to Gerrit Code Review in doc/realworld.html

Dominator008:
The code.google.com link is deprecated. Also add that CodeMirror
is being used as Gerrit's inline editor.


marijnh:
Thank you. Merged as fa705ea

PR #3911, opened 2016-03-27
PowerShell mode for CodeMirror

ashmind:
This is a PowerShell mode for CodeMirror.
I've done it a while ago, so I don't remember all details, but the following should work well:

1. Various literals
2. Variables, including `${}`
3. Various operators
4. String interpolation, including nested string interpolation
5. Builtins (known variables and functions)

And I've just added some tests (around 94% code coverage).

Let me know if there are any issues.

### DIFF
--- a/mode/powershell/index.html
+++ b/mode/powershell/index.html
@@ -3,15 +3,30 @@
   <head>
     <meta charset="utf-8">
     <title>CodeMirror: Powershell mode</title>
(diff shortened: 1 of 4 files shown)


marijnh:
Thanks! Mode tests are always great to have. Merged as fa705ea05a3...95f6840611f and integrated in c9b28f6268a596.

PR #3913, opened 2016-03-28
[vim bindings] Remove dead code and initialize a property

Dominator008:
The Error after the while(true) loop is not reachable and should
be removed. The "this.iterator" property should be given an
initial value of 0. This fixes two warnings reported by Closure
Compiler.


marijnh:
Thanks, looks good. Merged as c3d2246

PR #3914, opened 2016-03-28
[tiddlywiki mode] Fix bogus return statements

Dominator008:
8ba727a0310b2e87d6a2c310580d368b1e3cc791 caused two return
statements to return the wrong variable. In addition, they cause
warnings when compiled with Closure Compiler. Fixed them to
return what they should.

### DIFF
--- a/mode/tiddlywiki/tiddlywiki.js
+++ b/mode/tiddlywiki/tiddlywiki.js
@@ -324,12 +324,7 @@ CodeMirror.defineMode("tiddlywiki", function () {
     word = stream.current();
     known = keywords.propertyIsEnumerable(word) && keywords[word];
 
-    if (known) {
-      return known.style, word;
-    }
-    else {
-      return null, word;
-    }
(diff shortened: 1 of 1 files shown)


marijnh:
`return a, b` will return `b` (the result of the comma op is the last operand), so I've pushed alternative patch 420bdf1 instead.

Dominator008:
@marijnh I don't think the alternative patch is correct. See https://github.com/codemirror/CodeMirror/commit/8ba727a0310b2e87d6a2c310580d368b1e3cc791#diff-502b01b98b1553a622969a4b62a90a3eR328

If you check the original `ret()` method, you can see that these lines were supposed to return `known.style` and `null` respectively, not word. I guess you overlooked this when you did the refactoring :).

marijnh:
Thanks for tracking that down. The original `ret` was a completely unused bogus abstraction, and I guess I got a bit rough with the code when removing them. See ea6382a for a futher cleanup.

PR #3915, opened 2016-03-28
[dtd mode] Fix code that lacks side-effects

Dominator008:
The 6 "n"s are reported as suspicious code by Closure Compiler.
They can just be empty statements instead.


marijnh:
Thanks. Merged as ab455f2

PR #3916, opened 2016-03-28
[forth mode] Use string concatenation instead of multiline string

Dominator008:
This fixes a warning generated by Closure Compiler. Multiline
string literals are not part of ES3 / ES5 and is better replaced
by string concatenation. See
https://google.github.io/styleguide/javascriptguide.xml#Multiline_string_literals.


marijnh:
Multi-line strings are most certainly part of ES3 and 5 (see [here](http://www.ecma-international.org/ecma-262/5.1/#sec-5.1.3) for example), the ES3 grammar contains a similar clause. I know some tools are fussy about them, but I consider that a problem with the tools, and have no interest in purging them from CodeMirror.

Dominator008:
Ok fine, I think the section should be http://www.ecma-international.org/ecma-262/5.1/#sec-7.8.4.
Seems like the style guide needs to be corrected.

PR #3917, opened 2016-03-28
[properties mode] Use empty block instead of empty statement

Dominator008:
This eliminates a warning from Closure Compiler.


marijnh:
Agreed. Merged as aa0b249

PR #3919, opened 2016-03-28
[tiddlywiki mode] Fix a return statement

Dominator008:
According to
https://github.com/codemirror/CodeMirror/commit/8ba727a0310b2e87d6a2c310580d368b1e3cc791#diff-502b01b98b1553a622969a4b62a90a3eR328,
the return value should be known.style or null, not word.


marijnh:
Yeah, the code made no sense whatsoever to begin with, and my fix caused a linter error, so I followed up with ea6382a, which tries to fix the most obvious brokenness in that file. I think it also addresses this.

Dominator008:
This cleanup looks much better. Thanks!

PR #3921, opened 2016-03-29
Upon replace, search from the start of selection

ttyrueiwoqp:
Upon replace, if there is a match in the selection, it would be better to replace the current match, instead of the next match after the selection.


marijnh:
Looks good. Merged as 8f095df

PR #3924, opened 2016-03-29
Link to a third-praty minified CodeMirror distribution

Dominator008:
This is a minified CodeMirror distribution. All JavaScript files have
been minified with Google Closure Compiler using SIMPLE_OPTIMIZATIONS,
and all CSS files have been minified with clean-css. This projects has
CodeMirror's main repository as a Git submodule and is released once per
CodeMirror release using the same version number.

This distribution contains only the addon, keymap, lib, mode and theme
directories. Within them, only non-test JavaScript and CSS files are
included. See the file section in package.json for the list of things
included.

### DIFF
--- a/doc/docs.css
+++ b/doc/docs.css
@@ -159,7 +159,7 @@ section.first {
 
 .actions {
   margin: 1em 0 0;
-  min-height: 100px;
+  min-height: 130px;
   position: relative;
 }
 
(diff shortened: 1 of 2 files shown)


marijnh:
That's nice, but not something that warrants a prominent link on the project frontpage. Add it to [the wiki](https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons) if you want.

PR #3926, opened 2016-03-31
Add async/await keywords

developit:
I'm not quite familiar with CodeMirror's codebase, but this seems to work when I test it out locally.

**Screenshot:**


developit:
Actually it looks like someone did a far nicer job in PR #3479. Curiously, it's not published to NPM. @marijnh - is that still "awaiting" release? (sorry for bad pun)

marijnh:
The JavaScript mode does a rather detailed parse, and if that gets confused, it'll start getting context-dependent things like recognizing variable definitions, regexp-vs-division, and template strings wrong.

Since these do not syntactically behave like atoms (which are simply expressions), I don't think this is a good idea. With `await`, you could simply make it a “keyword c”, since it syntactically behaves pretty much like `delete` or `yield`. With `async`, you'll have to create a new keyword type, and handle is specially in the `statement` function (just skip it and start parsing a new statement, something like `return cont(statement)`).

PR #3927, opened 2016-03-31
Initial Push of SAS mode

jld23:
This is the initial push of a SAS mode for Codemirror. It will allow SAS files to correctly highlighted.
Here is the type definitions used:
//Definitions
//  comment -- text withing * ; or /* */
//  keyword -- SAS language variable
//  variable -- macro variables starts with '&' or variable formats
//  variable-2 -- DATA Step, proc, or macro names
//  string -- text within ' ' or " "
//  operator -- numeric operator + / - * ** le eq ge ... and so on
//  builtin -- proc %macro data run mend
//  atom
//  def


Dominator008:
Can you squash the commits into one? Thanks!

jld23:
I've squashed all the commits since my first one. If you need anything else please let me know.

Jared

jld23:
I'll open a new pull request that is cleaner instead of pulling in all those changes.

PR #3932, opened 2016-04-01
initial push of SAS mode

jld23:
Here is a clean pull request.

Here are the definitions of the style types 

//Definitions
//  comment -- text withing * ; or /* */
//  keyword -- SAS language variable
//  variable -- macro variables starts with '&' or variable formats
//  variable-2 -- DATA Step, proc, or macro names
//  string -- text within ' ' or " "
//  operator -- numeric operator + / - * ** le eq ge ... and so on
//  builtin -- proc %macro data run mend
//  atom
//  def

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -238,6 +238,7 @@ Jan Keromnes
 Jan Odvarko
 Jan Schär
 Jan T. Sott
+Jared Dean
 Jared Forsyth
 Jason
 Jason Barnabe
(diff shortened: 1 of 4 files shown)


marijnh:
Great. Could you add an `index.html` that shows an example of the mode? (You don't need to open a new pull request -- if you push something new to the branch that the PR is based on, it'll update the pull request, even if you force-push to overwrite previous commits. Be sure to leave a comment when you update, so that I notice that something happened.)

jld23:
I've pushed an index.html file
Thanks!
Jared

marijnh:
Merged as 7fa0cdc493e4, made to conform to our code style in 9389a050790

By the way, you don't have to add yourself to AUTHORS, that's done automatically on every release.

xiangzhimeng:
Hi there seems an issue with the sad mode. When I type empty string or string only with space in a quote, the sas mode will treat everything after the quote as sting too. You can try this with mode demo on the codemirror website.

marijnh:
Please don't report the same issue twice. This was handled in #4178

xiangzhimeng:
Sure, sorry I won't post in two places anymore.

PR #3933, opened 2016-04-03
[webidl mode] Add

TimothyGu:


mode/webidl/webidl.js
```diff
@@ -0,0 +1,189 @@
...
+function readToken(stream, state) {
+  // whitespace
+  if (stream.eatSpace()) return null;
+
+  // comment
+  if (stream.match("//")) {
+    stream.skipToEnd();
+    return "comment";
+  }
+  if (stream.match(multilineComments)) return "comment";
```
marijnh: A stream only encapsulates a single line, so you can't have tokens that span lines. Instead, you have to use your state to remember the fact that you're in a multi-line comment, and return a `"comment"` token for each line that's part of a the comment.

mode/webidl/webidl.js
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  // string
+  if (stream.match(strings)) return "string";
+
+  // identifier
+  var pos = stream.pos;
+  if (stream.match(identifiers)) {
+    if (state.startDef) return "def";
+    if (state.endDef && skipSpaceAndMatch(stream, /./)[0] == ";") {
```
marijnh: Doesn't `stream.match(/\s*;/, false)` make more sense here? It's simpler, and wouldn't require mucking with stream internals.

mode/webidl/webidl.js
```diff
@@ -0,0 +1,189 @@
...
+      return "def";
+    }
+    stream.pos = pos;
+  }
+
+  if (stream.match(keywords)) return "keyword";
+
+  if (stream.match(types)) {
+    var lastToken = state.lastToken;
+    var nextToken = skipSpaceAndMatch(stream, /.+?\b/)[0];
```
marijnh: What is `\b`?

mode/webidl/webidl.js
TimothyGu: Hmm you are right. However, is there a way to look ahead to the next line? Something that would make `endDef` recognize

```c
typedef a b
;
```

?

mode/webidl/webidl.js
TimothyGu: A word boundary: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#special-word-boundary or http://www.ecma-international.org/ecma-262/6.0/#sec-assertion

I'm just trying to fetch the next word/token.

mode/webidl/webidl.js
marijnh: There isn't (your code also won't be able to) -- if you need logic that goes across lines, you have to track state and continue reasoning in a later call to `token`.

mode/webidl/webidl.js
marijnh: Ah, sorry, not sure why I didn't recognize that. So again, it's easier here to just call `match` with a single regexp.

TimothyGu:
Fixed all comments.

marijnh:
Thanks! Looks good now. Merged as adb73ff

PR #3935, opened 2016-04-04
Fix mathematica mode by advancing the stream on error token

jvdl:
The mathematica mode breaks when it gets to the "error" case. In my case I saw it run in to an unrecognised character which brings it to returning "error" in its tokenizer, however the stream was not advanced for error tokens, and hence CM throws a "Mode _x_ failed to advance stream".


marijnh:
Thanks! Merged as 6fa5752

PR #3937, opened 2016-04-04
Add text/x-gql MIME type

garysheng:
GQL is Google's SQL-based Query Language.
See here: https://cloud.google.com/datastore/docs/apis/gql/gql_reference


marijnh:
Thanks! Merged as eb639c1

PR #3938, opened 2016-04-05
cleanup methods to help prevent DOM (detached node) leaks

scottschiller:
Hi CodeMirror folk,

This is my first fork and PR of a public repo in an ice age, so please forgive any obvious oversights. :laughing:

We use CodeMirror in a number of places within the Slack web app, which runs in desktop browsers and our own installable desktop apps which use web views (Safari/Webkit WebView) and Chromium (Electron) at present.

Given the dynamic nature of our web app, we regularly create and destroy views that can include CodeMirror instances. I noticed a number of detached DOM nodes building up very quickly over time, and found that CodeMirror just needs to release some DOM references when the view is being torn down to prevent the DOM tree(s) it creates from being leaked.

In our case, we'll instantiate with `editor = CodeMirror.fromTextArea(...)`, and later, call `editor.cleanup()` when the view is being destroyed. This calls a `cm.cleanup()` method specific to the `fromTextArea` instance, and finally `CodeMirror.cleanup()` which handles the rest.

Before my patch, running the contents of codemirror.js itself through the above and subsequently clearing the DOM underneath (i.e., canceling our "edit snippet" dialog) would result in 185,961 entries (yow, that's a lot of formatting nodes!) being shown under a Detached DOM tree item in Chrome Dev Tools. Over time, this would contribute to the webapp becoming slower as the detached DOM grew in size.

codemirror dom leaks 2

Whilst I'm pretty sure I'm preaching to the choir here, detached nodes can be found in Chrome Dev Tools under Profiles -> Heap Snapshots (then take a snapshot), and then changing the view to from Summary to Containment View, or, just looking for Detached DOM tree entries in the list.

Every so often I look over the Slack webapp for leaked DOM trees and such, and identified this pattern with CodeMirror. I made edits as best I could to isolate and prevent against dangling event, property and direct node references within the new `cleanup()` methods.

Per guidelines, before making this PR I ran tests locally, and did not see any new regressions or failures introduced with my change. `bin/lint` also did not throw any complaints.

We've been applying my patch for the last while as we update CodeMirror, and I realize that's rather silly to keep doing. Please modify this change as you see fit and if it's useful, I'd be happy to see it worked in upstream. Thanks! :bow:

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -5953,9 +5953,21 @@
     };
 
     textarea.style.display = "none";
+
     var cm = CodeMirror(function(node) {
       textarea.parentNode.insertBefore(node, textarea.nextSibling);
     }, options);
+
+    cm.cleanup = function() {
+      if (textarea.form) {
+        off(textarea.form, "submit", save);
+      }
+      cm.toTextArea();
+      textarea = null;
+      cm = null;
+      CodeMirror.cleanup.apply(this);
+    }
+
     return cm;
   };
 
@@ -8887,6 +8899,32 @@
       return order;
     };
   })();
+
+  CodeMirror.cleanup = function() {
+    // Preventing detached DOM leaks: loop over and null out view and display objects.
+    var item, i, j;
+    var d = this.display;
+    var view = this.display.view;
+
+    function nullObject(obj) {
+      if (!obj) return;
+      for (item in obj) {
+        if (obj.hasOwnProperty(item)) {
+          obj[item] = null;
+        }
+      }
+    }
+
+    for (i=0, j=view.length; i<j; i++) {
+      nullObject(view[i]);
+    }
+
+    measureText = null;
+
+    nullObject(d);
+
+    d = view = null;
+  }
 
   // THE END
 


vincentwoo:
This is a very cool writeup.

marijnh:
Hello Scott,

Thanks for looking into this. The need to call `toTextArea` on a textarea-based CodeMirror in order to clean it up was not mentioned in the manual, which was definitely an oversight. I've added patch 66fd40f to fix this.

As for actually leaking those DOM nodes, the model CodeMirror uses is that to destroy it, you must remove it from the DOM (which you did), _and_ drop the reference to the editor, so that the GC can collect it. Your measurements sound like you didn't do the latter. I set up some tests where I create a boatload of CodeMirrors, but don't keep references to them, and then remove them from the DOM. The result is a single detached DOM node (`measureText`).

I'd prefer to avoid adding a separate `cleanup` method. Would it be difficult to restructure your code to null references to CodeMirror that are no longer alive?

scottschiller:
Hi @marijnh,

Thanks; from what I'm seeing, we are nulling the sole CodeMirror reference ourselves when canceling our snippet creation / editing dialog - but CodeMirror is still hanging onto part of the DOM, possibly the `form` reference which you mentioned. It's hard to tell in these profile snapshots, exactly what the top-level reference is that is responsible for leaking the whole tree.

codemirror dom leak - form

From what I can see, we are destroying the form element being leaked from within Slack when the user cancels the snippet dialog, for example, followed by nulling the reference to the `codeMirror.fromTextArea()` instance - but because CodeMirror has a reference to the form, it doesn't get cleaned up.

I'd prefer not to have to call a method for clean-up either, but found that pattern did work for our use case. I probably went overboard in nulling out internal references, but it would be lovely to be able to have CodeMirror be able to clean up any dangling references when we "cancel" it.

Maybe a cancel-style DOM event (if not a new method altogether) would be the sort of thing that could be useful here, in terms of folks being able to signal that they're finished with the CM instance for whatever reason.

Let me know what you think!

marijnh:
> but it would be lovely to be able to have CodeMirror be able to clean up any dangling references when we "cancel" it.

Thing is, it's supposed to already do that, out of the box. If you can reduce your test case to a minimal, no-dependencies form, I'll gladly see if I can debug it.

scottschiller:
@marijnh: Good idea; I'll see if I can get a minimal repro. Thanks!

marijnh:
Closing until further information comes in.

PR #3939, opened 2016-04-06
miroire

akro3x:
<link rel="stylesheet" href="lib/codemirror.css">
<script src="lib/codemirror.js"></script>
<script>
  var editor = CodeMirror.fromTextArea(myTextarea, {
    lineNumbers: true
  });
</script>


Dominator008:
This does not look like a patch.

marijnh:
Agreed. Not sure what this is trying to do, but it looks like an accident of playing around with the github API. Closing.

PR #3942, opened 2016-04-06
Tweaks to lint gutter and icecoder theme

mattpass:
Slightly slimmer gutter, with a 2px margin to left, to give a little space when you have maybe a gutter to left. cursor changed to help icon which seems to make more sense.

Warmer background color added for icecoder theme, plus fixed matching bracket (was a bordered box, which caused jumping and char became invisible (black on almost black), now grey box with white char).

### DIFF
--- a/addon/lint/lint.css
+++ b/addon/lint/lint.css
@@ -1,6 +1,6 @@
 /* The lint marker gutter */
 .CodeMirror-lint-markers {
-  width: 16px;
+  width: 12px; margin-left: 2px;
 }
 
 .CodeMirror-lint-tooltip {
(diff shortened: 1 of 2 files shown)


marijnh:
I've merged the theme changes as d6e8f34 (though to me the new brackground looks colder, rather than warmer).

The gutter change for the lint addon isn't a good idea -- it causes the icons to stick out of the gutter, which will look terrible when the next gutter has no padding on the left.

mattpass:
OK I see your point on the gutter. There was also changing the cursor from pointer to help also which seemed to make sense but no biggie.

marijnh:
I'm not generally a fan of using elaborate pointers. On my (Linux) setup the help pointer is ugly as hell. Since the tooltips come up instantly anyway, there doesn't seem to be much need for an informative pointer.

PR #3945, opened 2016-04-07
Add __key__ keyword to GQL MIME definition

garysheng:


marijnh:
Thank you! Merged as 44e4bd4

PR #3946, opened 2016-04-07
Recognize Bazel, Buck and Pants build files as Python in mode/meta.js

Dominator008:
BUCK, BUILD, *.BUILD and *.bzl files are build files for the three build
systems mentioned, and they are all valid Python files.


marijnh:
Thanks! Merged as bef7e37

PR #3948, opened 2016-04-09
[git commit message mode] Add and integrate

Dominator008:
Add a simple mode to color the Parent, Author, AuthorDate, Commit,
CommitDate lines as keywords and any footers like Change-Id or
Signed-off-by as builtin style. This gives the screen a bit more color
than the flat black text.

Originally written by Shawn Pearce <sop@google.com>, tweaked a little
bit by Michael Zhou <zhoumotongxue008@gmail.com> to match the style of
other current modes. Also fixed matching of headers / footers when they
are indented.

### DIFF
--- a/mode/git-commit-message/git-commit-message.js
+++ b/mode/git-commit-message/git-commit-message.js
@@ -1,0 +1,44 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+// Originally written by Shawn Pearce <sop@google.com>,
+// tweaked by Michael Zhou <zhoumotongxue008@gmail.com>
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"));
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror"], mod);
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
+"use strict";
+
+CodeMirror.defineMode("git-commit-message", function() {
+  var header = /^(Parent|Author|AuthorDate|Commit|CommitDate):/;
+  var id = /^Change-Id: I[0-9a-f]{40}/; // Gerrit specific
+  var footer = /^[A-Z][A-Za-z0-9-]+:/;
+  var sha1 = /\b[0-9a-f]{6,40}/;
+
+  return {
+    token: function(stream) {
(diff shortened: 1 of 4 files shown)


marijnh:
Could you explain this header format ("Author: ...", etc)? Isn't the author/date/commit style metadata tracked outside of the commit message? Why would you want to put it in there?

Dominator008:
@marijnh Our project (Gerrit) displays the metadata, eg.: https://gerrit-review.googlesource.com/#/c/75967/9//COMMIT_MSG

marijnh:
This is very Gerrit-specific then. I think it'd make more sense to keep it part of the Gerrit project.

TimothyGu:
> Could you explain this header format ("Author: ...", etc)? Isn't the author/date/commit style metadata tracked outside of the commit message? Why would you want to put it in there?

They are technically outside of the message proper, but I think `git show --format=fuller` is a case legitimate enough to warrant a place in the general Git mode.

marijnh:
Sure, but hardly anyone types commit messages in in that format. I'm going to say no, I don't want to include this in my distribution.

Dominator008:
@marijnh Well technically this is similar to the newly added [mbox](https://codemirror.net/mode/mbox/index.html) mode. Although rarely edited, the metadata is useful for viewing.

marijnh:
The mbox mode was a close call as well. In general, the way we're currently distributing modes is starting to hit its limits -- all the modes blow up the main package's size, and I am less and less interested in maintaining all the code for them (some of it downright awful). I'm looking at ways to push modes into separate packages, with separate maintainers, without making the burden for the user too large. Maybe NPM is widely used enough nowadays to just depend on that.

Dominator008:
@marijnh That sounds good to me. The migration could be part of Version 6.

marijnh:
Agreed. 6 will happen somewhere this summer or early autumn.

marijnh:
(But note that it's already trivial to package up modes as NPM packages. Just no one seems to be doing it.)

Dominator008:
We can start with some documentation to encourage people to do so.

TimothyGu:
So I assume that there will be a set of "official" modes/addons/keymaps that are recommended by CodeMirror?

marijnh:
The minification helper is going to go when I release 6.0. I feel almost everyone that cares about performance has their own build step nowadays, so I'm going to leave minification to that.

There'll probably be a naming convention for mode packages, so that you can search NPM for, say, "codemirror-mode-" to easily find such packages. I'll continue maintaining the modes that I wrote or consider very important -- but as packages.

PR #3949, opened 2016-04-10
Use goLineStartSmart for cmd-left in Sublime keybindings

vincentwoo:
OSX users of Sublime will be used to this as the default behavior. Right now CodeMirror with Sublime bindings jumps all the way to pos 0 on cmd-left. In actual Sublime, it jumps to start of text before going to pos 0. Luckily there's already a handy command for that!


marijnh:
Is there an equivalent on non-Mac platforms? If so, could you add it, and if not, could you make this line conditional with `if (mac) ...`?

vincentwoo:
There is not. I updated the PR.

marijnh:
Thanks. Merged as 90ef4bd

PR #3952, opened 2016-04-11
Fix fenced code block mode switching for objective-c code blocks in github flavored markdown.

james-howard:
There were two issues:

1. Github flavored markdown calls Objective C `objective-c`
2. The gfm.js mode didn't include `-` in its fenced code block regex

Additionally, test/index.html did not import the mode/meta.js file, which prevented mode lookup by name from working in unit tests.

### DIFF
--- a/mode/gfm/test.js
+++ b/mode/gfm/test.js
@@ -51,6 +51,12 @@
      "[comment ```]",
      "bar");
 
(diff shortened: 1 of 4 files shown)


marijnh:
Thanks for taking care of that. Merged as 44410cd

PR #3959, opened 2016-04-15
Updated with a suggestion link to information providing site

manideepami:
adding a resource wher we can get intuitive information on professional certifications. We have successfully completed the project and its live!! Site will be updated regularly


marijnh:
Nice. How does it use CodeMirror, though?

manideepami:
using codemirror editor, we have made all the text edits from the backend. The very best text editor i've found so far.

marijnh:
Merged as a7868e1

PR #3960, opened 2016-04-15
Return null checks for annotation

SerCeMan:
Fix "Uncaught TypeError: Cannot read property 'severity' of undefined."

Previous changes made by this [commit](https://github.com/codemirror/CodeMirror/commit/9e31c01e11af098b76e059cf85ea5c9f9527fc99) broke my code. 
Here is the fix.

BTW, thank you for  the awesome tool!


marijnh:
Thanks. Merged as bf8e72f

PR #3964, opened 2016-04-16
[webidl mode] Fix stream handling

TimothyGu:
Unfortunately def70631ca7a50cfb5bf414c27df4db53f899a15 broke the mode. There are two easy ways to fix this:

1. revert that commit, and continue to use stream internals
2. merge this PR, which is cumbersome

@marijnh, what would you suggest?


marijnh:
Wouldn't it be much simpler to match different regular expressions (possibly generated from the same data) in the two different cases (`state.startDef` and `state.endDef`)?

TimothyGu:
Derp… fixed.

marijnh:
Thanks. Merged as 479cc65

PR #3966, opened 2016-04-17
[yacas mode] fix multiline comments

grzegorzmazur:
Obvious fix for a silly error in yacas mode causing multiline comments to be at most one line long


marijnh:
Thanks. Merged as 060445d

PR #3968, opened 2016-04-17
Do you have any color panel?

merbin2012:
I want to show the color picker, when clicking on the color related code. I found a plugin for codemirror "https://github.com/enjalot/Inlet" but it is not working on mobile device.


marijnh:
Nope.

PR #3969, opened 2016-04-18
[yacas mode] Implement indentation

grzegorzmazur:


marijnh:
Thanks! Merged as aa98087

PR #3974, opened 2016-04-19
Allow completions hinting functions to return Promises.

### LINKED ISSUE #3971: showHint: allow hint function to return promise ?
Does it make sens ?
Would a patch doing that be accepted ?
if so, any requirements beyond the usual ?

Carreau:
Even if we don't use promise internally, check for the existence of
`.then` attributes on the return results of a hinting function, and if
so assume it is a promise.

Internally also convert all hinters to use the asynchronous interface
for consistency and simplification of code paths.

Closes #3971

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -109,14 +109,12 @@
 
     update: function(first) {
       if (this.tick == null) return;
-      if (!this.options.hint.async) {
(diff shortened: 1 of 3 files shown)


marijnh:
Thanks. Merged as 9e1ad23ab756707fc and followed up with 87d10fa06bfe6

Carreau:
Thanks ! I didn't knew about `hintOptions` that seams undocumented.

PR #3975, opened 2016-04-19
Update valid string prefix for Python 3.

Carreau:
`u` prefix are valid (again) since 3.3, and `f` prefix are now allowed
starting at 3.6 (which is not released yet though).


marijnh:
Thanks. Merged as 03d684f

PR #3977, opened 2016-04-20
Add the mbox mode to mode/meta.js

Dominator008:


Dominator008:
Once this is merged, can we have a 5.14.1 release? Thanks!

marijnh:
Merged as 339e623. But I am not really prepared to start doing on-demand releases for every little fix -- thought I've automated a large part of the process, putting out a release requires some work, and minor releases clutter up the release history. So unless some bad bug sneaks into a release, you'll just have to wait for the next monthly one.

PR #3982, opened 2016-04-21
:memo: alm.tools is released on npm

basarat:
https://www.npmjs.com/package/alm


marijnh:
Nice! Merged as 255d019

PR #3989, opened 2016-04-23
sublime.js: auto-indent after insertLine

petrzpav:
Auto-indent inserted line (Ctrl + Enter) like in Sublime Text.


marijnh:
Looks good. Merged as 9ac5d20

PR #3990, opened 2016-04-25
Include all ASCII control codes in specialChars

ngn:
* `\x1a`-`\x1f` were missing though it looks like the intention was to cover all below `\x20` (space)
* added `\x7f` which is usually considered non-printable
* removed `\t` which was already covered in `\x00-\x19` as `\x09`


marijnh:
Thanks for catching that. Merged as bef43ea

PR #3992, opened 2016-04-26
Disables wrapping on whitespace in CodeMirror-dialog by default

### LINKED ISSUE #3391: lint markers tied to a particular doc
Is it possible to tie a lint call to a specific doc? 
In cases where there is a single editor and multiple doc, this is going to be very useful. 
My current issue is that I have this async lint process which might take up to 3 or 4 seconds to complete. If during that 3-4 seconds we switch docs (swapDoc), then the markers will appear in the new doc which is totally wrong. 
Do you have any idea how to fix this? I'd be willing to code a fix if you give me some hints. Thanks.

dbarnett:
Fixes #3391.

### DIFF
--- a/addon/dialog/dialog.css
+++ b/addon/dialog/dialog.css
@@ -6,6 +6,7 @@
   padding: .1em .8em;
   overflow: hidden;
   color: inherit;
+  white-space: nowrap;
 }
 
 .CodeMirror-dialog-top {


dbarnett:
The other option is for vim.js to pass inner prompt content with an explicit nowrap style, and leave dialog.css alone. Hard to say which is a better default for dialogs in the abstract. But if/when you do want wrapping, you can always wrap the inner content in an explicit `white-space: normal` style.

marijnh:
I definitely think doing this in vim.js is a better idea -- turning off line wrapping in all prompts is going to make some prompts impossible to read. I've pushed 0458827 as an alternative.

PR #3994, opened 2016-04-26
[fix] `state.localMode.startState` should check `undefined`

jaredleechn:
when I want to highlight ***```diff*** in section of markdown, it makes an error like

> markdown.js:221 Uncaught TypeError: state.localMode.startState is not a function

I think [markdown mode](https://github.com/codemirror/CodeMirror/blob/master/mode/markdown/markdown.js#L221) should check undefined or null for some case like [diff mode](https://github.com/codemirror/CodeMirror/blob/master/mode/diff/diff.js#L22)

image


marijnh:
There's a static helper method on the `CodeMirror` object just for this. I've pushed an alternative patch that fixes all modes that are calling a sub-mode's `startState` directly.

jaredleechn:
👌

PR #4000, opened 2016-04-29
Clj mode update

carocad:
Hey Marijn,
The release of Clojure v1.7 introduced some new functions which were missing in Clojure's mode, so I 
I added the new function names. 

While I was at it, I splitted those huge line-strings into multiple ones. At least for me, it is easier to work like that. But just tell me if it goes against your standards.

Main thing is to get the new function names there.


marijnh:
Thanks. Merged as 85ef20aa66a408f and ea22650592d378

PR #4002, opened 2016-05-01
Update URL for CodeWorld in real world uses

cdsmith:


marijnh:
Nice URL. Merged as f9b042f

PR #4003, opened 2016-05-02
Add Ctrl+K Ctrl+D support to skip last selection

petrzpav:
Support Ctrl+K Ctrl+D according to [Sublime Text behavior].(https://www.sublimetext.com/docs/3/multiple_selection_with_the_keyboard.html
)


marijnh:
The way I understand Ctrl-K Ctrl-D, it doesn't require you to keep global state that way -- just deselect the primary selection, and add a new selection after it. Is there something I'm missing?

PR #4004, opened 2016-05-02
[css mode] Update Grid Layout properties and values

mrego:
Add gutters properties (grid-gap, grid-column-gap and grid-row-gap):
https://www.w3.org/TR/css-grid-1/#gutters

Take advantage to add some missing keywors:
* grid and inline-grid: https://www.w3.org/TR/css-grid-1/#grid-containers
* dense: https://www.w3.org/TR/css-grid-1/#grid-auto-flow-property

It's been a while since the list of properties has been updated:
https://drafts.csswg.org/css-grid/#property-index


marijnh:
Thank you. Merged as 601cb37

PR #4016, opened 2016-05-09
Fix typo in Fortran demo

TDaglis:
In Fortran demo page, it is stated that `text/x-Fortran` MIME type is defined, while the correct MIME type is with lowercase `f`


marijnh:
Thanks! Merged as 90d8250

PR #4019, opened 2016-05-11
[vim mode] Fix fat-cursor border

TDaglis:
Any theme with css rules about the cursor results to the normal cursor being displayed along with the fat cursor in vim mode, as seen below:
image

This patch makes the `border: 0;` of the fat cursor `important` to avoid this issue.


marijnh:
Thanks! Merged as 5067bae

Issue #4021, opened 2016-05-12
Attempting to view a long file triggers unresponsive script alerts


2is10:
Building the initial document tree can take minutes in Chrome and 8-9 seconds in Safari and Firefox for a file with 3 million short lines, even with line wrapping disabled. Unresponsive script alerts appear in Chrome and Firefox.

I understand that this project may not be interested in supporting files of this size, but it turns out that not much code needs to change to bring all three browsers down to a 2-3 second range, and the speedup should also apply to operations on smaller files, especially on devices with less CPU and memory available. I’ll send a pull request momentarily.

Mac OS X El Capitan 10.11.2 (15C50)
Chrome 51.0.2704.36 beta (64-bit)
Firefox 46.0.1
Safari 9.0.2 (11601.3.9)

2is10:
Here are fiddles [**without the fix**](https://jsfiddle.net/sh183bjs/) and [**with the fix**](https://jsfiddle.net/oacbekvy/).

See the console log for running times. Here’s what I see:

```
           Original    With Fix
           ========    ========
Chrome      14.941s      1.824s
Firefox     19.440s      1.736s
Safari       6.101s      0.862s
```

marijnh:
Thanks! See #4022

PR #4022, opened 2016-05-12
Optimizing BranchChunk.insertInner for large files

2is10:
Fixes issue #4021 by avoiding repeated splicing of large arrays.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -7315,13 +7315,20 @@
         var child = this.children[i], sz = child.chunkSize();
         if (at <= sz) {
           child.insertInner(at, lines, height);
-          if (child.lines && child.lines.length > 50) {
-            while (child.lines.length > 50) {
-              var spilled = child.lines.splice(child.lines.length - 25, 25);
-              var newleaf = new LeafChunk(spilled);
-              child.height -= newleaf.height;
-              this.children.splice(i + 1, 0, newleaf);
-              newleaf.parent = this;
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for looking into that! Merged as 5cf02c9187a8, and updated in 9ad383797c28 to simplify, remove comparison with `undefined` (which worked, but is a little too obscure), and fix the way the new chunks are inserted into the parent `children` array (using `push` only works when the content is inserted into the last child, which is always the case when loading content, but not when, for example, pasting a big chunk in the middle of a file).

2is10:
Thanks for promptly evaluating and accepting the patch, and for the fix and improvements.

My company’s a prelaunch startup without a revenue stream yet, but I’ll see if I can get you a recurring contribution of some size. We appreciate all of your fantastic work on CodeMirror.

marijnh:
Woah, I somehow missed the second part of this patch (the spilling of branch chunks) when reviewing, and that turned out to be broken, corrupting documents when gradually grown above ~500 lines. Reverted that part in 0d4374d10ee32b

2is10:
Yikes! I thought the new code was functionally equivalent. My patch touched both functions b/c the first calls the second, and both were doing repeated splicing of large arrays. I’m sorry for the trouble.

marijnh:
Yeah, the splicing is less large-scale in the second, but it would probably still benefit from being optimized. We'll just have to figure out where the bug is.

Issue #3983, opened 2016-04-21
Make lineWiseCopyCut function properly


matthew-hayes:
I don't even understand how one is supposed to use the current implementation of lineWiseCopyCut conveniently.  
Please make this function similar to how Sublime does it!  It's a great feature when it actually works right :P

Issues:
 - If the cursor is in the middle of a line of text and you do Ctrl+C/Ctrl+V, text is pasted at the current cursor location instead of at the beginning of the next line... How often do you actually want to paste an entire line into the middle of another line?

 - If you're currently at the end of a line and do enter instead of right arrow before Ctrl+C/Ctrl+V to help avoid the above problem, then the resulting pasted line of text has the wrong indentation due to auto-indent of Enter.  It would be nice if Ctrl+V of a whole line undoes any auto-tabbing. 

- After Ctrl+V, cursor should be at corresponding position of the pasted line as either position in current line or position when Ctrl+C/X was applied to get line on clipboard.  

To fix these things all we really need to be able to do is change the user's cursor position before and after the current implementation takes place.  This should be perfectly doable in most modern browsers, e.g.:
http://stackoverflow.com/questions/512528/set-cursor-position-in-html-textbox

I think this would really improve coding workflows that use this feature in CodeMirror, so thank you for any attention this issue gets :D

matthew-hayes:
Broken by 28abd453ff7f6cf9b3ca35f74a2de6ff4e00d339 @marijnh

marijnh:
Argh, sorry, see 20d981f

PR #4028, opened 2016-05-17
LineWiseCopyCut: Make this function usefully

matthew-hayes:
See https://github.com/codemirror/CodeMirror/issues/3983 for complaints, 
this change makes it so that if the user only has empty selections and does a paste after a line-wise copy or cut, the line will be pasted at the beginning of current line instead of at the actual cursor position.  This makes the line-wise copy/cut feature way more convenient and useful.  E.g. user can simply do ctrl+C and then ctrl+V to duplicate a line quickly and keep the cursor in the same position.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1102,7 +1102,7 @@
   var lastCopied = null;
   // This will be true when the last copy/cut occurred with no selection
   // and lineWiseCopyCut was enabled.  This allows us to alter the paste
-  // behavior to be more convenient when lineWiseCopyCut is enabled. 
+  // behavior to be more convenient when lineWiseCopyCut is enabled.
   var lastCopiedLinewise = false;
 
   function applyTextInput(cm, inserted, deleted, sel, origin) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as 981217306, slightly updated in 889098f123

Pro tip: Next time, don't complain and whine in issue reports. You didn't pay for this software, and being rude and entitled about it does not put me into a mood where I want to collaborate with you.

matthew-hayes:
I am very sorry my issue report came across as "whiny".  For future
reference, are you mostly referring to the first line, which I could
understand may have come across that way?  I realize that I was a bit
frustrated at the moment, and was trying to express that to some extent as
a user to justify the issue report -- I imagine there are others that have
used the feature have had similar experiences.  Prior to familiarizing
myself with this code I had no real experience with text editor code, so I
was just trying to explain my experience as a user in as much detail as
possible.  Someone had suggested to me that this may not be possible in a
browser environment so I tried to do a little research (without
familiarizing myself with all the code) to ensure it was possible.  I'm not
sure why exactly the rest of the report besides that first line would come
off as "rude and entitled", as I am pleasantly asking and thanking for any
help received?  I just had a feeling that this would have taken ~5 minutes
to fix for anyone that was already familiar with the code.

Sorry, again, I did not mean to offend anyone.  Thank you so much for all
your help with the pull request and cleanup!
Matt

marijnh:
Re-reading the original issue, it isn't as bad as I remembered it. The subject commanding me to make the feature 'function properly' grated. I get a lot of email, most of which are requests for unpaid assistance, and that can get on one's nerves when there's other things to do. But since you submitted a solid implementation yourself, we're definitely good here.

PR #4029, opened 2016-05-17
Update the documentation for htmlembedded mode

SteveLTN:
1. Mention multiplex as a dependency
2. Add ERB as a MIME-type


marijnh:
Thanks. Merged as cdabb67

PR #4031, opened 2016-05-19
FIX #4030 -- Initialize new lastCopied object properly

### LINKED ISSUE #4030: Pasting using ctrl-v doesn't work in codemirror
It gives an error "codemirror.min.js:79 Uncaught TypeError: Cannot read property 'lineWise' of null"

You can try any demo example on codemirror website, for example http://codemirror.net/demo/activeline.html

Could you fix this soon? It breaks our product. Thanks.

matthew-hayes:


marijnh:
That would just cause a different problem when line 1113 tries to run `lastCopied && lastCopied.text.join("\n") == inserted`. I've pushed an alternative patch.

matthew-hayes:
It looks like your fix breaks my feature though.

We could do my solution and change line 1113 to:
lastCopied.text && lastCopied.text.join("\n") == inserted

On Thu, May 19, 2016 at 12:37 AM, Marijn Haverbeke <notifications@github.com

PR #4035, opened 2016-05-23
added support for codefolding with mobile (touchscreen) devices

callodacity:
This is my first Pull-Request, so bear with me.

I've just made a small modification so that CodeMirror recognises touch events inside the gutter (for use with `foldcode.js` addon).

Please let me know if I've done this wrong in any way. 
I read the `contributing.md` and it's passed all the tests.
Open to constructive feedback 😄

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -3443,6 +3443,7 @@
       return dx * dx + dy * dy > 20 * 20;
     }
     on(d.scroller, "touchstart", function(e) {
+      if (clickInGutter(cm, e)) return;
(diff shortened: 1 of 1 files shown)


marijnh:
The current idea is that browsers will fire mouse events if we don't handle the touch events. Does that not work in this case? (Do you maybe have your own touchstart handler that's calling `preventDefault`?)

callodacity:
Hmm ok I'll look into that!

For reference, I'm using Ionic2 (built upon Angular2) which is a framework for deploying to iOS, Android and Windows Phones, etc.

Everything seemed to work well except the gutters so I assumed that support hadn't been built in yet. 

Thankyou :D

PR #4037, opened 2016-05-25
NEW MODES: WordPress Post Editor - Shortcode + HTML

arniebradfo:
I created a set of modes for helping code in the WordPress TinyMCE post/page editor. Its part of a plugin I'm developing. I thought I might as well add it to the source. 

There are two new modes:
- **The shortcode mode:** Kinda like a WordPress version of xml with square brackets instead of angle brackets.
- **The WordPress post editor mode:** Combines the shortcode mode with HTMLmixed and some other WordPress specific stuff.

I put both of them in the same folder because I couldn't think of an situation where the shortcode mode would be used outside of a WordPress post editor.

Let me know what you think.

Thanks!
-James

### DIFF
--- a/mode/index.html
+++ b/mode/index.html
@@ -152,6 +152,7 @@ option.</p>
       <li><a href="vhdl/index.html">VHDL</a></li>
       <li><a href="vue/index.html">Vue.js app</a></li>
       <li><a href="webidl/index.html">Web IDL</a></li>
+      <li><a href="wordpresspost/index.html">WordPress Post</a> (HTMLmixed + shortcodes)</li>
       <li><a href="xml/index.html">XML/HTML</a></li>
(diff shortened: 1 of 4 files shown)


marijnh:
Hi James, thanks for writing more CodeMirror modes. I am planning to, in the coming months, start splitting off language modes from the main editor package, because the burden of maintaining all of them is becoming too much. As such, I'd appreciate if you'd set up a package, with a name like `codemirror-mode-wordpress` or so, and published it to NPM yourself. I don't want to merge this, just to split it out again next month.

arniebradfo:
That makes sense.

Ill wait till you split the modes out so I can copy their file structure and formatting faster.

Cheers,
-James

PS your profile pic makes me feel sad

Issue #4023, opened 2016-05-12
[sublime] Backspace for multi-cursor deletion broken for whitespace


mtaran-google:
repro:
1. open https://codemirror.net/demo/sublime.html
2. put cursors right before the starting " marks on two lines
3. press backspace

expected:
both cursors delete one char before

actual:
only the main cursor deletes one char before

marijnh:
Fixed by reverting 8dbb84f4

mtaran-google:
smart backspace is a good idea overall -- why not just apply it to all cursors?

marijnh:
I am not seeing this happen with by default in Sublime Text, and I currently have more pressing things to do than redoing people's pull requests. If someone wants to submit a proper implementation that'd be welcome, though.

PR #4038, opened 2016-05-27
[sublime] Implement smart backspace with multi-cursor support

vincentwoo:
Fixes the problem raised in #4023 by @mtaran-google, and reintroduces smart backspace support for the sublime keymap.

Does this look like a sane way to handle multicursor deletion? I took a look at what other commands were doing, and I assume my handling of what to do in the default non-smart case is equivalent to:

`delCharBefore: function(cm) {cm.deleteH(-1, "char");},`

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -420,6 +420,34 @@
 
   map[cK + ctrl + "Backspace"] = "delLineLeft";
 
+  cmds[map["Backspace"] = "smartBackspace"] = function(cm) {
+    if (cm.somethingSelected()) return CodeMirror.Pass;
+
+    cm.operation(function() {
+      var cursors = cm.listSelections();
+      var indentUnit = cm.getOption("indentUnit");
+
+      for (var i = cursors.length - 1; i >= 0; i--) {
+        var cursor = cursors[i].head;
+        var toStartOfLine = cm.getRange({line: cursor.line, ch: 0}, cursor);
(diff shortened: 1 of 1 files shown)


vincentwoo:
Hey @marijnh, did you get a chance to take a look at this? The code in the PR is very similar to what was [previously in sublime.js](https://github.com/codemirror/CodeMirror/blob/5067baec7bbbe33633c14049d82a0198d9e168b7/keymap/sublime.js#L423), just with support for multiple cursors.

keymap/sublime.js
```diff
@@ -420,6 +420,34 @@
...
   map[cK + ctrl + "Backspace"] = "delLineLeft";
 
+  cmds[map["Backspace"] = "smartBackspace"] = function(cm) {
+    if (cm.somethingSelected()) return CodeMirror.Pass;
+
+    cm.operation(function() {
+      var cursors = cm.listSelections();
+      var indentUnit = cm.getOption("indentUnit");
+
+      for (var i = 0; i < cursors.length; i++) {
```
marijnh: As far as I can see, this will only work if you iterate over the selections in inverse order -- right now, your deletes will, if multiple cursors are on the same line, misalign later deletes, which are still under the impression that the cursor is in the old place.

vincentwoo: Great catch. I've just applied your suggested fix.

marijnh:
Thanks, looks good now. Merged as 79bec105b4

PR #4044, opened 2016-06-01
[soy mode] Support backslash in strings

vrana:


marijnh:
Thank you. Merged as dabefce

PR #4046, opened 2016-06-02
Add clearOnEnter as foldOption

shw6rn:
Add option 'clearOnEnter' to foldcode.js and default it to true. This option determines whether the marker added when folding a line clears when the cursor enters it.


marijnh:
Thanks. Merged as bda854e

Issue #4041, opened 2016-05-31
Code folding arrow does not change appearance when collapsing within single line 


shw6rn:
When collapsing a section of code which is on a single line, the arrow does not toggle appearance.

To Reproduce: In the [Javascript code folding demo](https://codemirror.net/demo/folding.html), collapse a line where the opening and close characters are on a single line (e.g. line 20). Notice that the arrow does not change to closed appearance.

I would expect the arrow to change to the closed appearance when the line is collapsed.

I investigated a bit and found that this occurs because when `foldgutter.js` looks for marks at a given line in `isFolded`, the `ch` field on the cursor position is `undefined`. The `clipPos` function is called inside `findMarksAt` and sets the `ch` to the end of the line, which is outside the range of the marker span on the line. We verified this was the problem by commenting out the call to `clipPos` and allowing all spans to be added to `markers` if the `ch` field is undefined. I'm sure somebody who has more familiarity with the code can come up with a better solution.
I am on Code Mirror 5.12.0 and reproduced this on Chrome (but this will happen on all browsers)

shw6rn:
Fix was merged in #4048

PR #4048, opened 2016-06-02
Change arrow appearance when a single line is collapsed

shw6rn:
This pull request addresses https://github.com/codemirror/CodeMirror/issues/4041

Add findMarksOnLine function which finds all marks on a line. Use findMarksOnLine in foldGutter.js to correctly find marks when a single line is collapsed.

### DIFF
--- a/addon/fold/foldgutter.js
+++ b/addon/fold/foldgutter.js
@@ -50,7 +50,7 @@
   }
 
   function isFolded(cm, line) {
-    var marks = cm.findMarksAt(Pos(line));
+    var marks = cm.findMarks(Pos(line, 0), Pos(line + 1, 0));
     for (var i = 0; i < marks.length; ++i)
       if (marks[i].__isFold && marks[i].find().from.line == line) return marks[i];
   }


marijnh:
This pull request appears to just add a `findMarksOnLine` method (which is already well covered by `findMarks`), and doesn't do anything else. Did you forget a patch? (In any case, use `findMarks`, no need to add a new method.)

marijnh:
... Oh, on closer look, the patch would have an effect, due to the changed meaning of the call in foldcode.js. I get what it does. Could you do it differently?

shw6rn:
Yeah I could use `findMarks` instead

shw6rn:
@marijnh I tried to use `findMarks` and there is an issue with that approach.
When you use `findMarks` to find all marks on line like so...
```
var line = cm.getLine(lineNumber);
var marks = cm.findMarks(CodeMirror.Pos(lineNumber, 0), Pos(lineNumber, line.length));
```
If the mark starts on the very end of the line, the fold mark is not returned from `findMarks` because the cursor position at the end of the line is the same `ch` position as the start of the mark (see https://github.com/codemirror/CodeMirror/blob/master/lib/codemirror.js#L7642)

We could resolve this issue in one of two ways.
1. Add the `findMarksOnLine` function as is
2. Change `findMarks` so that if the we are at the end of the line, use `span.from > to.ch` instead of `span.from >= to.ch` - this seems a little hacky and I'm not sure if this would have any adverse effects

shw6rn:
Actually, `findMarks` looks like it will work if I go from from `Pos(lineNum, 0)` to `Pos(lineNum + 1, 0)`

shw6rn:
@marijnh I made the fix using `findMarks` instead. Could you take a look?

marijnh:
Thanks, looks good now. Merged as 431dd1b

PR #4049, opened 2016-06-02
Add a dataset option to markText

alur:
Lets you attach something like an id to a piece of code; without having to worry about it showing up on hover, like it would if you used title.


marijnh:
Firstly, `dataset` isnt supported by all browsers CodeMirror supports.

But also, I think it's a better idea to go through the actual mark value (using `findMarksAt`) to get at any info you associated with the mark, rather than pushing all that into the DOM, which is more expensive and messy.

alur:
Fair enough.

PR #4050, opened 2016-06-03
Prevent native search dialog from opening when persistent search dialog is open

40thieves:
Prevents events for find commands when the persistent search dialog is open. This helps with accidental extra key presses.

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -68,10 +68,7 @@
         var overriddenCommands = ['find', 'findPersistent', 'findNext', 'findPrev'];
 
         if (keyCommand && overriddenCommands.indexOf(keyCommand) !== -1) {
-          startSearch(cm, getSearchState(cm), val);
-          doSearch(cm, ev.shiftKey, true);
-          CodeMirror.e_stop(ev);
-          return false;
+          f(val, ev);
         }
       }
     });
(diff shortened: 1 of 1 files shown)


marijnh:
Would it make sense to also distinguish _which_ command the key refers to, and execute the appropriate action? It seems that for some of these commands, we could just execute the command itself, no?

40thieves:
Hmm, yes I can how that would be the better approach.

However I've just tested this out and there's a problem when calling the command for the first time. So for example: 

`find` command -> type "Hello" -> `findNext` command

Here, the search state has not been set to "Hello", and so the search is not run correctly. I suppose it could check if the search state is set and if not, trigger a `find` command.

40thieves:
I just realised I could call the callback instead of the `find` command - as this handles checking the search state

40thieves:
Updated to use the callback instead of custom logic. This doesn't necessarily solve the original problem, but it does clean up the original implementation.

marijnh:
Any way we can at least make `findPrev` search backwards in this situation?

40thieves:
Updated again. Now `findNext` and `findPrev` commands will be executed (after setting the search state), whereas `find` and `findPersistent` will act as if enter was pressed.

I think this is a good heuristic...

marijnh:
Thanks, that's what I was also thinking of. Merged as 5c6025a and adjusted a bit in b7b4269

40thieves:
Great! Thanks for the help in working out the implementation

PR #4051, opened 2016-06-03
Persistent search - find next

40thieves:
We're switching over from the `find` to `findPersistent` functionality (which we prefer) I've found that it can be a little confusing for users who are used to the "old" find method - e.g. there is no way to open the persistent search and jump to the next or previous match (which is our the current find next/previous behaviour).

This PR adds two new search commands `findPersistentNext` and `findPersistentPrev` which do this. I would have liked to do this just using the existing search functions but I could't find a hook to be able to trigger a search immediately after the persistent dialog opens. I therefore added an `onOpen` hook.

Note: I've also included the changes from #4050 because this PR is a bit more destructive and so I wanted to split them up. However I've included it here as I realised that the extra commands would also need to be [caught](https://github.com/codemirror/CodeMirror/compare/master...40thieves:persistent-next?expand=1#diff-b39f347c31c96f6bba9d7d2aef74c267R69). Let me know if this is a problem and I'll pull it out.

Happy to make adjustments with your feedback 😄

### DIFF
--- a/addon/dialog/dialog.js
+++ b/addon/dialog/dialog.js
@@ -65,6 +65,8 @@
         }
       }
 
+      if (options.onOpen)
+        options.onOpen(inp.value, close);
       if (options.onInput)
         CodeMirror.on(inp, "input", function(e) { options.onInput(e, inp.value, close);});
       if (options.onKeyUp)
--- a/addon/dialog/dialog.js
+++ b/addon/dialog/dialog.js
@@ -65,8 +65,6 @@
         }
       }
 
-      if (options.onOpen)
-        options.onOpen(inp.value, close);
       if (options.onInput)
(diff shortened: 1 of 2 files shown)


marijnh:
Why are you putting the init logic in an `onOpen` callback, rather than running it directly after opening the dialog?

40thieves:
Yep, you're totally right. Thanks for the feedback 😄 . I've updated to remove the callback and perform the immediate search in `doSearch` instead.

marijnh:
Merged as 440395d

40thieves:
Thanks 👍 . Do you have a timeline for a release?

marijnh:
I release around the 20th of each month, so in about two weeks.

PR #4058, opened 2016-06-07
Issue Fix #4057

### LINKED ISSUE #4057: Sql-Hint Bug [ConvertNumbertoCur]
The ConvertNumberToCur method in addon/hint/sql-hint.js taken in input a number in floating point representation and converts it back to a Position object. The input number is obtained by giving Pos object as input to another method ConvertCurToNumber.

If the input does not have all the 6 digits after decimal, it results in incorrect outputs
For Example  
If Input is 1.00003 , it is resolved to Pos (1,3 ) in place of Pos (1, 30 ) which results in incorrect parsing of results.

harshvardhan145:
Issue fixed #4057

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -184,7 +184,7 @@
   }
 
   function convertNumberToCur(num) {
-    return Pos(Math.floor(num), +num.toString().split('.').pop());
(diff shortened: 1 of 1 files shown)


marijnh:
Merged as ecd6976f2be2, but followed up by 92e9634f25 which removes that rather pointless conversion entirely.

PR #4068, opened 2016-06-14
[go mode] Add `error` as highlighted type

mtaran-google:
It is, per https://golang.org/ref/spec#Predeclared_identifiers


marijnh:
Thank you! Merged as 047afd2

PR #4073, opened 2016-06-17
Use key definition from keyMap before calling call

dapatil:
This fixes behaviour that is broken in the vim keymap. Specifically keys explicitly defined in vim-insert and vim-replace modes do not work.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -5871,7 +5871,7 @@
 
   var lookupKey = CodeMirror.lookupKey = function(key, map, handle, context) {
     map = getKeyMap(map);
-    var found = map.call ? map.call(key, context) : map[key];
+    var found = map[key] || (map.call ? map.call(key, context) : null);
     if (found === false) return "nothing";
     if (found === "...") return "multi";
     if (found != null && handle(found)) return "handled";


marijnh:
What do you mean by 'keys explicitly defined in [...] modes'? Keys you added to those modes from external code? I don't think the vim mode is supposed to be used that way.

dapatil:
Hi Martin, 
I should have been clearer in my comment.
Look at: https://github.com/codemirror/CodeMirror/blob/master/keymap/vim.js#L4771

You'll see that the keymap defines a separate "vim-insert" keymap and explicitly defines what "Ctrl-N", "Ctrl-P" and "enter" do. I believe the assumption there was that keys defined explicitly there would be handled without going through the "call" handler. This change makes vim keymap work in vim's insert and replace modes.

vincentwoo:
Can you give a reproducible error that this fixes? What steps should we follow on https://codemirror.net/demo/vim.html to cause broken behavior?

marijnh:
@vincentwoo You can enter replace mode in the vim demo (i, insert) and then press backspace. The keymap tries to bind this to goCharLeft, but it is ignored.

This is odd. These seem to have been there for a long time, but they never worked since the rewrite in late 2012.

Such bindings, even if we fix them with the proposed patch, wouldn't work with vim's macro recording and multi-stroke keys, though. Also, keymaps with a `call` method were never supposed to also behave like regular keymaps, so I'm not really happy about making this change.

I think I'm gong to consider these bindings a bug in the vim mode, and would prefer for them to be fixed there. Unfortunately, the vim mode is without maintainer, and I don't really know how it works.

vincentwoo:
@mightyguava I know you're not active developing the vim mode, but this might be an important enough fix to take a look?

mightyguava:
Sure. I'll try to take a look at this tomorrow, or latest by Friday.
Cheers,Yunchi






On Mon, Jun 20, 2016 at 5:20 AM -0400, "Vincent Woo" <notifications@github.com> wrote:












@mightyguava I know you're not active developing the vim mode, but this might be an important enough fix to take a look?

mightyguava:
Okay, yeah looks like this probably broke in https://github.com/codemirror/CodeMirror/commit/4fad3f82c4a40ef91883d04b83f1c1751b6c022a where I added the `call` handlers. I undid most of the work in that commit later but didn't revert that part.

I've moved the Backspace handler the `defaultKeyMap`, and removed the autocomplete bindings. If you want to add custom bindings like `autocomplete`, you should be able to do it via the [vim mode API](https://codemirror.net/doc/manual.html#vimapi).

It's been quite a while since I touched this project so I'm not fully confident on the changes. Please see if it looks sane and I'll merge it https://github.com/codemirror/CodeMirror/pull/4088

mightyguava:
@vincentwoo

vincentwoo:
That seems sane to me but I am by no means a vim-mode expert either.

marijnh:
I've merged the patch. So the recommendation for adding key bindings to a vim-enabled editor is to add a _separate_ keymap with the `addKeyMap` method, and put your bindings in there.

PR #4074, opened 2016-06-17
Mimic sublime on OS X with ctrl-shift for selectLines

jeffjenkins:
The existing behavior seems like it's using the linux or windows keyboard shortcuts. This maintains those but adds ctrl+shift, which is the mac default


marijnh:
Thanks. Merged as 3aba5a8 . Though I'm not sure why you claim it maintains the existing bindings -- it doesn't. It changes the bindings on non-Mac platforms from the Linux to the Windows defaults, and does not leave those bindings on Mac. I guess this is okay (if I understand the ST docs correctly, which are sort of opaque because of the inconsistent and cryptic symbols they use)

PR #4075, opened 2016-06-17
Use Sublime Text word/subword movement with alt/ctrl

jeffjenkins:
The only per-word movement in the existing sublime controls is sub-word, but it uses the full word keyboard shortcuts. This PR switches the alt shortcut to be full word and adds ctrl as subword to sync up with Sublime's defaults

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -43,6 +43,15 @@
     return Pos(start.line, pos);
   }
 
+  function moveWord(cm, dir) {
+    cm.extendSelectionsBy(function(range) {
+      if (cm.display.shift || cm.doc.extend || range.empty())
+        return cm.findPosH(range.head, dir, "word");
+      else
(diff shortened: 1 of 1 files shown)


marijnh:
You once again changed the behavior in a way that breaks the default linux bindings. I'm going to hold off merging/fixing this until someone can point me at defining information about per-platform bindings for ST. Unfortunately, and bafflingly, no such thing seems to exist online (at least in the first page of google results)

marijnh:
They are probably also in some JSON file in the ST distribution tarballs.

marijnh:
The main library is recognizing windows with `/win/i.test(navigator.platform)`, so I guess you could use that, the existing `mac` check, and just call everything else 'unix'? That's not super accurate, but close enough for this purpose, I think.

I do prefer to get a single patch (use amend or rebase), but if you're having trouble I can also do the squashing myself.

PR #4084, opened 2016-06-23
Add Panda Syntax to CodeMirror

siamak:


marijnh:
I'm no longer accepting new languages into the main distribution (and will start splitting off the existing language modes into separate modules in the near future). As such, I ask that you distribute and maintain this yourself, preferably under some easily-findable name like `codemirror-mode-panda`.

ficristo:
@marijnh This seems a theme and not a mode. Do you are going to split themes too?

marijnh:
Eventually, yes, but for now I guess they are low-impact enough to merge a few more. Merged this as a3d9a31

siamak:
Thanks ;) @marijnh

Issue #4082, opened 2016-06-23
[markdown mode] image alt-texts should have more token styles


HoverBaum:
So you define an image with `![alt text](http://link.to/image.jpg)` in markdown.

The markdown mode renders this to: 
  ``` 
    <span style="padding-right: 0.1px;">
        <span class="cm-tag">![alt text]</span>
        <span class="cm-formatting cm-formatting-link-string cm-string cm-url">(</span>
        <span class="cm-string cm-url">http://link.to/image.jpg</span>
        <span class="cm-formatting cm-formatting-link-string cm-string cm-url">)</span>
    </span>
```

While a link with similar Syntax `[link text](http://link.url)` becomes

```
    <span style="padding-right: 0.1px;">
        <span class="cm-formatting cm-formatting-link cm-link">[</span>
        <span class="cm-link">linkText</span>
        <span class="cm-formatting cm-formatting-link cm-link">]</span>
        <span class="cm-formatting cm-formatting-link-string cm-string cm-url">(</span>
        <span class="cm-string cm-url">http://google.com</span>
        <span class="cm-formatting cm-formatting-link-string cm-string cm-url">)</span>
    </span>
```

Images should also come with the information to style the alt-text. So that a design can create a cohesive look. Images should render to:

```
    <span style="padding-right: 0.1px;">
        <span class="">!</span>
        <span class="cm-formatting cm-formatting-link cm-link">[</span>
        <span class="cm-link">alt text</span>
        <span class="cm-formatting cm-formatting-link cm-link">]</span>
        <span class="cm-formatting cm-formatting-link-string cm-string cm-url">(</span>
        <span class="cm-string cm-url">http://link.to/image.jpg</span>
        <span class="cm-formatting cm-formatting-link-string cm-string cm-url">)</span>
    </span>
```

I will aim to create a Pull Request implementing this some time next week.

HoverBaum:
After some more consideration I would adjust the expected render to:

```html
<span style="padding-right: 0.1px;">
    <span class="cm-formatting cm-formatting-image cm-image cm-image-marker">!</span>
    <span class="cm-formatting cm-formatting-image cm-image cm-image-alt-text">[</span>
    <span class="cm-image cm-image-alt-text">alt text</span>
    <span class="cm-formatting cm-formatting-image cm-image cm-image-alt-text">]</span>
    <span class="cm-formatting cm-formatting-link-string cm-string cm-url">(</span>
[...]

PR #4092, opened 2016-06-30
Refine image highlighting in markdown mode

HoverBaum:
According to the format suggested in #4082 add more information to images.

This implements formatting as suggested in the issue and adds tests for the implemented tokens. Also updates some tests with the new format.

With highlighting mode an image will now look like this:

```html
<span style="padding-right: 0.1px;">
    <span class="cm-formatting cm-formatting-image cm-image cm-image-marker">!</span>
    <span class="cm-formatting cm-formatting-image cm-image cm-image-alt-text">[</span>
    <span class="cm-image cm-image-alt-text">alt text</span>
    <span class="cm-formatting cm-formatting-image cm-image cm-image-alt-text">]</span>
    <span class="cm-formatting cm-formatting-link-string cm-string cm-url">(</span>
    <span class="cm-string cm-url">http://link.to/image.jpg</span>
    <span class="cm-formatting cm-formatting-link-string cm-string cm-url">)</span>
</span>
```

Giving a great deal of freedom to styling it however you like.

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -444,7 +444,6 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
     }
 
     if (ch === '[' && state.imageMarker) {
-      console.debug('imageAltText')
       state.imageMarker = false;
       state.imageAltText = true
       if (modeCfg.highlightFormatting) state.formatting = "image";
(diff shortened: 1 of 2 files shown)


marijnh:
Theme's don't come with an `image` style, so removing the `tag` styling will cause these to show up unstyled for everybody who isn't adding custom styling. That's not a good idea -- please preserve the `tag` styling.

marijnh:
I'm okay with using the link tag, since there'll be additional classes to target for people who want to style these differently.

HoverBaum:
Added `link` and adopted the tests accordingly.

marijnh:
Looks great now! Merged as 34b3a7b

PR #4093, opened 2016-07-01
[sublime] Make goSubwordLeft/Right shortcuts match Sublime on OSX

mtaran-google:


marijnh:
Thanks, merged as a0547b7

PR #4094, opened 2016-07-03
V3

pinezeng:
CodeMirror V3 pull


marijnh:
I am not sure what this is trying to do, but it doesn't seem like it's something I can work with.

PR #4100, opened 2016-07-07
[merge demo] set allowEditingOriginals to true

mtaran-google:
Makes it easier to repro diff/merge issues on the demo page


marijnh:
The page is mainly intended to show the addon though, and I think having an un-editable original makes it easier to understand how this is supposed to be used. I make one-off pages to reproduce bugs anyway, so I'm going to decline this one.

PR #4106, opened 2016-07-10
[gfm] Prevent markdown mode from parsing formatting within vanilla links (closes #4079)

### LINKED ISSUE #4079: gfm mode underscore in link url starts emphasis
Reproducible here: http://codemirror.net/mode/gfm/index.html

While using the gfm mode if there is an underscore in the url get parameters this triggers the start of an emphasis block. So if you have an inline link with an underscore the text following is emphasised.

`Here is a link, https://test.com?_test and this text is in italics`

image

I would expect the underscore within the link to be ignored for formatting.

I took a look at the code but seems like it may be a non-trivial fix as the underscore logic is in the markdown mode where as this "autolink" logic is in the gfm mode. Is it as simple as setting state.linkHref in the gfm link logic? https://github.com/codemirror/CodeMirror/blob/master/mode/gfm/gfm.js#L106

Is the state shared between modes? Sorry this is the first time at looking at how CodeMirror works so I don't have a clear understanding yet.

0b10011:
The bug that #4079 was coming across was the fact that markdown mode parses separately from gfm mode, and isn't aware of the parsing happening in gfm mode. Additionally, either an asterisk anywhere or an underscore after a non word character had to appear in the url. That leaves a few different routes for fixing that I can think of.

First, we could make markdown mode aware of autolinking, but this is pushing more of the gfm mode logic into markdown mode that should really be avoided. Additionally, putting small hacks to prevent formatting within links would be more fragile than just building in the autolinking entirely in markdown mode (and verbose, as we'd have to put it in all of the formatting logic).

Secondly, we could give gfm mode access to markdown mode's state and have gfm mode try to manage it, but that would make refactoring states within a mode more prone to breaking overlays. So, also not ideal.

Thirdly, we could simply give the gfm mode's tokens priority, which would fix the link, but everything after it would still be in a different state than it should be (italicized when it shouldn't be).

Lastly, we can force markdown to freeze state until after the link, which is ideal and the route I ended up taking. That is, the state before the link in markdown should be exactly the same as after the link. The simplest way I could think of to achieve this was to add a `freezeBaseState` flag to gfm mode that the overlay.js code would look for and skip base mode parsing when it was encountered. It would then set the base position to the overlay position so they both start parsing again from the same spot (after the link).

### DIFF
--- a/addon/mode/overlay.js
+++ b/addon/mode/overlay.js
@@ -4,8 +4,12 @@
 // Utility function that allows modes to be combined. The mode given
 // as the base argument takes care of most of the normal mode
 // functionality, but a second (typically simple) mode is used, which
-// can override the style of text. Both modes get to parse all of the
-// text, but when both assign a non-null style to a piece of code, the
+// can override the style of text.
+// If state.overlay.freezeBaseState is true,
+// only the overlay mode will parse the text
+// until state.overlay.freezeBaseState is set to false.
+// Otherwise, both modes get to parse all of the text,
+// but when both assign a non-null style to a piece of code, the
 // overlay wins, unless the combine argument was true and not overridden,
 // or state.overlay.combineTokens was true, in which case the styles are
 // combined.
@@ -47,7 +51,7 @@ CodeMirror.overlayMode = function(base, overlay, combine) {
         state.basePos = state.overlayPos = stream.start;
       }
 
-      if (stream.start == state.basePos) {
+      if (stream.start == state.basePos && !state.overlay.freezeBaseState) {
         state.baseCur = base.token(stream, state.base);
         state.basePos = stream.pos;
       }
@@ -55,6 +59,9 @@ CodeMirror.overlayMode = function(base, overlay, combine) {
         stream.pos = stream.start;
         state.overlayCur = overlay.token(stream, state.overlay);
         state.overlayPos = stream.pos;
+      }
+      if (state.overlay.freezeBaseState) {
+        state.basePos = state.overlayPos;
       }
       stream.pos = Math.min(state.basePos, state.overlayPos);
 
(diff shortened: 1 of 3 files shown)


marijnh:
This seems to be pulling the mode/overlay addon in the direction of the mode/multiplex one.

Do you, at a glance, know any compelling reason why the GFM mode shouldn't be implemented through hooks provided by the Markdown mode, the way the clike and sql dialects work? A lot of its complexity seems to come from overlays just beeing a rather bad fit for such an interwoven set of modes.

0b10011:
> This seems to be pulling the mode/overlay addon in the direction of the mode/multiplex one.

Yeah, I think you're right. I was thinking the multiplex addon relied more on the base mode being aware of the overlay mode.

> Do you, at a glance, know any compelling reason why the GFM mode shouldn't be implemented through hooks provided by the Markdown mode, the way the clike and sql dialects work?

Hmm, that might be an option. It would require a general `token` hook to work with URLs, SHAs, etc.

Is it possible to extend modes defined in this way, though? That is, if gfm mode was defined using hooks instead, would someone else be able to come along and extend gfm mode, adding their own hooks? Or would they have to manually copy and paste the hooks into their own mode? If there's no way to extend, I think that would be a reason not to.

> A lot of its complexity seems to come from overlays just beeing a rather bad fit for such an interwoven set of modes.

Yeah, I agree. I can't remember what made me choose the overlay addon 4 years ago, but with how it functions currently, it just isn't a good fit. Hooks may be the way to go. I'll look into it and get back to you. Thanks for the direction!

marijnh:
> Is it possible to extend modes defined in this way, though?

The idea is to provide configuration parameters and/or callbacks that allow all aspects of the language that might change between dialects to be customized. What this looks like will differ per language, and you'll probably not be able to (or want to) cover all possible dialects, but yes, the experiences with the clike, css, and sql modes has been that this is a reasonable way to do language variations -- it's definitely easier to extend than the current GFM mode.

Issue #3238, opened 2015-05-05
Copy/Cut/Paste not available in context-menu


yelvert:
This is Chrome on OSX 
![image](https://cloud.githubusercontent.com/assets/7054/7483790/8c487f08-f347-11e4-87c9-779048cdf338.png) copy/cut/paste isn't available on any browser on any OS. Im using angular ui-codemirror but ive tried initializing it manually and have the same problem so I don't think its related.

The options I'm using are:
```
$scope.editorOptions = {
              tabSize: 4,
              mode: 'none',
              viewportMargin: Infinity,
              lineWrapping: true,
              extraKeys: {
                'Shift-2': function (cm) {
                  // custom stuff :)
                }
              }
            };
```

marijnh:
Does the problem occur on the demos on http://codemirror.net? I just tried (Chrome 42, Mountain Lion), and copy/paste from the context menu worked fine.

yelvert:
Well, yes, the demos work, and after some digging I realized that the fix I did here https://github.com/yelvert/CodeMirror/commit/714cf11e6cc7d2aea96f51b6dbb0d9ccf178f689 to fix ![image](https://cloud.githubusercontent.com/assets/7054/7497332/5538bef4-f3df-11e4-9e77-fc771341f442.png) is what caused the context menu yo stop working correctly. So my real issue is the text box showing up when right clicking and never going away. If I need to create a new ticket I will.

Veejay:
I'm experiencing the same issue. The demos on CodeMirror's website are fine (I have copy/cut/paste in the context menu), but in the editor we use in our application, no dice in any browser. Is there a workaround I could use to work around this issue? Thank you.

marijnh:
Try to narrow it down -- something in your site is causing it, so remove pieces one by one until it goes away, then figure out what precise line did it. It is likely a CSS conflict, but might be something else.

roelvanduijnhoven:
I can reproduce this in Chrome using the example given in #2154 (http://output.jsbin.com/ihunin/328/). Thus that is not only a FF bug.

The editor is in my case in a modal screen which has transform CSS options. When I move the editor out of the modal screen it *does* work.

Any clue? I guess using a CSS transform should not remove copy/paste functionallity.

marijnh:
In the JSBin case, it seems to be caused by a mysterious no-op CSS transform they inject:

```
body { transform: translate3d(0,0,0); }
```

[...]

PR #4109, opened 2016-07-12
Force editor layer for faster paint.

Tihauan:
Implement the layer separation suggested in [1] to reduce the print area when scrolling. The editor will get its own GPU accelerated layer [2]. It applies to Chrome/WebKit.

Before
before

After
after

[1] [Tricks for high performance scrolling](https://groups.google.com/forum/#!searchin/codemirror/translate3d/codemirror/kQirncJtBjI/rdzEf1zOeRwJ)
[2] [Accelerated Rendering in Chrome](http://www.html5rocks.com/en/tutorials/speed/layers/)

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -5,6 +5,7 @@
   font-family: monospace;
   height: 300px;
   color: black;
+  transform: translate3d(0, 0, 0);
 }
 
 /* PADDING */


writemonkey:
This will disable editor's subpixel anti aliasing.

marijnh:
That's definitely a downside. And there are other [invasive effects](https://github.com/codemirror/CodeMirror/issues/3238#issuecomment-175849461) from applying transforms. Since this is easy enough to enable with a custom CSS rule if you really want it, I think I'd prefer not to merge it.

marijnh:
Closing until further notice.

PR #4110, opened 2016-07-12
Make match-highlight scrollbar annotations case-sensitive

alur:
To match the actual match-highlights in the editor.


marijnh:
Thanks. Merged as 8019d7c

PR #4113, opened 2016-07-13
spellcheck option

kofifus:
add spellcheck options to ContentEditableInput

to use:

```
cm = CodeMirror(elem, {
	  inputStyle: "contenteditable",
	  spellcheck: true,

```

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -592,8 +592,12 @@
     var comp = compensateForHScroll(display) - display.scroller.scrollLeft + cm.doc.scrollLeft;
     var gutterW = display.gutters.offsetWidth, left = comp + "px";
     for (var i = 0; i < view.length; i++) if (!view[i].hidden) {
-      if (cm.options.fixedGutter && view[i].gutter)
-        view[i].gutter.style.left = left;
+      if (cm.options.fixedGutter) {
+        if (view[i].gutter)
+          view[i].gutter.style.left = left;
+        if (view[i].gutterBackground)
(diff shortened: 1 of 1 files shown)


StudioJunkyard:
Check out www.grammarly.com

kofifus:
thx StudioJunkyard but that's a chrome app so totally different thing. I'm looking for multi platform / native browser support vie contenteditable.

StudioJunkyard:
@kofifus It's not just a chrome app, they have other options. Unfortunately, though they don't have an exposed public API. I have asked about it, but they neither confirmed or denied if they'd make it available in the future, I also asked about a native Linux app, which may or may not be in the works.

marijnh:
I've pushed an alternative patch c5c1e92. The option isn't in the documentation yet, and might change or be dropped in the future, but for now it's there, and you can see how well it works (seems to work reasonably well, but the way the browser shows underlines in unpredictable and not really great, due to CodeMirror's re-rendering of content)

kofifus:
Thx marijnh, 

'the way the browser shows underlines' is the main feature of spellchecking ... if it does not work predictably, then basically this patch does not work :( and personally I don't think should be added.

There are two options I can see, either fix CM to allow typo recognition predictably, which I understand may not be possible, or use an external spellchecker (as ie i've done in [gist](https://gist.github.com/kofifus/4b2f79cadc871a29439d919692099406) ) which works well but has the disadvantages of loading big dictionaries into memory and being a bit slow ...

PR #4117, opened 2016-07-15
Minor modification to clean up after hasBadBidiRects test

paulmasson:
In going through Google webmaster tools for one of my websites, I was surprised to see "AخA" listed as a keyword in my pages. I finally traced it back to your repository, since I use embedded SageMathCells and that uses CodeMirror. I've noticed that every website using CodeMirror comes up in a Google search for that three-character combination, so perhaps you left it in intentionally.

If it was not intentional, this small modification to the test function removes the leftover textnode.


marijnh:
Thanks. Merged as 43749f8

PR #4118, opened 2016-07-15
Added support for "text/babel" script type

leitzler:
In many places (react tutorial, babel-standalone usage page, etc) babel scripts can be defined using script type "text/babel" without setting any script lang.


micahstubbs:
💯 this.  would be super useful.

marijnh:
The htmlmixed mode intentionally does not include declarations for all the various non-standard script types people use. It is configurable, so the recommended way to get this to work is to just add the ones you need yourself.

PR #4124, opened 2016-07-20
fixed #3860 (deleting caracters on android)

### LINKED ISSUE #3860: Position on deleting last character in line is not calculated correctly for mobile (android)
Bug can be reproduced using latest version of CodeMirror on Android 5.1 or 4.4.2
Deleting last character in line breaks because (if I understood correctly) DOM node on which calculating position for replacing range depends is already deleted when trying to detect position.

I've managed to fix this with this condition inside pollContent() function but I'm not sure is this fix enough or did I forget to consider some case scenario?
```
if (from.line == to.line && from.ch == to.ch && fromNode != toNode) {
  // if no selection but different lines, assuming last character in previous line was selected
  var chFrom = Pos(fromLine - 1, cm.doc.getLine(fromLine - 1).length);
  var chTo = Pos(fromLine, 0);
  replaceRange(cm.doc, "", chFrom, chTo, "+input");
  return true;
}

var newText = cm.doc.splitLines(domTextBetween(cm, fromNode, toNode, fromLine, toLine));
```

domagojk:
As explained in #3860 - added condition which fixes this bug for me
(only affects mobile versions)

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -1803,6 +1803,14 @@
         var toNode = display.view[toIndex + 1].node.previousSibling;
       }
 
+      if (from.line == to.line && from.ch == to.ch && fromNode != toNode) {
+        // if no selection but different lines, assuming last character in previous line was selected
+        var chFrom = Pos(fromLine - 1, cm.doc.getLine(fromLine - 1).length);
+        var chTo = Pos(fromLine, 0);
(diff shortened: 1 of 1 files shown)


marijnh:
Commit 6019b1308d4c51 does something similar

marijnh:
(Apologies for leaving this entirely ignored for so long. The mobile/IME stuff is subtle and scary and I am not often in the mood for working on it.)

PR #4136, opened 2016-07-27
Variable width tabs via tabStops option

TheNeuralBit:
Added support for variable width tabs via a `tabStops` option.
`tabStops` defaults to `[]` which doesn't change the current behavior at all. When it is defined, tab characters will shift to the lowest tab stop that is greater than the current column. If the current column is greater than the largest `tabStop` we fall back to using the `tabSize`. At that point new tabs are offset _from the last tab stop_ rather than from the beginning of the line.

For example if we have `tabStops = [2]` and `tabSize = 4` then a sequence of tab characters will look like:
```
--|----|----|----|
```
Where - is one character width and | is a tab stop. This behavior can be changed easily enough if something else is preferred, but this seemed like the most logical to me.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -8318,7 +8318,7 @@
   };
 
   function tabStopAt(column, tabSize, tabStops) {
-    if (tabStops) {
+    if (tabStops.length > 0) {
       var stop = tabStops.find(function (d) { return d > column; });
       if (stop) {
         return stop
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -4780,8 +4780,9 @@
         pos = tabStops[i];
         indentString += "\t";
       }
-      if ( i === tabStops.length )
+      if ( i === tabStops.length ) {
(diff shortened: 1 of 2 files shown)


marijnh:
What are you planning to use this for? This is quite a bit of extra complexity, and you didn't even touch the `runmode`, `tern`, `sublime`, `vim`, and `indent-fold` addons yet, which all deal with `countColumn` or `tabSize` in some way. So I'm not sure I want to incur that complexity for something that's not typically used in code editing.

TheNeuralBit:
I admit my use case is a bit specialized - I'm not actually using codemirror as a code editor.
This is just a feature I needed for my application, and I couldn't figure out a way to do it without touching the codemirror internals. I figured I might as well contribute back the result.

But if I'm the only person interested in this feature I don'd mind working off of my own fork.

marijnh:
You might even be able to get something to work without modifying the core by using `markText`-based widgets for the tabs (though that's be somewhat more involved to manage, especially if you expect a large number of tab characters). But yeah, I'm sorry, but I'm going to decline merging this.

PR #4138, opened 2016-07-29
Make findMarks inclusive of endpoints

DylanFlanders:
This change makes findMarks include any marks located at the boundaries "from" and "to."
https://discuss.codemirror.net/t/is-findmarks-from-to-supposed-to-be-inclusive/844


marijnh:
The current behavior is intentional.

PR #4139, opened 2016-07-29
Fix gutterBackgrounds to a fixed gutter

JoelEinbinder:
When scrolling horizontally with a fixed gutter, gutter backgrounds were left behind.


marijnh:
Thank you. Merged as 6d1379e

PR #4153, opened 2016-08-08
Re run mode and overlays on refresh

kofifus:
this fix extends refresh to also re run all modes and overlays on all lines affected with refresh

This allows updating all (visible) text with any changes made to the modes or overlays which is the expected behavior for a refresh. 

Case in point, having a spellcheck overlay with an 'ignore all' option which adds a word to a list of words ignored by the spellchecker. Calling refresh will update the display so that the word will not be marked as a typo in any line.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -5336,6 +5336,7 @@
 
     refresh: methodOp(function() {
       var oldHeight = this.display.cachedTextHeight;
+      this.state.modeGen++;
       regChange(this);
       this.curOp.forceUpdate = true;
       clearCaches(this);


marijnh:
The recommended way to have configurable/changeable modes is to pass their configuration as a mode option when creating the mode, and updating the mode option (or removing/adding in case of an overlay). That will cause the highlighting to be updated. `refresh` is not meant to do this, as modes are expected to be stable and always return the same result (and a full refresh is going to be overkill for this kind of changes anyway).

kofifus:
Thx marijnh,

How should I then solve the problem with opaque overlay and overlay order described in https://discuss.codemirror.net/t/addoverlay-removeoverlay-and-opaque/835/3 ?

PR #4036, opened 2016-05-24
Overhaul Solarized theme

TimothyGu:
- Make gutter more similar to official Solarized style
   - Remove border
   - Use highlighted line color
- Custom fat cursor style

----

### Light

image

### Dark

image

----

### Light
image

### Dark: 
image

PR #4154, opened 2016-08-08
Color fix for solarized light vim mode

SuNaden:
There is a colour issue with the solarized light theme. In vim mode, the cursor is not visible because it has the same colour as the background. This screenshot has the upper brackets selected, and you can see only the brackets get highlighted, but not the cursor.

screen shot 2016-08-08 at 11 41 04

This is the fixed one with colour taken from the previous version in code mirror.

screen shot 2016-08-08 at 15 01 45

After a small research, this was most probably introduced in #4036 . The fat cursor colour is also from that issue.


marijnh:
Thanks. Merged as 524d006

PR #4155, opened 2016-08-09
fixed typo in comments

callodacity:
Self-explanatory, came across it while perusing your lovely code!


marijnh:
Thanks. Merged as 92b64d2

PR #4156, opened 2016-08-09
Fix bug where immediate search would throw error if no previous query

40thieves:
Searching with `findPersistentNext` or `findPersistentPrev` will immediately run a search with the previous query. However it will throw and error if no previous query exists.

This PR fixes this bug by preventing immediate execution if no previous query is set.

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -146,7 +146,7 @@
           searchNext(query, event);
         }
       });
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as e47ab5f

40thieves:
Awesome, thanks

PR #4158, opened 2016-08-10
Look up key command in extra keys when in persistent search

40thieves:
When the persistent search is open, the `find`, `findNext` and `findPrev` commands can be executed to jump between matches. If a keymap has been changed to add additional bindings for these commands or if bindings for `findPersistentNext` or `findPersistentPrev` have been added, then the search will not behave in the same way.

This PR adds a look up in `extraKeys` for the key bind, and adds `findPersistentNext`/`findPersistentPrev` as valid commands to be executed.

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -136,7 +136,9 @@
         })
       };
       persistentDialog(cm, queryDialog, q, searchNext, function(event, query) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as e47ab5f92d...26d8c7109aa

40thieves:
Great, thanks!

Issue #4163, opened 2016-08-13
Speciality JSON MIME-types Missing


xeenon:
There are over 50 "+json" MIME-types [registered with IANA](http://www.iana.org/assignments/media-types/media-types.xhtml). They should be recognized as JSON like the "+xml" MIME-types are treated as XML in `CodeMirror.resolveMode`.

```javascript
    } else if (typeof spec == "string" && /^[\w\-]+\/[\w\-]+\+xml$/.test(spec)) {
      return CodeMirror.resolveMode("application/xml");
    }
```

marijnh:
Fixed via #4164

PR #4164, opened 2016-08-13
Special case JSON based MIME-types.

xeenon:
Fixes issue #4163.


marijnh:
Good idea. Merged as 5738f9b

Issue #4171, opened 2016-08-17
[javascript hint] does not include non-enumerable properties


orionlee:
In the current implementation (post v5.17.0), javascript hint does not include non-enumerable properties.

### Impact: 
In practice, it means it does not include most top-level classes, at the minimal in recent Chrome environment. 

### How to reproduce:
- Go to autocomplete demo page:  http://codemirror.net/demo/complete.html
- in the top editor, try to auto-complete some top-level classes, e.g., Event, Promise, etc.
- it will not complete (tested on Chrome 51 and 52 on windows 7).

orionlee:
I have a patch in my repository by using Object.getOwnPropertyNames() , supported by most modern browsers. I could create a pull request with the patch. Let me know.

marijnh:
That'd be useful (but make sure to fall back to `for/in` when `getOwnPropertyNames` isn't available)

marijnh:
Fixed via #4175

PR #4175, opened 2016-08-18
[javascript hint] autocomplete non-enumerable properties, e.g., Promise

orionlee:
@marijnh  - The initial cut for #4171 . Please review. Thanks.

- It'd be nice to add a test to test sutie as the difference is subtle. Thoughts? If so, I can follow the pattern of sql-hint-test.js, given there is no existing regression test for javascript-hint.


marijnh:
Thanks! Merged as 1c884ebabf6 and simplified in 5fea9c5737eaa

PR #4182, opened 2016-08-19
Add f to delimiter stripping for Python mode

takluyver:
So f-strings get terminated correctly.

This illustrates the problem I'm trying to fix.

screenshot from 2016-08-19 11-27-02


marijnh:
Thanks. Merged as c45674b

PR #4191, opened 2016-08-23
Introduce getBidiPartAtWithOther

adrianheine:
… instead of leaking internal information into a global and using it 6000 lines earlier. Helps me with introducing ES6 modules.

### DIFF
--- a/lib/codemirror.js
+++ b/lib/codemirror.js
@@ -2873,7 +2873,9 @@
     }
     var order = getOrder(lineObj), ch = pos.ch;
     if (!order) return get(ch);
-    var partPos = getBidiPartAt(order, ch);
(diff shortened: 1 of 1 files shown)


marijnh:
Leaking is such a dirty word. This was my way of returning multiple values with allocation. In ProseMirror, I've settled on a pattern for these cases where I statically allocate a single object, and then reuse that (assigning to it and then returning) for all returns for a given function. That looks better, and still doesn't allocate. It does require you to immediately destruct the result to make sure it's not overwritten by a subsequent call. Would this pattern be palatable to you, or do you prefer to just allocate something?

adrianheine:
Actually, I think we should leave it as it is. With modules, it even gets a bit nicer.

PR #4195, opened 2016-08-24
Added CALL and YIELD as keywords to Cypher

pontusmelke:


marijnh:
Thanks. Merged as e865bde

Issue #4097, opened 2016-07-06
Bug with measuring the size of a character + caret misplaced


firasdib:
I've found this bug where the function `charCoords` is telling me the size of a character is the entire line length. And in combination with this, the caret might end up on the very far right of the editor even if there are no characters there. Please see the images attached. 

Note that this is with line-wrapping enabled and the `mode` for `charCoords` is `local`!

1. Editor as normal, typing no problemo. 
<img width="969" alt="screen shot 2016-07-06 at 10 56 20" src="https://cloud.githubusercontent.com/assets/1335165/16612763/e849c6d2-4368-11e6-9b8f-c26cf0c18416.png">

2. Remove all characters until you reach the `!`. Note the caret on the far right.
<img width="974" alt="screen shot 2016-07-06 at 10 56 30" src="https://cloud.githubusercontent.com/assets/1335165/16612777/fbbc8a42-4368-11e6-9e16-687167d4694a.png">

3. Note the odd size of the `!` character (start and end rects), which it claims spans from `0-950px` (ie the entire line)
<img width="695" alt="screen shot 2016-07-06 at 10 58 06" src="https://cloud.githubusercontent.com/assets/1335165/16612793/1ace0d2a-4369-11e6-9741-237af7bae32d.png">

Thanks for maintaining Codemirror!

firasdib:
You can test here: https://codemirror.net/demo/indentwrap.html

marijnh:
I think attached patch helps. Could you confirm?

firasdib:
Thanks, that fixed it @marijnh

Issue #237, opened 2011-11-08
Implement goal column


marijnh:
This should be rather easy with the new keymap system.

PR #4197, opened 2016-08-24
Fix space regression

danieleds:
In old versions of WebKit, with CodeMirror's wrap mode enabled, an odd behavior happens. That is, the first whitespace keypress in a sequence is not displayed until another space is added.

This is the commit that introduced the issue: https://github.com/codemirror/CodeMirror/commit/e60f4b7dd88b71e227f79200932c028ca047d78e

I have added a special case to the code to fallback to the old method in case an old version of WebKit is used.


marijnh:
My Chrome 52 matches as Webkit 537, so this also applies to recent versions of Chrome (which don't suffer from issue #237), and breaks the fix for #4097 there. I don't think that's acceptable.

(Also, if someone's userAgent string happens to contain `Webkit/` without a number following it for some reason, this'll crash the library on startup.)

PR #4200, opened 2016-08-25
Split into ES6 modules (Closes #747)

### LINKED ISSUE #747: Splitting CodeMirror into specific files
Could we make an effort to split CodeMirror's source intro specific files?

I've noticed the branch v3, and maybe there we can split it like ace's source, I can put the effort to fork and rewrite the code, if we can agree on how to split it and choose a code style guide. (also I could add grunt.js for build, min, etc)

Thoughts?

adrianheine:
This is work in progress and more hitting with a hammer than slicing with a knife. Specific questions:
  * Default exports or not? (See Pos)
  * Are the CodeMirror.* properties going to stay in the long run or should they be kept separately and eventually removed?

Tests pass for me except core_lineWidgetChanged and vim_zb_to_bottom, but they both fail for me on master, too.


marijnh:
I prefer not to use default exports (as they tend to drag one towards a single-export-only style which leads to awkward nonsense).

Which is of course the exact kind of awkward nonsense that CodeMirror's export structure is currently based on (everything hangs off a single global). I want to remain 100% compatible at least in the first release that includes this (so that we don't try to refactor and change APIs at the same time), but I think we can have the old interface as a 'bundling' module that just pulls all the various pieces out of other modules and puts them together into the old interface. I guess we'll eventually want to move to a more ES6-y interface. (Though even then we might want to keep the old behavior as a deprecated interface for a while, depending on how breaking we want the next major to be.)

adrianheine:
Ok, I'll remove the few default exports then. I'll also move the CodeMirror.* setting to a separate module. I would also try to come up with reasonable subdirectories. What do you think about the size of the files? They have 10 to 571 lines, a lot are around 200 lines.

marijnh:
That module size sounds good.

Which browser are those tests failing on? On master, I have one RTL test failing on Chrome 50, and everything passing on Firefox 46.

adrianheine:
Firefox 48.

I have a problem with a circular dependency. `startWorker` through `highlightWorker` (in `src/display/highlight_worker.js`) calls `runInOp` (in `src/operations.js`), which triggers `endOperation`, `endOperations`, `endOperation_W1`, which finally calls `updateDisplayIfNeeded` (in `src/display/update_display.js`), which might call `startWorker`. I don't know how and where to put these things.

adrianheine:
Also, phantomjs downloading fails.

package.json
```diff
@@ -6,12 +6,15 @@
     "license": "MIT",
     "directories": {"lib": "./lib"},
     "scripts": {
+      "prepublish": "npm run-script build",
+      "build": "rollup --banner \"`cat src/banner.js`\" --format umd -n CodeMirror src/codemirror.js -o lib/codemirror.js || true",
```
marijnh: Is the `|| true` temporary? The cruft npm adds when a script fails is pretty awful, but I think in this case we'd want to preserver the return value.

src/CodeMirror.js
```diff
@@ -0,0 +1,218 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
```
marijnh: Two files whose case-insensitive names are the same seems like a bad idea -- some fool will want to check this out on a FAT filesystem at some point. Maybe rename codemirror.js to index.js?

src/CodeMirror.js
```diff
@@ -0,0 +1,218 @@
...
+import { captureRightClick, ie, ie_version, mobile, webkit } from "./util/browser";
+import { bind, copyObj, Delayed } from "./util/misc";
+import { eventInWidget } from "./utils_widgets";
+import { e_preventDefault, e_stop } from "./util/event";
+import { clipPos } from "./utils_pos";
+import Display from "./Display";
+import Doc from "./Doc";
+
+var defaults;
+var optionHandlers;
```
marijnh: These seem like they should be imported from an options (I'd rename default_options to options, since it also holds the definition of `defineOption`) module, and only attached to `CodeMirror` in this file (along with `defineOption`)

src/CodeMirror.js
```diff
@@ -0,0 +1,218 @@
...
+
+// The default configuration options.
+defaults = CodeMirror.defaults = {};
+// Functions to run when options are changed.
+optionHandlers = CodeMirror.optionHandlers = {};
+
+export default CodeMirror;
+
+// Attach the necessary event handlers when initializing the editor
+function registerEventHandlers(cm) {
```
marijnh: Since this is rather random, fiddly DOM code, maybe move it to event/index.js and the other event code to event/mouse.js event/key.js etc?

src/ContentEditableInput.js
```diff
@@ -0,0 +1,469 @@
+import { replaceRange } from "./changes";
+import { contains, range, removeChildrenAndAdd, selectInput } from "./util/dom";
+import { on, signalDOMEvent } from "./util/event";
+import { applyTextInput, copyableRanges, disableBrowserMagic, handlePaste, hiddenTextarea, lastCopied, setLastCopied } from "./input";
+import { operation, runInOp } from "./operations";
+import Pos from "./Pos";
+import { cmp, maxPos, minPos } from "./Pos";
```
marijnh: And that (awkward double imports, or alternatively adding `default: Pos` in this one) is why I don't like default exports very much.

src/ContentEditableInput.js
```diff
@@ -0,0 +1,469 @@
...
+    this.div.contentEditable = String(val != "nocursor")
+  },
+
+  onContextMenu: nothing,
+  resetPosition: nothing,
+
+  needsContentAttribute: true
+  }, ContentEditableInput.prototype);
+
+function posToDOM(cm, pos) {
```
marijnh: I'd put all the scary DOM reading stuff that follows close to the display drawing code, rather than here (yeah, I know, it was here in the original file as well)

src/Display.js
```diff
@@ -0,0 +1,107 @@
+import { elt } from "./util/dom";
```
marijnh: (Having a display/ dir with all the display-related code is probably a good idea)

src/Doc.js
```diff
@@ -0,0 +1,443 @@
...
+  },
+  lineSeparator: function() { return this.lineSep || "\n"; }
+});
+
+// Public alias.
+Doc.prototype.eachLine = Doc.prototype.iter;
+
+// Used both to provide a JSON-safe object in .getHistory, and, when
+// detaching a document, to split the history in two
+function copyHistoryArray(events, newGroup, instantiateSel) {
```
marijnh: Seems like a good candidate for history.js

src/Doc.js
```diff
@@ -0,0 +1,443 @@
...
+          lst(newChanges)[prop] = change[prop];
+          delete change[prop];
+        }
+      }
+    }
+  }
+  return copy;
+}
+
+function findSharedMarkers(doc) {
```
marijnh: This one and the functions below might be more appropriate in mark_text.js

src/cm_utils.js
```diff
@@ -0,0 +1,64 @@
+import { addClass, elt, removeChildren, rmClass } from "./util/dom";
+import { on } from "./util/event";
+import { clearCaches } from "./measurement/position_measurement";
+import { setScrollLeft, setScrollTop } from "./scroll_events";
+import { scrollbarModel } from "./display/scrollbars";
+import { updateGutterSpace } from "./display/update_display";
+import { indexOf } from "./util/misc";
+
+export function initScrollbars(cm) {
```
marijnh: How about display/scrollbars.js for this one?

src/cm_utils.js
```diff
@@ -0,0 +1,64 @@
...
+
+export function themeChanged(cm) {
+  cm.display.wrapper.className = cm.display.wrapper.className.replace(/\s*cm-s-\S+/g, "") +
+    cm.options.theme.replace(/(^|\s)\s*/g, " cm-s-");
+  clearCaches(cm);
+}
+
+// Rebuild the gutter elements, ensure the margin to the left of the
+// code matches their width.
+export function updateGutters(cm) {
```
marijnh: Maybe a display/gutters.js for this one and the one below? That leaves only `themeChanged` in this file, which you could smuggle into some other relatively generic file to drop this module entirely.

src/default_options.js
```diff
@@ -0,0 +1,191 @@
...
+import { updateScrollbars } from "./display/scrollbars";
+import { updateSelection } from "./display/selection_draw";
+import { mobile, windows } from "./util/browser";
+import { findMaxLine } from "./line/spans";
+import { loadMode, resetModeState } from "./state";
+import { regChange } from "./display/view_tracking";
+
+export var Init = {toString: function(){return "CodeMirror.Init";}};
+
+export function setDefaultOptions(CodeMirror) {
```
marijnh: Why is this a function instead of the top-level code it used to be?

src/display/selection_draw.js
```diff
@@ -0,0 +1,139 @@
+import Pos from "../Pos";
```
marijnh: I think `display/selection.js` sufficiently suggests that this is about the way the selection is displayed (i.e. maybe leave off `_draw` in the filename)

src/document_data.js
```diff
@@ -0,0 +1,251 @@
+import { cleanUpLine, Line, updateLine } from "./line/line_data";
```
marijnh: I'd be tempted to move the chunk prototypes to `Doc.js` and the update logic to `changes.js`, and drop this module (since it mostly contains stuff that I wouldn't expect here, based on the filename)

src/editor_methods.js
```diff
@@ -0,0 +1,571 @@
...
+
+// The publicly visible API. Note that methodOp(f) means
+// 'wrap f in an operation, performed on its `this` parameter'.
+
+// This is not the complete set of editor methods. Most of the
+// methods defined on the Doc type are also injected into
+// CodeMirror.prototype, for backwards compatibility and
+// convenience.
+
+export default function(CodeMirror) {
```
marijnh: Why not directly export the prototype object?

src/keynames.js
```diff
@@ -0,0 +1,18 @@
...
+  3: "Enter", 8: "Backspace", 9: "Tab", 13: "Enter", 16: "Shift", 17: "Ctrl", 18: "Alt",
+  19: "Pause", 20: "CapsLock", 27: "Esc", 32: "Space", 33: "PageUp", 34: "PageDown", 35: "End",
+  36: "Home", 37: "Left", 38: "Up", 39: "Right", 40: "Down", 44: "PrintScrn", 45: "Insert",
+  46: "Delete", 59: ";", 61: "=", 91: "Mod", 92: "Mod", 93: "Mod",
+  106: "*", 107: "=", 109: "-", 110: ".", 111: "/", 127: "Delete",
+  173: "-", 186: ";", 187: "=", 188: ",", 189: "-", 190: ".", 191: "/", 192: "`", 219: "[", 220: "\\",
+  221: "]", 222: "'", 63232: "Up", 63233: "Down", 63234: "Left", 63235: "Right", 63272: "Delete",
+  63273: "Home", 63275: "End", 63276: "PageUp", 63277: "PageDown", 63302: "Insert"
+};
+(function() {
```
marijnh: We can unwrap the anon function here, since the module system will isolate this (and if we move to `let` it won't leak any bindings anyway)

src/line/line_data.js
```diff
@@ -0,0 +1,329 @@
...
+export function defaultSpecialCharPlaceholder(ch) {
+  var token = elt("span", "\u2022", "cm-invalidchar");
+  token.title = "\\u" + ch.charCodeAt(0).toString(16);
+  token.setAttribute("aria-label", token.title);
+  return token;
+}
+
+// Build up the DOM representation for a single token, and add it to
+// the line map. Takes care to render special characters separately.
+function buildToken(builder, text, style, startStyle, endStyle, title, css) {
```
marijnh: I'd prefer to the line/token-rendering to live under display/ somewhere.

src/line/line_data.js
```diff
@@ -0,0 +1,329 @@
...
+      style = interpretTokenStyle(styles[i++], builder.cm.options);
+    }
+  }
+}
+
+
+// These objects are used to represent the visible (currently drawn)
+// part of the document. A LineView may correspond to multiple
+// logical lines, if those are connected by collapsed ranges.
+export function LineView(doc, line, lineN) {
```
marijnh: Since the rest of the stuff related to views is in `display/view_tracking.js`, I'd expect this there as well

src/state.js
```diff
@@ -0,0 +1,21 @@
+import { startWorker } from "./display/highlight_worker";
```
marijnh: This one's name doesn't say much about its content. Maybe move it into a bigger utility module, or call it mode_state.js

marijnh:
Scary how big this system really is, when you scroll over all of it like this. I've added a few comments, I hope they make sense. I did succumb to code-reading-fatigue towards the end, but we can always adjust more stuff after merging this (I'll do another pass then).

package.json
adrianheine: Yeah, that was an accident.

src/CodeMirror.js
adrianheine: renamed codemirror.js to main.js and legacy.js.

src/ContentEditableInput.js
adrianheine: I already fixed that one locally :)

src/Display.js
adrianheine: There is, I just didn't move everything yet.

src/default_options.js
adrianheine: Actually, I didn't know how to just require with rollup. Also has the nice advantage of not depending on CodeMirror.js.

src/editor_methods.js
adrianheine: See above.

adrianheine:
It is big, indeed. Thanks for your comments, I will address them. Do you want me to squash the changes into one or should I leave them as they are? I also just rebased on current master so that travis works again.

src/CodeMirror.js
adrianheine: I think this is currently too entangled with CodeMirror logic and concepts to be considered util code.

src/ContentEditableInput.js
adrianheine: Good catch. I think that would currently fit in with measurement/, which I'm not too happy with.

src/Doc.js
adrianheine: I'd even say there should be a `History.prototype.copy` that wraps this.

src/display/selection_draw.js
adrianheine: I agree.

marijnh:
I like squashed commits, for things that fix/modify patches that haven't been landed yet.

src/line/line_data.js
adrianheine: `LineView` is used in `measurement/`, and the way I laid it out `measurement/` should not depend on `display/`.

src/line/line_data.js
adrianheine: See comment below. `line/` only exists for untangling dependencies between `display/` and `measurement/`.

src/cm_utils.js
adrianheine: Done.

src/line/line_data.js
marijnh: Fair enough, but note that LineView is also used for lazy display updating (see regChange and the way the line update code works)

adrianheine:
I think this is quite ok now. We have 7 directories, in order: edit, input, model, display, measurement, line, util. Each directory can only include from those after it. There's still `src/modes.js`, because I couldn't get myself to move it to util. I didn't address some of your comments yet, but I'm planning to do it. Issues:
*  I think `line` is a really bad name, it's more like primitive model stuff.
* The separation between edit and input is unclear to me
* model should be called document and include Doc, but Doc currently depends on CodeMirror.

marijnh:
I would prefer more emphasis on 'theme', and less on dependency order, in the organization. The code is, unfortunately, not written with a clean dependency graph in mind, and I don't think we have the time budget to do much about that right now. So for now I'd prefer for things to be easily findable by thinking about what they are related to, without constraining ourselves too much by what depends on them and what they depend on.

So something like `util` should really only have generic utilities, the `line` stuff can probably go under `model` -- i.e. don't let the fact that measurement uses those data structures prevent you from doing that -- and I think a directory `input` with all the event handling stuff would be a good addition. Does that make sense?

test/lint.js
```diff
@@ -8,4 +8,12 @@ var blint = require("blint");
   });
 });
 
+["src"].forEach(function(dir) {
+  blint.checkDir(dir, {
+    browser: true,
+    blob: "// CodeMirror, copyright (c) by Marijn Haverbeke and others\n// Distributed under an MIT license: http:\/\/codemirror.net\/LICENSE\n\n",
```
marijnh: (I'm happy to get rid of the per-file header thing. If that's in the way at any point, feel free to just drop it. I should probably have mentioned that earlier.)

adrianheine: I think it's okay that way. It's a separate commit, so we can easily take it out again, though.

adrianheine: Hm, the license header appears multiple times in the built file. That could be a reason for removing it.

marijnh: Let's just get rid of it. The more this becomes a group project the less comfortable I am with plastering my name all over it.

adrianheine: Ok, did that.

adrianheine:
This is getting rebase-needy. Maybe you want to merge it unless you have specific issues with it?

marijnh:
Did you address the stuff we discussed last week? (I'm not really re-reading the changeset every day, so I don't keep up with what happens unless you add a comment.) I'll work on landing this on Monday!

adrianheine:
I think I did everything we talked about.

marijnh:
Landed! (As is)

marijnh:
I wrote a little announcement post [here](https://discuss.codemirror.net/t/code-organization-overhaul/928)

PR #4203, opened 2016-08-26
Fix for modifiers on class members

### LINKED ISSUE #4188: TypeScript: broken with private, protected and public modifiers
ref: https://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers

image

```
class SomeClass {
    private prop1: string;
    protected prop2: string;
    public prop3: string;
}
```

DanielRosenwasser:
Fixes #4188.

Basically given that `static` doesn't use any lookahead to determine whether it should be treated as a modifier, I took the same approach here. Let me know if there's anything you'd prefer I fix up.

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -617,6 +617,12 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
   }
   function classBody(type, value) {
     if (type == "variable" || cx.style == "keyword") {
+      if (isTS) {
+        // Technically any of these modifiers could be standalone property names, but this is rare.
+        if (value == "public" || value == "private" || value == "protected" || value == "abstract") {
+          return cont(classBody);
+        }
+      }
       if (value == "static") {
         cx.marked = "keyword";
         return cont(classBody);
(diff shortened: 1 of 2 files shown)


mode/javascript/javascript.js
```diff
@@ -617,6 +617,12 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
   }
   function classBody(type, value) {
     if (type == "variable" || cx.style == "keyword") {
+      if (isTS) {
+        // Technically any of these modifiers could be standalone property names, but this is rare.
+        if (value == "public" || value == "private" || value == "protected" || value == "abstract") {
```
marijnh: Is a method like `public(...) {}` allowed in TS? (Similar to how you can do `get(...) {}` even though `get` is also the prefix for getters).

DanielRosenwasser: Yup, it's allowed and basically like that.

DanielRosenwasser: Whoops, I now see the logic you have below for that. But doesn't `static` have the same issue as well?

marijnh:
Thanks for looking into this. I've pushed an alternative patch 607d0e8

PR #4218, opened 2016-09-02
add string state to simplemode example definition

### LINKED ISSUE #4217: closebrackets incompatible with simplemode demo
I'm using the `edit/closebrackets` and `mode/simple` addons, with the (unmodified) "simplemode" [example definition](https://codemirror.net/demo/simplemode.html).

Closing brackets are automatically inserted as expected. But double-quote characters are either ignored entirely (maybe intended behaviour? #4052) or----more frustratingly----if they enclose an empty string, a _third_ double-quote character is auto-inserted after it.

(Edit: possible duplicate of #3660)

jc275:
- to play nicely with the closebrackets addon (closes #4217)
- introduces 'next' property a bit prematurely, but
- comment (/*) regex moved up, to mention 'next' next

### DIFF
--- a/demo/simplemode.html
+++ b/demo/simplemode.html
@@ -132,7 +132,9 @@ CodeMirror.defineSimpleMode("simplemode", {
   // The start state contains the rules that are intially used
   start: [
     // The regex matches the token, the token property contains the type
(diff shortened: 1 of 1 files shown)


marijnh:
I've pushed an alternative patch 45cf0ae

PR #4232, opened 2016-09-07
Mode for Application Cache files

orionlee:
Please consider incorporating the mode for .appcache files. Thanks.

### DIFF
--- a/mode/appcache/appcache.js
+++ b/mode/appcache/appcache.js
@@ -1,0 +1,43 @@
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"));
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror"], mod);
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
+"use strict";
+
+  CodeMirror.defineMode("appcache", function() {
+
+  return {
+    token: function(stream) {
+      var style;
+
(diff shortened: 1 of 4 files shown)


marijnh:
I'm not accepting new modes in the main distribution anymore. The recommended approach is to create your own repository and npm package to distribute your mode.

orionlee:
@marijnh If I keep the new mode separate, please advise on how to use the new mode along with auto load mode? 

Specifically, autoload mode requires all modes are stored in a single directory, e.g, `CodeMirror.modeURL = "node_modules/codemirror/mode/%N/%N.js";` . If the new mode is kept in a separate npm, it will be in a directory different from core codemirror.

Thanks.

marijnh:
Indeed, auto loading of modes isn't going to work anymore when modes aren't kept in a centralized place. Though you could of course set up your build script to copy your mode to the right directory.

orionlee:
How about changing auto loading addon slightly to allow user-defined modeURL resolution? It'd be something like this in `CodeMirror.requireMode()`, of [loadmode.js](https://github.com/codemirror/CodeMirror/blob/master/addon/mode/loadmode.js#L38)

```
  // accept optional user-defined function CodeMirror.resolveModeURL(modeName)
  var file = CodeMirror.resolveModeURL ? CodeMirror.resolveModeURL(mode) :  CodeMirror.modeURL.replace(/%N/g, mode);
```

Thanks.

marijnh:
I'm not really interested in going down the path of writing a parallel module loading system now that people tend to be using real module loaders in the browser. So the dynamic mode loader probably doesn't have much of a future, and is likely to be dropped in the next major release.

PR #4237, opened 2016-09-11
Localization of search dialogs

Frigory33:
Hi,
I'm using CodeMirror on French platforms, therefore I'm eager to have it easily translatable.
I added localization support (used for search) and extended the demo.
In my second commit, I roughly put subsections in the docs' Addons section, as it is currently very hard to browse.
Do you have any suggestions regarding these changes?


marijnh:
I don't want to incur more UI complexity in the core library. In fact, I'm thinking about moving stuff like the search functionality out of the distribution entirely, since I'm not really interested in maintaining them. As such, I'm not interested in growing them further, and recommend you set up a fork with your changes.

Frigory33:
OK, thanks for your reply. Yeah, let's follow the KISS rule. Then I'll propose my changes to this possible separate project.
Happy hacking.

PR #4240, opened 2016-09-13
Add passing the FocusEvent to blur and focus

reminyborg:
Just added the FocusEvent as a second parameter to blur and focus.
Left the first param for backwards compatibility.

Ready to make any changes that might be needed.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -626,10 +626,10 @@
       You can <code>preventDefault</code> the event, to signal that
       CodeMirror should do no further handling.</dd>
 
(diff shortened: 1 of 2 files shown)


marijnh:
What additional information does the event provide?

reminyborg:
Right now i am using it for [relatedTarget](https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget) to know what got focus on blur.

marijnh:
Makes sense. Merged as 3d89b71

PR #4245, opened 2016-09-19
Add mode and theme for Igor Pro

t-b:


marijnh:
Hi Thomas. We're not accepting new modes for the main CodeMirror distribution. The recommended way to distribute a mode is to set up your own package, call it something like codemirror-mode-igor, and publish it to NPM.

PR #4250, opened 2016-09-22
Add 'exec' keyword to mssql

luke-browning:


marijnh:
Thanks. Merged as 562e8ef

PR #4259, opened 2016-09-26
Remove semicolons

adrianheine:
Except for [two cases](https://github.com/marijnh/blint/issues/2#issuecomment-249652612) this is blint-clean. I used eslint's fix mode for the bulk work.


marijnh:
Nice! Merged as ac205e6. I released a new version of blint and that accepts the code fine.

PR #4260, opened 2016-09-26
Disallow semicolons in src/

adrianheine:
See #4259 and marijnh/blint#2. Does not pass for me.


marijnh:
Oh, I didn't realize you hadn't turned that on yet. Merged as 2182dfd79e92d and made passing in 5bb6a94ea4ed9 (as I mentioned in marijnh/blint#2 , I think it's preferable to put semicolons on such lines.

PR #4261, opened 2016-09-27
Replace var with let

adrianheine:
This also adds bublé to the rollup call. Build time with watch roughly doubles, so it's usually between 300 and 600 for me.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,3 +1,5 @@
+import buble from 'rollup-plugin-buble';
+
 export default {
   banner: `// CodeMirror, copyright (c) by Marijn Haverbeke and others
 // Distributed under an MIT license: http://codemirror.net/LICENSE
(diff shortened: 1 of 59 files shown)


rollup.config.js
```diff
@@ -0,0 +1,17 @@
+import buble from 'rollup-plugin-buble';
+
+export default {
+  banner: `\/\/ CodeMirror, copyright (c) by Marijn Haverbeke and others
```
marijnh: Why the escaped slashes?

adrianheine: I got myself intimidated by vim's apparently wrong syntax highlighting.

marijnh: Ah, right. The situation is no better in Emacs' js-mode, unfortunately.

marijnh:
So when I merge this the resulting codemirror.js fails on parsing in Chrome with

```
codemirror.js:4674 Uncaught SyntaxError: for-in loop variable declaration may not have an initializer.
```

Which appears to be a bug in Bublé that I myself introduced. I'll look into it.

marijnh:
There's already a Bublé patch at https://gitlab.com/Rich-Harris/buble/merge_requests/85/diffs . Until that gets merged and released, I've changed that single `let` back to `var` in 6992c53, and merged this.

PR #4267, opened 2016-09-29
[swift] Update to Swift3

ApolloZhu:
- Added missing keywords
- Removed non-swifty `atom`s
- Rearranged more logically


marijnh:
Thanks. Merged as 581854e

PR #4268, opened 2016-09-29
Convert anonymous to arrow functions

adrianheine:


kofifus:
Heu Adrian, Marjin .. as you're doing all this restructuring work, maybe have a quick look at my project here https://github.com/kofifus/New ... IMO a convenient way to create javascript 'classes' which is better than the standard 'class' way ... anyway great work and thx

marijnh:
Hi @kofifus, we're aiming for a plain, widely understood code style here. Also, your approach (which looks a lot like the thing Crockford has been advocating) will allocate a lot of unneccesary function objects and throw off the compiler's inlining behavior.

marijnh:
Merged as 9ad40ee76ff

(Feel free to merge in uncontroversial stuff yourself, by the way.)

adrianheine:
Thanks. I think I'll need a few more weeks before I allow myself to consider something uncontroversial.

PR #4273, opened 2016-09-30
Update highlight.js

hsk81:
Ensure that when an overlay is added that the class `cm-overlay` is used to denote a `span` instead of `cm-cm-overlay` (with the `cm-` prefix repeated twice).

* E.g. before fix:
```
<span class="cm-cm-overlay cm-spell-error">zpelling</span>
```

* And after fix:
```
<span class="cm-overlay cm-spell-error">zpelling</span>
```


marijnh:
Merged as 850671f (which also changes the other occurrence of the string)

PR #4275, opened 2016-10-01
[real-world uses] Add nteract

rgbkrk:
Adds [nteract](https://github.com/nteract/nteract) to the list of real world users! Thanks!


marijnh:
Thanks. Merged as 2701827

rgbkrk:
Thank _you_. CodeMirror is so excellent, and I'm ever so delighted the more I dig into the extension points available.

PR #4278, opened 2016-10-02
Don't sort Range's anchor and head arguments

adrianheine:
They are sorted by Range again anyway.

It would probably make sense to have non-sorting construction and retrieval options for Ranges.


marijnh:
> They are sorted by Range again anyway.

What do you mean by that? The `Range` constructor just assigns its arguments to `head` and `anchor`.

adrianheine:
> What do you mean by that? The Range constructor just assigns its arguments to head and anchor.

I was referring to the fact that `from` and `to` sort. I didn't expect so many places to directly access `head` and `anchor`. I suppose it makes sense to not loose this information here.

PR #4279, opened 2016-10-02
Document transposeChars, fix some inconsistencies

adrianheine:
Since it used `head`, the direction a selection was made would change which
chars get transposed. Also, a cursor at line start was not moved on successful
swapping.

### DIFF
--- a/src/edit/commands.js
+++ b/src/edit/commands.js
@@ -106,9 +106,17 @@ export let commands = {
     if (cm.somethingSelected()) cm.indentSelection("add")
     else cm.execCommand("insertTab")
   },
+  // Swap the two chars left and right of each selection's head.
+  // Move cursor behind the two swapped characters afterwards.
+  //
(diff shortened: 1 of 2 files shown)


marijnh:
Maybe make this only do anything for empty selection ranges? Whenever we do anything with a non-empty one, the difference between `head` and `anchor` _should_ be significant. The head is the acting side, and if this is to work on non-empty selections, I think using head is entirely sensible.

adrianheine:
I'm not familiar with such semantics for selections (i. e., directionality). The way transposing works on non-empty selections is broken, anyway, though, so maybe we should follow your suggestion.

marijnh:
Sublime Text's transpose doesn't do anything for non-empty selections either. So let's go with that. Do you want to amend your patch?

adrianheine:
Amended.

marijnh:
Wonderful. Merged as 6b3b6a6

Issue #4078, opened 2016-06-20
Text selection with lineWrapping: true not working as expected


nickbrowne:
I've had a few users report issues with text selection recently. After some investigation it seems related to the line wrapping functionality. It seems that long "words" without spaces wrapping onto multiple lines don't interact with the text selection as people expect.

After reproducing it in our own system, I was also able to reproduce it in `demo/indentwrap.html`.

In our own system, this most often happens with urls, as they tend to be the longest strings with no whitespace to interrupt them. I've also been able to reproduce the problem with a simpler case, which you can see demonstrated in this gif:

![cm5 15 2](https://cloud.githubusercontent.com/assets/2295892/16184844/d89d89e8-3702-11e6-9e34-b8dfd67a73d5.gif)

What I would expect to happen is to be able to select the last character at the end of the line without having to move the mouse to the start of the next line.

I've observed the bug in the following browser/os combinations:
 - Firefox 46.0.1, Linux Mint
 - Chrome 51.0.2704.63, Linux Mint
 - Chrome 51.0.2704.84, OSX

marijnh:
I've fixed this in a slightly awful way in attached patch. (There are plans to do something more systematic with cursor positions around line breaks, but not sure when the time/funding for that will happen.)

nickbrowne:
Great! I can confirm this fixes the issue. Thanks very much for taking care of it so quickly.

PR #4280, opened 2016-10-02
Add regression test for #4078

adrianheine:
If I comment out the kludge, this test fails again.


marijnh:
Thanks. Merged as a194325

PR #4281, opened 2016-10-02
[dracula theme] consistent selection styling regardless of focus

leeoniya:
This prevents selections from inheriting the default `#d9d9d9` background when editor is defocused.

reproducible @ https://codemirror.net/demo/theme.html#dracula

### DIFF
--- a/theme/dracula.css
+++ b/theme/dracula.css
@@ -16,7 +16,7 @@
 .cm-s-dracula .CodeMirror-gutters { color: #282a36; }
 .cm-s-dracula .CodeMirror-cursor { border-left: solid thin #f8f8f0; }
 .cm-s-dracula .CodeMirror-linenumber { color: #6D8A88; }
-.cm-s-dracula.CodeMirror-focused div.CodeMirror-selected { background: rgba(255, 255, 255, 0.10); }
+.cm-s-dracula .CodeMirror-selected { background: rgba(255, 255, 255, 0.10); }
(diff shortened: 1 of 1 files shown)


adrianheine:
I checked [dracula-theme](https://github.com/dracula/dracula-theme), and couldn't any more specific colors. Interestingly, the atom theme [seems to do the same](https://github.com/dracula/atom/blob/master/styles/base.less#L14), but I still think it's wrong.

leeoniya:
@adrianheine 

there actually seem to be a bunch of other themes that do this [1]. i think they're all wrong and probably a result of copy/paste/modify. additionally, all the `div.` prefixes should be dropped imo.

[1] https://github.com/codemirror/CodeMirror/search?utf8=%E2%9C%93&q=CodeMirror-focused

marijnh:
I guess the dark themes should definitely either provide their own unfocused selection color or apply their color regardless of focus, because the default color will look awful. For light ones, I don't care all that much.

leeoniya:
FWIW, even for light ones i dont understand why a defocused editor's selection would need to look different. Another thing i found in dracula (maybe others), is that `string` and `string2` share the same color [1], even though literal regexps in js mode get classed as `string2`. Perhaps this is by design of the theme, i dunno. I split them up locally, among other tweaks.

[1] https://github.com/codemirror/CodeMirror/blob/master/theme/dracula.css#L23

PR #4288, opened 2016-10-06
Add `mergeMarkerLocation` to options to allow customization

tberman:
This makes it easy to customize how the markers are added. We are using this to extend the highlight through the gutter.


marijnh:
This appears to be about the kind of line classed added for the changed chunks. So I don't think 'merge marker location' is a very good name. What exactly are you using it for?

tberman:
Yeah, the name is not optimal at all.

I am using it to customize the line classes that get added for merge-l/merge-r to include the gutter as well so we can style that. I made it also support an array to let people get complete customization as `wrap` and `gutter` + `background` do very different things, which could be useful depending on what you need.

What about `lineClassLocation` instead since the editor API is `[add|remove]LineClass`?

marijnh:
Thanks. I've merged this as 1502f805a86 and updated it in 9a316b33d724.

Issue #4284, opened 2016-10-04
Semantic meaning of applied token classes vary wildly between modes.


leeoniya:
I'm discovering while working with javascript & php that the parsers have surprising differences with how they categorize semantically similar tokens. This makes it impossible to rely on a single theme to deliver a uniform experience across multiple languages.

The javascript mode does more-or-less the correct thing. In php `$abc` is a basic variable, but is categorized as `cm-variable-2`, function arguments also get categorized as `cm-variable-2` rather than `cm-def`, `->abc` gets `cm-variable` but should be `cm-property`. javascript mode has no `cm-builtin` classification, for `parseInt`, `Math`, etc.

Is there any interest in adjusting the modes to provide better normalization? Would a pull request be considered for merging if it has to modify existing tests and cause style-related BC?

marijnh:
Improvements to the way the modes categorize things are welcome (such as making PHP properties actually use the property style). But langauges _are_ different, and as such, there's no single consistent way to tokenize them. The PHP mode uses `variable` for function names, and I don't think it's necessarily a bad idea to give $-prefixed variables a different style. The JS mode understands the difference between local and global variables, and uses `variable-2` for that. This is, I think, okay.

PR #4290, opened 2016-10-07
[javascript mode] add function & function-2 styles

leeoniya:
This implementation accounts for both, variables & properties:

```js
foo()
foo.call()
foo.apply()
foo.bar()
foo.bar.call()
```

etc..

This is not quite ready for merging, since it would require that all themes add

`.cm-function`
`.cm-function-2`

which *replace* (when needed)

`cm-variable`
`cm-variable-2`

While languages *are* different [1], functions exist in a great number of them (with similar calling conventions), and it seems appropriate that each CM mode is not required to overload an unpredictable `cm-variable-*` or other unused class [2] to tag them.

If this change is too breaking perhaps for modes that rely on the old parser classifications, what would be a good way to add *additional* `cm-function` (maybe even more specific `cm-method`) classes to these tokens? 

[1] https://github.com/codemirror/CodeMirror/issues/4284#issuecomment-251505455
[2] https://github.com/codemirror/CodeMirror/blob/master/lib/codemirror.css#L113-L133

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -272,10 +272,29 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
         while(cc.length && cc[cc.length - 1].lex)
           cc.pop()();
         if (cx.marked) return cx.marked;
-        if (type == "variable" && inScope(state, content)) return "variable-2";
+        if (type == "variable") {
+          var isFn = isTokenFunc(stream);
+
+          if (inScope(state, content))
+            return isFn ? "function-2" : "variable-2";
+
+          return isFn ? "function" : style;
+        }
         return style;
       }
     }
+  }
+
+  function isTokenFunc(stream) {
+    var isFn = false;
+    var ch = stream.peek();
+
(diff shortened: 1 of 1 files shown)


leeoniya:
i've added a commit that adds support for methods and functions in es6 objects and classes.

one point i'm not sure about is using the temporary flag `cx.isGetterSetter` on the context to discern getters and setters (which are not really functions) from callable methods. it seems a bit dirty to do this, but i don't see an obvious way to do this another way.

it may not be worth treating getters/setters as properties and they can just be treated the same as the rest.

there is possibly a bug in the first commit where `new Foo()` is treated as a function. which is not entirely wrong since it invokes the constructor. but `new Foo` is treated as `cm-variable`. This difference is not great. Again, some temporary `cx` flag can be used to remember the `new` so that `Foo` would never be marked as a function.

marijnh:
I'm not interested in this. See my comment [on the forum](https://discuss.codemirror.net/t/adding-more-detailed-classnames-to-tokens/951/2)

PR #4293, opened 2016-10-09
Update patterns

idleberg:
* commands are first in a line
* constants can contains more than `\w`, e.g. `${xml::CreateNode}`
* same goes for language strings


marijnh:
Thanks. Merged as 32d02f0

PR #4294, opened 2016-10-09
check to see if in a string before skipping a quote

### LINKED ISSUE #4287: Bug: when editing text, quote should only skip if it's the second of a matched pair.
I like the auto-insert of matching quotes, and I think it's the desired behaviour to have pressing quote skip over the second of a matching pair, but it's not good (also non-standard afaik, at least in emacs it doesn't do this) to have it for the first, it makes it hard to insert new elements in arrays of strings. To illustrate:

Desired:

    ['i want the cursor to skip over the quote when i'm done']
                                              cursor ->    ^
Undesired:

    ['<-- i am trying to insert a new string in my array to the left here']
    ^ <- cursor

themrmax:
Fixes #4287

### DIFF
--- a/addon/edit/closebrackets.js
+++ b/addon/edit/closebrackets.js
@@ -115,7 +115,7 @@
       var next = cm.getRange(cur, Pos(cur.line, cur.ch + 1));
       if (opening && !range.empty()) {
         curType = "surround";
-      } else if ((identical || !opening) && next == ch) {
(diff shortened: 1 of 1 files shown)


marijnh:
I initially tried fixing it like this, but figured it wasn't quite optimal for typing (different) quotes within a string, so I ended up with edd8e6180a520

themrmax:
cool thanks for this :)

PR #4299, opened 2016-10-11
[javascript mode] Fix tokenization for class expressions

Victorystick:
I noticed that syntax highlighting of method declarations in class expressions doesn't work. Class declarations work just fine though.

class-expression-bug

It's even more prevalent when using another theme, like Monokai:
class-expression-bug-2

This PR includes both tests and a fix.


marijnh:
Thank you. Merged as 6954ddd

PR #4301, opened 2016-10-11
[markdown] auto-close backticks

mtaran-google:


marijnh:
Looks good. Merged as 9903967

PR #4302, opened 2016-10-11
[go] auto-close backticks

mtaran-google:


marijnh:
Thanks. Merged as 358086a

PR #4309, opened 2016-10-13
add Stylus to meta.js

Rokt33r:
It should be included `mode/meta.js`.

https://codemirror.net/mode/stylus/index.html


marijnh:
Thanks. Merged as 09594e2

Issue #1725, opened 2013-08-04
Inconsistent tab vs space insertion on col 0


vincentwoo:
Right now CodeMirror behaves strangely around. If you go to http://codemirror.net/mode/python/index.html and type the following

```
def foo:
```

and hit enter, your cursor will be placed on line 2, and column 4. There will be 4 spaces between col 0 and your cursor. If you then type:

```
def foo:
    pass
```

and hit enter again, you will be placed on line 3, column 0. Hitting tab yet again here will insert a **hard tab** which is very strange and actually a bug when working with python. This problem appears to occur in other modes as well, and setting `insertWithTabs: false` doesn't appear to help either.

Is there a way I can disable hard tabs until this is fixed?

PR #4317, opened 2016-10-15
Insert soft tab when `indentWithTabs` is false

Rokt33r:
https://github.com/codemirror/CodeMirror/issues/1725

The default behavior is quite wield comparing modern editor.
I think inserting `Tab` character when using spaces as an indent is quite rare case.

Tab vs Spaces

### DIFF
--- a/src/edit/commands.js
+++ b/src/edit/commands.js
@@ -104,7 +104,13 @@ export let commands = {
   },
   defaultTab: cm => {
     if (cm.somethingSelected()) cm.indentSelection("add")
-    else cm.execCommand("insertTab")
+    else {
+      if (cm.getOption("indentWithTabs")) {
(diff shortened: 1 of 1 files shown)


marijnh:
Nope, the default behavior isn't going to change. It's easy enough to change these bindings if you want.

Rokt33r:
Okay. 🤕  I left my snippet for someone who want to tackle this.

```js
CodeMirror(el, {
  extraKeys: {
    Tab: function (cm) {
      if (cm.somethingSelected()) cm.indentSelection('add')
      else {
        if (cm.getOption('indentWithTabs')) {
          cm.execCommand('insertTab')
        } else {
          cm.execCommand('insertSoftTab')
        }
      }
    }
  }
})
```

PR #4318, opened 2016-10-15
No transposing char on OS X

Rokt33r:
Sublime Editor on OS X doesn't support transpose by itself because the OS supports it by default.
Beside, It is provided `Ctrl + T` not `Cmd + T`.

> `Cmd + T` executes `Go to anywhere` on OS X
http://docs.sublimetext.info/en/latest/reference/keyboard_shortcuts_osx.html#navigation-goto-anywhere

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -294,7 +294,7 @@
     });
   };
 
-  map[ctrl + "T"] = "transposeChars";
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as 0733ad5

PR #4321, opened 2016-10-17
Add jsnext:main entry to package.json

Thristhart:
https://github.com/rollup/rollup/wiki/jsnext:main

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "codemirror",
   "version": "5.19.1",
   "main": "lib/codemirror.js",
+  "jsnext:main": "src/codemirror.js",
   "description": "Full-featured in-browser code editor",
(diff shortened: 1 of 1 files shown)


adrianheine:
We are using `let` and arrow functions, possibly newer ES features like default parameters. The wiki page you linked to recommends to use `jsnext:main` with a file that would normally work in target engines except for the usage of modules. That's not true in our case.

Thristhart:
Ah, right. It works in my use case because I transpile afterwards. Do you think it's worth it to set up a transpile step that keeps the module loading? I'm not really sure what that would look like.

marijnh:
Rollup is great, but I've been holding off on adding jsnext:main fields to my libraries until something more well-defined and widely agreed on comes along.

PR #4322, opened 2016-10-17
Update cypher mode to handle unbalanced strings

pe4cey:


marijnh:
Thanks. Merged as bda9879

PR #4336, opened 2016-10-20
CodeMirror.version is behind actual release number

vincentwoo:


marijnh:
That was a mistake in the release script. I've released a 5.20.2 with a correct version number.

PR #4338, opened 2016-10-21
Add classes to each pane for isolated style changes, handle document swapping

tberman:
Fixes a bug where a document swap wouldn't cause the diff to be recalculated.

Also adds unique class names to each editor pane.


marijnh:
Thanks for the patch! Merged as 63bf659

PR #4339, opened 2016-10-21
Coordinate DOM updates across multiple CodeMirror instances.

sedwards2009:
My application Extraterm https://github.com/sedwards2009/extraterm uses CodeMirror a lot. "A lot" meaning that typical terminal tab may contain 10s to 100s of CM instances interleaved with other content. This works quite well except for things which affect the whole terminal tab and scrollback. For example, resizing and refreshing CSS styles. I'm looking for ways to speed up these kinds of bulk operations across multiple CM instances.

This pull request adds a `cm.coordinatedOperation()` method which allows for the coordinated update of the DOM across multiple CodeMirror instances. When CM finishes an operation updates the DOM, it will try to group DOM reads and writes into batches so as to avoid DOM thrashing. (I'm referring to the `endOperation_R1`, `endOperation_W1`, `endOperation_R2`, `endOperation_W2` functions.) `coordinatedOperation()` lets the caller operate on multiple CM instances and then execute all of the instances' `endOperation_R1` functions, then all of the `endOperation_W1` functions etc etc. This groups the update operations.

This chrome time line below shows the performance differences when using 100 CM instances. Normal `cm.refresh()` vs `cm.refresh()` with `coordinatedOperation()`, and also `cm.setSize(null,null)` vs the same using `coordinatedOperation()`.

codemirror_coordinated_operation_stats

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -2000,6 +2000,22 @@ editor.setOption("extraKeys", {
       lot faster. The return value from this method will be the return
       value of your function.</dd>
 
+      <dt id="coordinatedOperation"><code><strong>cm.coordinatedOperation</strong>(func: () → any)  → () → boolean</code><dt>
+      <dd>This is a variant of <a href="#operation"><code>operation</code></a>
+      which allows for the coordinated update of the DOM accross multiple
+      CodeMirror instances. This can be a significant speed up when many
+      instances are involved. It is intended to used as follows:
+        <ol>
+          <li>Call <code>coordinatedOperation</code> on each CodeMirror
+            instance and collect the function it returns into a list.</li>
+          <li>Execute each function in the list one at a time.</li>
+          <li>Execute each function in the list again.</li>
+          <li>Repeatedly execute each function in the list until all of
+            them return <code>true</code> to indicate that they are have
+            completed.</li>
+        </ol> 
+      </dd>  
+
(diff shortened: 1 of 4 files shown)


marijnh:
CodeMirror operations already have this property -- if you create an operation (in any editor) and then, _within_ that operation, touch any number of other editors, their updates will all be batched together when the outer operation ends.

sedwards2009:
The problem I have with that is that I may potentially go hundreds deep in the call stack.

marijnh:
I don't understand the deep call stack problem, could you elaborate?

sedwards2009:
(Correct me if I'm wrong.)  To refresh() (for example) multiple CM instances and group the DOM updates you need to effectively do:

    cm1.operation( () => {
        cm1.refresh();

        cm2.operation( () => {
          cm2.refresh();

          cm3.operation( () => {
            cm3.refresh();
            ...// now cm4, cm5 etc
          });
        });
    });

or is it just:

    cm1.operation( () => {
        cm1.refresh();
        cm2.refresh();
        cm3.refresh();
    });

marijnh:
No, operations are created implicitly for any method call that requires them, so you only need one wrapping `operation` call, and anything you do inside of that, to any number of instances, will be grouped.

sedwards2009:
So, that sounds like my second example above.

I'll try that out. Thanks for your time.

PR #4343, opened 2016-10-23
Remove the end class when removing line classes

tberman:
Minor typo-induced bugfix.


marijnh:
Argh, that was a dumb mistake. Thanks for fixing it. Merged as db12d64

PR #4344, opened 2016-10-23
Add yarn.lock

idleberg:
Since many people will start using [yarn](https://github.com/yarnpkg/yarn) to install Node dependencies…

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 .idea
 *.iml
 /lib/codemirror.js
+yarn.lock


marijnh:
I don't think adding a lock file to a library is a good idea. In which case would this be useful? The end apps will want to lock down their dependencies. Libraries should leave them flexible in order to make it less likely that multiple versions of a dependency have to be installed.

idleberg:
You're right, the makers of Yarn [actually recommend](https://yarnpkg.com/en/docs/yarn-lock#toc-check-into-source-control) checking the lock file into source control

PR #4347, opened 2016-10-24
TL-Verilog mode cleanup/rewrite.

stevehoover:
This push request is cleanup/rewrite of the Verilog mode support for Transaction-Level Verilog, currently with a limited user base (edaplayground.com and makerchip.com).  The only code change to the main Verilog mode is a check for the return value of the verilog token(..) hook, which now has an optional return value of 'style'.  TL-Verilog mode is largely rewritten to be cleaner and more maintainable with additional styling and a more comments.

### DIFF
--- a/mode/verilog/verilog.js
+++ b/mode/verilog/verilog.js
@@ -438,7 +438,8 @@ CodeMirror.defineMode("verilog", function(config, parserConfig) {
   var tlvTrackStatements = false;
   var tlvIdentMatch = /^([~!@#\$%\^&\*-\+=\?\/\\\|'"<>]+)([\d\w_]*)/;  // Matches an identifiere.
   // Note that ':' is excluded, because of it's use in [:].
-  var tlvLinePrefixMatch = /^[! ]  /;
+  var tlvFirstLevelIndentMatch = /^[! ]  /;
+  var tlvLineIndentationMatch = /^[! ] */;
   var tlvCommentMatch = /^\/[\/\*]/;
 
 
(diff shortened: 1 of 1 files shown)


marijnh:
@0b10011 @dwikle @ratchup Want to review this (or at least take a look whether it breaks anything obvious)?

ratchup:
This is not a review, but just a few comments...

This code revision much more faithfully models the TLV syntax and semantics
than the current release does. So, the revision is a major improvement over
the current release.

Some code comments:

---line 306
    if (hooks.token) {
        // Call hook, with an optional return value of a style to override
verilog styling.
        var style = hooks.token(stream, state);
        if (style !== undefined) {
          return style;
        }
      }

The return  model for the verilog mime token hook function style-not-found
case is now "undefined". So future verilog token hook functions, if any,
need to follow this convention. (Impact: guess it is small)




---line 489
         // Process TLV line prefix -- correct indentation.
          if (state.tlvCodeActive && (state.indented == 0) &&
stream.match(tlvLinePrefixMatch)) {
            stream.match(/ */);
            state.indented = stream.pos;
            stream.pos = 0;
          }

Direct update of parent internal state (stream.pos). (Impact:
maintainability of StringStream "class". Regression test desirable?)



On Fri, Oct 28, 2016 at 11:49 AM, Marijn Haverbeke <notifications@github.com

marijnh:
> Direct update of parent internal state (stream.pos). (Impact:
> maintainability of StringStream "class". Regression test desirable?)

It is probably possible to avoid this kludge by passing `false` as second argument to `stream.match`. I'd prefer that approach.

marijnh:
Great. Apart from the fact that you didn't balance your parentheses, it looks good now. Merged (with balanced parens) as d34e947

PR #4352, opened 2016-10-26
Adding text/yaml to the mime media types list

BigBlueHat:
Sadly, YAML doesn't (yet?) have a [registered media type](http://www.iana.org/assignments/media-types/media-types.xhtml).

Consequently, sometimes libraries (such as [mime-db](https://github.com/jshttp/mime-db/blob/9dd00b34556a8cdd6f3385f09d4989298c4b86e1/src/custom-types.json#L652-L654)) use `text/yaml` rather than `text/x-yaml`.

This solves for that.

Thanks!
🎩


marijnh:
Thanks. Merged as 0164f02

PR #4354, opened 2016-10-27
Fix Vue.js Single File Component Media Type

BigBlueHat:
Demo had the correct `text/x-vue`
defineMIME was using `script/x-vue`

There is, however, no `script/*` top-level media type:
http://www.iana.org/assignments/media-types/media-types.xhtml

Left, old `script/x-vue` for backwards compatibility.


marijnh:
Thanks! Merged as 2eb94e5

PR #4357, opened 2016-10-28
[mscgen mode] adds language constants

sverweij:
... and adds the xù specific keyword 'xu'


marijnh:
Thanks. Merged as c4d9363

PR #4358, opened 2016-10-29
Fix constructor typo for MergeView.

adriweb:


marijnh:
Thanks for spotting that. Merged.

PR #4365, opened 2016-11-02
Draw rulers all the way when scrollPastEnd is on.

alur:
Previously the "bottom: -20px" style placed the bottom of the rulers 20px into the bottom padding added by the scrollpastend addon, but the min-height on the CodeMirror-rulers div made them move around based on which lines were rendered.


marijnh:
Thanks. Merged as 2466392

PR #4372, opened 2016-11-08
Move setGutterMarker, clearGutter and lineInfo to Doc

### LINKED ISSUE #4369: cm.setGutterMarker and cm.clearGutter seem like doc methods, not cm methods
I can move them myself if the patch would be accepted, opening the issue to get a 👍 before doing it and to check to see if I am missing something that keeps them on the editor.

tberman:
Fix #4369 by moving these methods over to `Doc`


marijnh:
Thank you! Merged as 0e54532

PR #4374, opened 2016-11-08
Swift mode enhancements and fixes

TheSoundDefense:
A number of changes and fixes have been compressed into one pull request. Here is the full list of changes:

- `for` has been added to the list of defining keywords, as it defines a variable to be used within the `for`'s closure.
- The following types have been added: `Character`, `Int8`, `Int16`, `Int32`, `Int64`, `Set`, `UInt8`, `UInt16`, `UInt32`, `UInt64`.
- The following operators have been added: `~` (bitwise NOT), `^` (bitwise XOR), `?` (ternary operator and optional specifier), and `!` (forced unwrapping of optionals).
- `:` has been added as punctuation, as described by Apple in the Swift Language Reference (https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html).
- The regex for detecting numbers has been split into the four base types (binary, octal, decimal and hexadecimal), which have been rewritten to conform to the official grammar.
- The `property` type has now been split into the `property` type (for object or class properties such as `.count`), the `builtin` type (for compiler instructions and other literals such as `#file`), and the `attribute` type (for declaration or type attributes such as `@objc`). The words "property" and "attribute" are used by Apple in their language reference, which is convenient, and all of the possible `#file`-style literals are specified in their entirety in the grammar, making them effectively built-in. So it works out well.
- The rules for identifiers, properties, builtins and attributes were rewritten based on the grammar.
- JavaScript-style regex objects are no longer identified, as they are not currently supported in Swift, and continuing to highlight them resulted in certain bugs popping up. The regex for detecting them was left in, as this may become supported in the future.
- Attributes, builtins, numbers and properties are matched before operators and punctuation. This fixes a bug where properties and certain numbers were never identified as such.
- Types and atoms are now matched before standard keywords. This fixes a bug with the `Set` type.
- Closing brackets have been specified.
- Code can now fold.

A test file has also been included covering base cases and a number of edge cases. There's still some work to be done here, particularly with defining multiple variables in a single statement, but this update should solidify a lot of the Swift rules as they are specified in the grammar. Swift-specific hinting and linting functions are also written, but they depend on the changes that have been made here and need a bit of reworking to conform to last-minute changes, so those will come in their own pull requests.

The grammar used for reference was provided by Apple and can be found here: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/zzSummaryOfTheGrammar.html#//apple_ref/doc/uid/TP40014097-CH38-ID458 Subpages under "Language Reference" go into more detail.


marijnh:
Thanks. Merged as 4a1ed9134137 and fixed in 18c1bcb556bd1e (next time, make sure to actually _run_ the tests before submitting a PR)

TheSoundDefense:
Wow, I'm not sure what happened. I was running the tests constantly, I must have missed them after I made some last-minute changes. Sorry!

PR #4375, opened 2016-11-09
[vim mode] Add keymap to indent

hokaccha:
I added keymap to indent in insert mode. this behavior is vim compatible.
http://vimdoc.sourceforge.net/htmldoc/insert.html#i_CTRL-T

demo:

cm


marijnh:
Were you able to make this actually work on any browser? Neither Chrome nor Firefox seems to allow page scripts to handle ctrl-t.

hokaccha:
@marijnh It worked on the Mac, but it did not work on the window. thanks.

marijnh:
Ah, that makes sense. That's not necessarily a problem -- having this work on Mac is better than not having it work at all. I'll try to test on OS X soon.

marijnh:
Merged as 333a1f2

PR #4383, opened 2016-11-13
[julia mode] Fix string tokenizer

### LINKED ISSUE #4381: Julia: end of command literals not recognized
In Julia mode, the backtick closing a [command literal](http://docs.julialang.org/en/release-0.5/manual/running-external-programs/) seems not to be recognized as such; instead, the rest of the text is treated like a continuation of the literal, like a string:

```
readchomp(`pwd`) * "/bla"
    dostuff()
```

Furthermore, since the closing `)` is ignored, the second line here is automatically indented.

In jupyter, where I discovered this, the next line is at least highlighted normally, although the `)` is still not recognized; in the [web showcase](https://codemirror.net/mode/julia/index.html), the whole rest becomes red.

pabloferz:
Fix #4381

### DIFF
--- a/mode/julia/julia.js
+++ b/mode/julia/julia.js
@@ -338,23 +338,20 @@ CodeMirror.defineMode("julia", function(config, parserConf) {
   }
 
   function tokenStringFactory(delimiter) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Merged as e6eebeb

PR #4389, opened 2016-11-14
[SQL] Remove non-strict useless comparison for booleans

haskellcamargo:
This pull-request removes the ocurrences of `== true`. This operation is useless in boolean contexts and only adds readability issues.


marijnh:
Thank you. Merged as 532ae31

PR #4390, opened 2016-11-14
Update CSS mode to support user-select.

takamori:
As described in http://caniuse.com/#feat=user-select-none and https://developer.mozilla.org/en-US/docs/Web/CSS/user-select


marijnh:
Thanks. Merged as beb8382

Issue #4392, opened 2016-11-16
document.activeElement throws in Edge when viewed in an iframe.


jdalton:
I haven't narrowed down a simple vanilla JS repro yet, but using something the [react-native-web-player](https://jsbin.com/kuvabar/), which uses codemirror, in an iframe will error in MS Edge.

It errors [in the method which loops over `e.root.activeElement`](https://github.com/codemirror/CodeMirror/blob/6992c534ad33ce60e7500fe04a99a481ef8f548e/src/util/dom.js#L61-L66). It seems initializing with `document.activeElement` throws an `Unspecified Error` much like the old IE9 behavior. Wrapping it in a try-catch and returning `document.body` when it throws seems to avoid the issue.

Related to https://github.com/dabbott/react-native-web-player/issues/16.

jdalton:
Moved to #4393.

PR #4393, opened 2016-11-16
Avoid “Unspecified Error” in MS Edge when accessing `document.activeE…

jdalton:
This PR [avoids a tricky “Unspecified Error”](https://github.com/codemirror/CodeMirror/pull/4392)  in MS Edge 14 when accessing
`document.activeElement` from inside an iframe.

### DIFF
--- a/src/util/dom.js
+++ b/src/util/dom.js
@@ -1,4 +1,4 @@
-import { ie, ie_version, ios } from "./browser"
+import { ie, ios } from "./browser"
 
 export function classTest(cls) { return new RegExp("(^|\\s)" + cls + "(?:$|\\s)\\s*") }
 
@@ -58,17 +58,19 @@ export function contains(parent, child) {
   } while (child = child.parentNode)
 }
 
-export let activeElt = function() {
-  let activeElement = document.activeElement
+export function activeElt() {
+  // IE and Edge may throw an "Unspecified Error" when accessing document.activeElement.
+  // IE < 10 will throw when accessed while the page is loading or in an iframe.
+  // IE > 9 and Edge will throw when accessed in an iframe if document.body is unavailable.
+  let activeElement
+  try {
+    activeElement = document.activeElement
+  } catch(e) {
+    activeElement = document.body || null
+  }
   while (activeElement && activeElement.root && activeElement.root.activeElement)
     activeElement = activeElement.root.activeElement
   return activeElement
-}
-// Older versions of IE throws unspecified error when touching
-// document.activeElement in some cases (during loading, in iframe)
-if (ie && ie_version < 11) activeElt = function() {
-  try { return document.activeElement }
-  catch(e) { return document.body }
 }
 
 export function addClass(node, cls) {


src/util/dom.js
```diff
@@ -58,17 +58,15 @@ export function contains(parent, child) {
   } while (child = child.parentNode)
 }
 
-export let activeElt = function() {
```
jdalton: I removed the function forking because it seemed like overkill.

src/util/dom.js
```diff
@@ -58,17 +58,15 @@ export function contains(parent, child) {
...
-// Older versions of IE throws unspecified error when touching
-// document.activeElement in some cases (during loading, in iframe)
-if (ie && ie_version < 11) activeElt = function() {
-  try { return document.activeElement }
-  catch(e) { return document.body }
+export function activeElt() {
+  // Older versions of IE and sometimes Edge throws unspecified error when touching
+  // document.activeElement in some cases (during loading, in iframe)
+  try {
+    let activeElement = document.activeElement
```
marijnh: Does it still work when you only wrap the first access to `document.activeElement` in a `try`? Non-trivial code wrapped in error-swallowing try/catch scares me, since when someone introduces a bug there, we probably won't notice.

jdalton: I'll check it out. What's interesting is the old fallback for IE9 didn't have the loop logic.

jdalton: I'm triaging the bug from the MS Edge side and trying to narrow down the `Unspecified Error` which is a fallback error for things just went totally wonky.

jdalton: @marijnh Sorry for more q's but when is there a `root` property on an element. Is this something CodeMirror adds?

marijnh: The loop is to 'break out' of shadow DOMs (which old IE's don't support, hence it was missing from the other implementation)

jdalton: Updated!

jdalton:
@marijnh Do you have any other insight related to [my previous comment](https://github.com/codemirror/CodeMirror/pull/4393#discussion_r88304719)?

jdalton:
@marijnh Ok I found the issue. By spec `document.activeElement` will be `document.body` and if `document.body` isn't available it should be `document.documentElement` or `null`. However, in Edge in an iframe if `document.body` isn't available it will throw `Unspecified Error` and out of an iframe it will return `null`.

jdalton:
Oh snao it turns out it will do that in Edge, IE 11, IE 10, IE9, IE8... I've updated the code comment to describe the issue better.

marijnh:
Thanks! Merged as 69669e4.

>  // IE < 10 will throw when accessed while the page is loading or in an iframe.
>  // IE > 9 and Edge will throw when accessed in an iframe if document.body is unavailable.

Could it be that these are actually the same case? It was observed 'during loading' before, but most likely that was just because no body was available.

jdalton:
> Could it be that these are actually the same case? It was observed 'during loading' before, but most likely that was just because no body was available.

They could be similar.
I wasn't sure if IE < 10 may also error out of an iframe while the page is loaded so I left that bit in.

PR #4400, opened 2016-11-22
Fix the `combine` option for overlay modes inside blankLines

tberman:
Before, the base methods were being called but nothing was being returned in either case. This should fix it both for normal and combine overlays.

### DIFF
--- a/addon/mode/overlay.js
+++ b/addon/mode/overlay.js
@@ -76,14 +76,14 @@ CodeMirror.overlayMode = function(base, overlay, combine) {
     innerMode: function(state) { return {state: state.base, mode: base}; },
 
     blankLine: function(state) {
-      var baseToken, overlayToken
+      var baseToken, overlayToken;
       if (base.blankLine) baseToken = base.blankLine(state.base);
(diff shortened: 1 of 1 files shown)


addon/mode/overlay.js
```diff
@@ -76,8 +76,15 @@ CodeMirror.overlayMode = function(base, overlay, combine) {
...
 
     blankLine: function(state) {
-      if (base.blankLine) base.blankLine(state.base);
-      if (overlay.blankLine) overlay.blankLine(state.overlay);
+      var baseToken, overlayToken;
+      if (base.blankLine) baseToken = base.blankLine(state.base);
+      if (overlay.blankLine) overlayToken = overlay.blankLine(state.overlay);
+
+      if (combine) {
+        return baseToken + " " + overlayToken;
```
marijnh: This looks like it will often append undefined/null to a string.

tberman: Good catch. Will update when I get back to my computer.

tberman:
Alright, @marijnh, should be fixed.

marijnh:
Thanks. Merged as d0cde7f8470d66

PR #4402, opened 2016-11-23
Correct bidi types for some chars

adrianheine:


marijnh:
Thanks! I originally got these from pdf.js project, and never looked at them very closely.

adrianheine:
I see: mozilla/pdf.js#7844. Bugs like these really trip me up with the visual cursor movement :)

PR #4403, opened 2016-11-24
Correct bidi types for remaining Arabic chars

adrianheine:
As pointed out in mozilla/pdf.js#7844 by @timvandermeij. Maybe we should pull this into a shared module?


marijnh:
Merged. Now that we have a build system, relying on NPM modules would be workable, so a shared module could work.

PR #4413, opened 2016-11-30
[Behavior change] Highlight active line even when text is selected

tklancer:
Currently, when a line is highlighted and the user selects text, the highlight goes away. This PR would change that behavior by keeping the highlight unless the text selection extends outside of the current line.

The current behavior was introduced several years ago: https://github.com/codemirror/CodeMirror/commit/fac94560d2a1fb26902ea74df8de0fefc87d2457#diff-081cd675c6a483cb58a765919bdab1e9

If it makes more sense to have the behavior controlled by a configuration option, let me know and we can implement that as well.

### DIFF
--- a/addon/selection/active-line.js
+++ b/addon/selection/active-line.js
@@ -3,14 +3,13 @@
 
 // Because sometimes you need to style the cursor's line.
 //
-// Adds two options:
 // 'styleActiveLine': when enabled, gives the active line's wrapping
 // <div> the CSS class "CodeMirror-activeline", and gives its background
 // <div> the class "CodeMirror-activeline-background".
 //
-// 'styleActiveSelected': When enabled and 'styleActiveLine' is
-// enabled, keeps the active line's styling active even when text is
-// selected within the line. Has no effect if 'styleActiveLine' is not enabled
+// 'styleActiveSelected': An optional parameter of 'styleActiveLine'.
+// When enabled, keeps the active line's styling active even when text is
+// selected within the line. Has no effect if 'styleActiveLine' is not enabled.
 
 (function(mod) {
   if (typeof exports == "object" && typeof module == "object") // CommonJS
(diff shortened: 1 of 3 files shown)


marijnh:
Hi. Yes, definitely make it a default-off option. We don't want to break existing behavior unless there's a really good reason to.

tklancer:
@marijnh Thanks for the quick response! We've updated to add a new property 'styleActiveSelected', along with documentation and a button in the demo. Let me know what you think.

mdeng123:
@marijnh I just wanted to quickly follow up and get your thoughts on this pull request. I am working with @tklancer on this.

addon/selection/active-line.js
```diff
@@ -3,9 +3,14 @@
...
-// active line's wrapping <div> the CSS class "CodeMirror-activeline",
-// and gives its background <div> the class "CodeMirror-activeline-background".
+// Adds two options:
+// 'styleActiveLine': when enabled, gives the active line's wrapping
+// <div> the CSS class "CodeMirror-activeline", and gives its background
+// <div> the class "CodeMirror-activeline-background".
+//
+// 'styleActiveSelected': When enabled and 'styleActiveLine' is
+// enabled, keeps the active line's styling active even when text is
+// selected within the line. Has no effect if 'styleActiveLine' is not enabled
```
marijnh: I prefer that you make `styleActiveLine` optionally take an options object as value, instead of a boolean, and read a `nonEmpty` option from there when it is given. That's how other addons tend to do it.

addon/selection/active-line.js
```diff
@@ -33,6 +38,15 @@
     }
   });
 
+  CodeMirror.defineOption("styleActiveSelected", false, function(cm, val, old) {
+    var prev = old && old != CodeMirror.Init;
+    if (val && !prev) {
+      cm.state.maintainHighlight = true;
```
marijnh: You don't need to store anything in the state. You can just read the options when you need them using `getOption`

marijnh:
Added some more comments!

marijnh:
Merged as 41804498097a446 and cleaned up in 33d0057f1edb (renamed the option to `nonEmpty`, made sure updating the editor option works, clarified docs)

tklancer:
Thanks very much, @marijnh ! 💯

PR #4421, opened 2016-12-07
[emacs keymap] export kill, killRegion, repeated

achengs:
so other potential emacs-type modules can use


marijnh:
Thanks. Merged as 5e342f2

PR #4426, opened 2016-12-10
improved markdown image lookaheads

callodacity:
Just improved the recognition of images in the markdown mode.

If the url doesn't have correct syntax in a link `[link](url)` the styles aren't applied. This isn't true for image links `image` where if it's not finished it will continue to parse the rest of the document.

This pull request just ensures the image syntax acts like the link image syntax.

**Examples**:

Before editing, the lines under the unclosed image syntax are parsed as `url`.
before

After, it acts just like the link syntax.
after

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -436,7 +436,7 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
       return getType(state);
     }
 
-    if (ch === '[' && state.imageMarker) {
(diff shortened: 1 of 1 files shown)


callodacity:
Also, I'm happy to make a pull request that makes the markdown mode style image titles differently too. I thought I might keep that as a separate PR, since I'm not sure if that's something you want. 

A quick example would be like this:
title

Let me know and I'll submit another PR :smile:

marijnh:
Thank you. Merged as 35ec5ee

PR #4428, opened 2016-12-12
NASM (Assembly AT&T Intel x86 and x86_64 syntax) CodeMirror mode

converge:


marijnh:
Hi. Thanks, but we aren't accepting new modes anymore. The recommended way to publish a mode is to create a separate package, say `codemirror-mode-nasm`, and publish that to NPM.

PR #4429, opened 2016-12-12
Remove !important from CodeMirror-scroll css class

evrial:
Forget that !important exists. Breaks Jupyter notebook and can't be simply fixed.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -156,7 +156,7 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket {color: #f22;}
 }
 
 .CodeMirror-scroll {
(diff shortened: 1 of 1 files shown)


marijnh:
I am well aware what `!important` does, and in this case you really don't want to override this, since it'll break your editor. What did you want to do in Jupyter notebook that required changing this?

evrial:
Trivial — disconnecting mouse and inputs look like this: http://take.ms/6h5Kd
If you sure that it'll break editor, maybe fix in correct place.

marijnh:
If you can reduce this problem to a minimal HTML page, I'll look into it. If you're just going to be arrogant, you can go away.

PR #4434, opened 2016-12-14
Add will-change property and its values

ficristo:
I hope it is correct.


marijnh:
Thanks. Merged as c2a11a3

ficristo:
So fast! Thank you

PR #4435, opened 2016-12-14
Fix errors being thrown if an event emission results in a listener being synchronously removed

connor4312:
It would result in the array length changing and (from experience!) result a hard-to-track-down error.

### DIFF
--- a/src/util/operation_group.js
+++ b/src/util/operation_group.js
@@ -62,7 +62,7 @@ export function signalLater(emitter, type /*, values...*/) {
     setTimeout(fireOrphanDelayed, 0)
   }
   for (let i = 0; i < arr.length; ++i) {
-    const fn = arr[i];
+    const fn = arr[i]
     list.push(() => fn.apply(null, args))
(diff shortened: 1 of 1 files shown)


connor4312:
I apologize for introducing the dreaded scourge, the Semicolon, into this pristine codebase. I have exorcised it, lint should now pass.

MattJustMatt:
@marijnh apologize for the tag/pressure here, although this PR is a blocker for us. Would love to get a review when you get a moment!

marijnh:
Event handler arrays are not mutated anymore since d7b1370ca45d74 (5.21.0), so I am not sure what is going wrong here. Can you submit a test case that fails?

connor4312:
I may have been on 5.20.x still, I'll check when I get into the office

marijnh:
Can I close this?

connor4312:
Yep, sorry for not getting back to you. This appears to be fixed in recent versions of CodeMirror.

PR #4442, opened 2016-12-16
Convert some classes to ES6

adrianheine:
Needs `buble@0.15.1`, but `buble-rollup-plugin@0.15.0` can pull that in.

### DIFF
--- a/src/display/scrollbars.js
+++ b/src/display/scrollbars.js
@@ -3,7 +3,7 @@ import { on } from "../util/event"
 import { scrollGap, paddingVert } from "../measurement/position_measurement"
 import { ie, ie_version, mac, mac_geMountainLion } from "../util/browser"
 import { updateHeightsInViewport } from "./update_lines"
-import { copyObj, Delayed } from "../util/misc"
+import { Delayed } from "../util/misc"
 
 import { setScrollLeft, setScrollTop } from "./scroll_events"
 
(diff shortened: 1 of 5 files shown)


src/display/scrollbars.js
```diff
@@ -27,26 +27,26 @@ export function measureForScrollbars(cm) {
...
+    on(horiz, "scroll", () => {
+      if (horiz.clientWidth) scroll(horiz.scrollLeft, "horizontal")
+    })
+
+    this.checkedZeroWidth = false
+    // Need to set a minimum width to see the scrollbar on IE7 (but must not set it on IE8).
+    if (ie && ie_version < 8) this.horiz.style.minHeight = this.vert.style.minWidth = "18px"
+  }
 
-NativeScrollbars.prototype = copyObj({
```
marijnh: Why did you remove the copyObj? This'll remove the `constructor` property from instances.

marijnh:
I merged your Bublé patch. Not sure when the next release will happen.

src/display/scrollbars.js
marijnh: Err, duh, you made it a regular class, so that's not an issue at all. Ignore this comment.

adrianheine:
That Bublé patch is not necessary for this, since there are no object literals with short-hand methods, only classes.

marijnh:
You don't appear to be using any inheritance or class expressions, so it should work with 0.15.0 too, no?

In any case, I've merged this.

Issue #4441, opened 2016-12-15
Composition erased when press Enter key


louisatome:
To reproduce : 
- Go to CodeMirror homepage with a mobile that allow composition
- Write some text which enable composition
- Press Enter key
=> The new line has been created but the selected word has been erased from the previous line

The expected behaviour would be to validate the selected composition and then append the new line.

Tested with a Sony Xperia 3 tablet with android 6 on the latest CodeMirror release (5.21.1)

marijnh:
Could you see if attached patch helps with this problem?

louisatome:
Now the composition is not erased anymore but the composed text is added to the previous line and the new line is not created. But i think this is linked to the commit 6019b13 as i said here : https://github.com/codemirror/CodeMirror/issues/4307#issuecomment-267339571

Without this commit it's better but there is still a problem : the composed text is append to the new line instead of the previous line.

marijnh:
I'm having trouble testing this. On my desktop machine, pressing enter during a composition simply confirms the composition. On my phone, I don't have any non-latin IME installed, and I can't get anything like this to happen with the regular composition for European languages.

Without being able to reproduce this, it's really hard to debug it. Could you give more specific instructions, including platform, browser, and virtual keyboard software used?

louisatome:
I've experienced the issue with a Sony Xperia 3 tablet on android 6. The virtual Keyboard used is the Xperia keyboard and the browser is Chrome 55.

adrianheine:
This is one of many issues which are difficult to debug and solve with the fundamental approach currently taken by CodeMirror.

We are working on a rewrite (CodeMirror 6) that will overhaul input handling and should address this issue, and we are currently raising money for this work: See [the announcement](https://codemirror.net/6/) for more information about the rewrite and a demo.

Note that CodeMirror 6 is by no means stable or usable in production, yet. It is highly unlikely that we pick up this issue for CodeMirror 5, though.

PR #4443, opened 2016-12-16
Validate composition when type enter key

louisatome:
This PR fixes the issue #4441.
I don't have an exhaustive view of CodeMirror so let me know if some adjustments are needed.

### DIFF
--- a/src/edit/key_events.js
+++ b/src/edit/key_events.js
@@ -104,6 +104,11 @@ export function onKeyDown(e) {
   // IE does strange things with escape.
   if (ie && ie_version < 11 && e.keyCode == 27) e.returnValue = false
   let code = e.keyCode
+  // if Enter key, force read the DOM if a reading is required
+  // before creating the new line
+  if(code == 13 && cm.display.input.readDOMTimeout){
+    cm.display.input.readFromDOM()
+  }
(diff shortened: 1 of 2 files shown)


marijnh:
Making input polling synchronous by default like that is a bad idea, since there'll often be a bunch of extra events (input, or another composition) happening directly after a compositionend event.

The idea is that `doHandleBinding` in `key_event.js` already calls `input.ensurePolled`, which calls `forceCompositionEnd` which turn forces the composition to end and synchronously polls the content. Is this not working in your case?

louisatome:
Thanks for the review @marijnh!
OK i understand the problem with the synchronous call. I missed the ensurePolled but it doesn't work in my case, this is what happens : 

0. The user press Enter key with a selected composition
1. A `compositionend`  event is triggered
2. The handler on compositionend call the `readFromDOMSoon` and set `this.composing` to null
3. The `keydown` event is triggered
4. The handler on keydown call the `ensurePolled` which do nothing because `this.composing` is null
5. The new line is inserted and the previous is erased
6. The `pollContent` from step 2 is called but nothing has changed

marijnh:
Thanks for diagnosing this. That is indeed broken behavior. I've pushed an alternative patch d756ea1. Let's continue discussion in #4441.

PR #4453, opened 2016-12-20
Don't sort last line with no selected chars

vrana:
Selecting two full lines including line ends makes a selection from line 1 to line 3 (ch: 0). This command used to sort three lines (1 to 3) which is not what I expect. This change makes it sort only two lines if there are no selected characters on the last line.

It also fixes a fatal error if there are more than one selection on the same line (the code used 'range' instead of 'ranges').

It also selects the trailing newline after sorting the lines so that the whole lines are selected.


vrana:
Any feedback?

marijnh:
Looks great. Merged as 73c4e24a

PR #4457, opened 2016-12-22
[Sass mode] rework the mode to uniform to CSS / SCSS modes

ficristo:
I was trying to add some tests to the Sass mode, but running locally CodeMirror/test/index.html#sass_* zero test are run.
What am I missing?

### DIFF
--- a/mode/sass/test.js
+++ b/mode/sass/test.js
@@ -1,0 +1,103 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function() {
+  var mode = CodeMirror.getMode({indentUnit: 2}, "sass");
+  // Since Sass has an indent-based syntax, is almost impossible to test correctly the indentation in all cases.
+  // So disable it for tests.
+  mode.indent = undefined;
+  function MT(name) { test.mode(name, mode, Array.prototype.slice.call(arguments, 1)); }
+
+  MT("comment",
+     "[comment // this is a comment]",
+     "[comment   also this is a comment]")
+
+  MT("comment_multiline",
+     "[comment /* this is a comment]",
+     "[comment   also this is a comment]")
+
+  MT("variable",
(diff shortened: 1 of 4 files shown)


mode/sass/test.js
```diff
@@ -0,0 +1,9 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function() {
+  var mode = CodeMirror.getMode({indentUnit: 2}, "sass");
+  function MT(name) { test.mode(name, mode, Array.prototype.slice.call(arguments, 1)); }
+
+  MT("comment", "[comment // this is a comment]")
+})
```
marijnh: You're not calling your namespace function, so the code is never executed.

ficristo: Ops, I missed the obvious... Thank you.

ficristo:
I tryed to uniform the Sass mode to the CSS one, but there are a lot of edge cases, so tests still fail.
Is there a way to reuse the CSS mode even for a braceless language?
If not is there a way to at least reuse its variables somehow, like `propertyKeywords` for example?

_Note to self: check how the indentation works, some tests fail bacause of it._

ficristo:
I disabled the indentation on tests. While probably it could be improved, some test will never pass.
There are still some cases that do not corrispond to css style but for now it's good for me.

ficristo:
Simply rebased and squashed the commits.

marijnh:
Could you make this mode depend on `../css/css.js` and use the enumerated property and value lists from `CodeMirror.mimeModes["text/css"]` instead of copying them here?

ficristo:
Thank you for the guidance. Now depends on CSS mode and indeed I'm happier with it.

marijnh:
Thanks for working on this! Merged as 71af74da

ficristo:
Thank you!
I suppose I have to wait next month for a release, right? I'm not in a hurry, I would like to only have an idea for when will be released.

marijnh:
Yes, I do releases monthly (around the 20th)

Issue #4458, opened 2016-12-22
Screenreader Improvement: Remove unseen nodes from accessibility tree


schanzer:
Relevant bug in firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=995064
(Note: I know it refers to radio buttons, but this isn't actually about buttons or DOM trees at all -- it's about the _accessibility_ tree constructed by the browser.)

The Use Case:
I'm using widgets in my application, and some of the intended audience is visually-impaired. I'm exposing these widgets (and their children) using `role="tree"` on the wrapper element and `role="treeitem"` on the widgets, allowing them to navigate using standard conventions. One such convention is ordinality: the screenreader will infer this based on the accessibility tree, and add annotations like "4 of 11".

The Bug:
The accessibility tree is inferred from the DOM tree. Since CM has a complex DOM tree working behind the scenes, this inference breaks pretty badly and reports all widgets as "1 of 1".

The Fix:
It's really easy. Adding `role="presentation"` to the non-interactive nodes removes them from the accessibility tree, allowing screenreaders to function correctly. The following items need this attribute added:

1. The widget container (`<span class="CodeMirror-widget">` )
2. The line container (`<span style="padding-right: 0.1px;">`)
3. The line itself (`<pre>`)

This doesn't change anything for users who don't rely on screenreaders, but it's an enormous improvement for those who navigate CM instances which use widgets. 

Added: [example](http://www.bootstrapworld.org/preview/aria-count.html) which mimics CM's DOM, and fixes screen readers. View the source. Removing `role="presentation"` from any of these nodes destroys the accessibility tree.

marijnh:
That's a great idea. Since you've already worked through the DOM structure, and seem to understand aria attributes better than I do, maybe you could submit a pull request making this change?

schanzer:
Thanks for merging my PR!

PR #4459, opened 2016-12-22
screenreader fixes

schanzer:
Quick fix for #4458


marijnh:
Are you sure hiding widgets is a good idea? Those might contain information that's useful. Also, you ended up calling `setAttribute` on a widget object. I've merged the uncontroversial part of your patch as 3dc1a5db

schanzer:
This doesn't hide the widget. All it does it remove the container from the accessibility tree. Roll attributes are not recursive, so the actual widget dom remains part of the tree.

marijnh:
Ah, that makes sense. Merged as 85956bed

PR #4463, opened 2016-12-26
Implement sticky cursor

adrianheine:
Visual movement still doesn't always correctly work in ContentEditableInput or Webkit, but it didn't work before, either.


marijnh:
⭐ Great! ⭐ Will review later this week.

marijnh:
I'm getting test failures in Firefox 52 (core_wrappingAndResizing, core_move_bidi_Say ا ب جabj S, core_move_bidi_Sayyy ا ا ب ج) and Chrome 55 (core_rtlMovement) though the tests do pass in Phantom.

adrianheine:
I'll look into that. `core_move_bidi_` are skipped in Webkit, that's why they don't list as failures in Chrome or Phantom.

adrianheine:
Apparently Firefox 51 changed behavior around what I described in https://bugs.chromium.org/p/chromium/issues/detail?id=673405. Now, if two spaces from different bidi sections meet, only the one first in moving direction will have full width, while the second one can have a smaller width down to 0 in order to fit more content into the box. Looks especially strange with monospace text.

marijnh:
This may be a rabbit hole that has no end. I'd be in favor of continuing to selectively disable tests (though introducing a concept of expected failures might also be a good idea, to not obscure them entirely).

Did you do any measurement of performance? From what I can see, there's still no caching of line-break info. The test suite _feels_ a lot slower with these changes, but I didn't actually measure it either.

adrianheine:
>Did you do any measurement of performance? From what I can see, there's still no caching of line-break info. The test suite feels a lot slower with these changes, but I didn't actually measure it either.

I didn't do any conclusive benchmarks, but interactions felt okay. The test suite has a lot more to do at least in Firefox. I could measure phantom before and after.

> I'm getting test failures in Firefox 52 (core_wrappingAndResizing, core_move_bidi_Say ا ب جabj S, core_move_bidi_Sayyy ا ا ب ج)

I have a local workaround for this.

> and Chrome 55 (core_rtlMovement) though the tests do pass in Phantom.

For me, `يتم السحب في 05 فبراير 2014` fails. I noticed that the (visually) last character is missing, too, maybe that's the reason? Do you have any idea what's happening there? If I add a margin on the line or increase the font size, the character appears; it seems to me like it's just cut off?

adrianheine:
I can reproduce the chromium failure with master, too, same issue. Phantom passed for me locally, but travis doesn't agree. What's the current status for you?

marijnh:
I'm still seeing 18 failures in the core tests on Chrome 55 -- a bunch of which (core_moveVstuck, core_measureEndOfLine) don't have anything to do with bidi, so I assume something broke.

adrianheine:
I cannot reproduce the test failures … unless I checkout master, build, then checkout the branch and run the tests without re-building.

marijnh:
> I cannot reproduce the test failures … unless I checkout master, build, then checkout the branch and run the tests without re-building.

Odd. I'm definitely seeing them when I rebuild (at f5764e34), & triple-verify that I'm getting the rebuilt codemirror.js

adrianheine:
Ok, I'll try out other Chromiums. My Chromium 55 definitely passes all but the one test I mentioned above, whereas master fails 18 of the branch's tests. However, my Phantom also passes a lot more tests than travis.

marijnh:
Err, I don't know what I did before (really, I double-checked everything), but when I try _now_ (both 42d2188 and f5764e3) I get only a single failure (core_rtlMovement) on your branch.

So, uh, that's good! Sorry for causing confusion. I can't figure out what I was doing before. Going to blame gremlins.

marijnh:
As far as I'm concerned, the only thing holding this back from being merged is some kind of definitive proof that the performance of line-wrapping-but-bidi-less editing isn't measurably affected. Maybe set up a simple  program that 'types' out a few thousand characters and profile it in the old and new code.

adrianheine:
I've written a small script that does that and with the patch I just pushed I get between 5% and 15% performance decrease with this branch. I'm pretty sure that's still in `coordsChar`.

adrianheine:
No failure in Safari 10.

marijnh:
IE11 fails on core_move_bidi_Όȝǝڪȉۥ״ۺ׆ɀҩۏ ҳ, core_move_bidi_ŌӰтقȤ؁ƥ؅٣ĎȺ١ Ϛ, core_move_bidi_ٻоҤѕѽΩ־؉ïίքǳ ٵ, and core_move_bidi_؅؁ĆՕƿɁǞϮؠȩóć ď , but since I all the other bidi tests pass, I guess things are largely okay.

Now going to try IE8... wish me luck.

adrianheine:
Doesn't really matter, too, since Windows users normally don't get visual movement.

marijnh:
IE8 showed a whole bunch of ES5 assumptions in the tests, and a few tests failing for incidental reasons that were easy to work around. After fixing those, the move_bidi ones still fail, but the rest pass.

I wouldn't say this is entirely irrelevant, since sites can still change the default bidi motion option, but yeah, it's no huge issue.

I'm adding my pedantic code style patches and landing this at the moment.

marijnh:
Finally landed this. I added 3357762162b1a...4627e849852312, and some trivial IE-test patches.

(Still interested in caching wrapped line extent, and unifying the code that determines the wrapped line in `coordsCharInner` with `getVisualLineAround`, if that's possible.)

PR #4464, opened 2016-12-27
[css mode] Add caret-color property

mrego:
The property is defined in this spec:
https://drafts.csswg.org/css-ui/#caret-color

It has been recently implemented on Chromium and Firefox.


marijnh:
That's a draft, though, not a standard. You could add it to `nonStandardPropertyKeywords` if you want.

mrego:
Actually it's already in a Candidate Recommendation: https://www.w3.org/TR/css-ui-3/#caret-color

It was marked as *at-risk* due to missing implementations but now both Chromium and Firefox will be shipping it very soon (~March 2017).

mrego:
BTW, this has been shipped in Chrome 57 (Opera 44) and Firefox 53.

marijnh:
All right, merged as b9d0442

PR #4465, opened 2016-12-27
[css mode] Add "auto-flow" value for grid property

mrego:
See: https://drafts.csswg.org/css-grid/#grid-shorthand


marijnh:
Thanks!

PR #4467, opened 2016-12-28
Fix bug when popping scopes

Victorystick:
Although only a single scope should be popped when ending a loop (like with `{/foreach}`), the bug had the equivalent effect of all scopes being popped.

**Currently:**
soy-broken

**Fixed:**
soy-fixed


marijnh:
Thanks. Merged as 117ecfca

PR #4470, opened 2016-12-28
Fixed Material Theme Styling

### LINKED ISSUE #4469: Material Theme Selector
**Browser**: Latest versions of Chrome, Firefox, and Safari (on Mac OSX 10.11)
**Version**: 5.22.1

As is, the base [material theme selector](https://github.com/codemirror/CodeMirror/blob/master/theme/material.css#L10-L13) does not correctly override the `.CodeMirror` styles for background and text color. This causes the background and text color to remain the defaults of white and black, respectively.

image

The correct css selector should be `.cm-s-material.CodeMirror`

jakepeyser:
Fixed #4469

### DIFF
--- a/theme/material.css
+++ b/theme/material.css
@@ -7,7 +7,7 @@
 
 */
 
-.cm-s-material {
+.cm-s-material.CodeMirror {
   background-color: #263238;
   color: rgba(233, 237, 237, 1);
 }


marijnh:
Merged as 7a094220

Issue #4468, opened 2016-12-28
Don't highlight keywords that are parts of other words


Victorystick:
2bed274 had the side-effect of hightlighting parts of `worrying` in the example below.

![example-soy](https://cloud.githubusercontent.com/assets/703602/21522364/38294260-cd05-11e6-81eb-41f2db72309c.png)

dbarnett:
Looks like you just need to use `\b` in the regex and add a test.

Victorystick:
@dbarnett Right you are.

PR #4475, opened 2016-12-29
[soy mode] Add tests for 9fe2053, find bug and fix it

Victorystick:
I got too eager with #4474; forgot to add tests.

The behaviour on https://codemirror.net/ is just as #4468 described:
keyword-1-now

The situation improved after #4474, but didn't take into account the greedy nature of the parser. The regex only examined parts of the words at a time and only the latter `\b` had any effect, meaning that words ending with a keyword would have that incorrectly highlighted:
keyword-2-partial

With these changes, we grab the whole word at once and _then_ check to see if it is in fact a keyword:
keyword-3-fix

### DIFF
--- a/mode/soy/soy.js
+++ b/mode/soy/soy.js
@@ -198,8 +198,8 @@
             if (match = stream.match(/^\$([\w]+)/)) {
               return ref(state.variables, match[1]);
             }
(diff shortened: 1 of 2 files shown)


marijnh:
Merged!

PR #4476, opened 2016-12-29
Added SVG to modeInfo

teolag:
SVG files should be handled as xml files


marijnh:
Mime types ending in `+xml` are automatically treated as XML, so this isn't necessary.

teolag:
Yes I know that, but the functions findModeByExtension and findModeByFileName returns null with a filename ending with .svg
Or am I doing something wrong?

marijnh:
That's a good point. Attached patch adds `svg` as a file extension, and makes `findModeByMIME` handle fallback for `+xml` and `+json` MIMEs.

teolag:
Thank you for seeng my problem!
Your commit looks great! Happy new year!

PR #4497, opened 2017-01-16
TypeScript: test for multiple extends used

zaggino:
Test for curretly failing real project code:
```
export interface UserAttributes {
  id?: number;
  createdAt?: Date;
}
export interface UserInstance extends Sequelize.Instance<UserAttributes>, UserAttributes {
  id: number;
  createdAt: Date;
}
```
NOTE: I'm not sure which elements should be marked `variable` and which should be marked `variable-3` so feel free to alter the test case as you see proper. What's the difference between those two? Thanks!

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -634,7 +634,11 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
     if (type == "variable") {register(value); return cont(classNameAfter);}
   }
   function classNameAfter(type, value) {
-    if (value == "extends" || value == "implements") return cont(isTS ? typeexpr : expression, classNameAfter);
+    if (value == "extends" || value == "implements") {
+      return isTS ?
+        cont(commasep(typeexpr, "{"), pushlex("}"), classBody, poplex) :
+        cont(expression, classNameAfter);
+    }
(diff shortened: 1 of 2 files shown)


zaggino:
@marijnh I'd try to implement the solution if you're busy, but please hint me how the test case should look like. I'm not sure about the `variable` vs `variable-3` thing and how they differ.

adrianheine:
I think `variable-3` in this mode is for variables used as a type, so `[variable UserAttributes]` should be `[variable-3 UserAttributes]`.

zaggino:
Thanks @adrianheine , that makes sense. Sadly I doesn't seem to be able to use `commasep(typeexpr)` in the `classNameAfter` without breaking other stuff I'm not sure how to fix ... :-/

zaggino:
Probably it'd be best to define something like `typeReference` because not everything from `typeexpr` is allowed in this place, is there some hint that explains how `cont`, `pass`, `pushlex` and `poplex` things work? I seem to be very much confused in usage of these.

zaggino:
@marijnh found a solution to my testcase finally, please review

marijnh:
I've merged this as 88b37cf580032 and followed it up with 53612ac34c5c5e4, which makes it a bit simpler.

PR #4498, opened 2017-01-17
sTeX folding mode

edemaine:
CodeMirror's fold mode is awesome!  There are built-in folding modules for code, braces, Markdown, XML/HTML, but not for sTeX.  (For comparison, Ace supports all of these.)  This PR fills this hole by providing **folding support in sTeX/LaTeX**.

Conveniently, @williamstein already wrote this code; I'm just putting this together into a PR (and converting from CoffeeScript to JavaScript).  See https://github.com/sagemathinc/smc/issues/1509 which requests that I do this.

I'm new to the CodeMirror source so let me know if I missed anything.  In addition to the new fold module, I added to the obvious demo.

### DIFF
--- a/demo/folding.html
+++ b/demo/folding.html
@@ -13,10 +13,12 @@
   <script src="../addon/fold/brace-fold.js"></script>
   <script src="../addon/fold/xml-fold.js"></script>
   <script src="../addon/fold/markdown-fold.js"></script>
+  <script src="../addon/fold/stex-fold.js"></script>
   <script src="../addon/fold/comment-fold.js"></script>
   <script src="../mode/javascript/javascript.js"></script>
   <script src="../mode/xml/xml.js"></script>
   <script src="../mode/markdown/markdown.js"></script>
+  <script src="../mode/stex/stex.js"></script>
   <style type="text/css">
     .CodeMirror {border-top: 1px solid black; border-bottom: 1px solid black;}
   </style>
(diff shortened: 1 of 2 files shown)


demo/folding.html
```diff
@@ -13,10 +13,12 @@
   <script src="../addon/fold/brace-fold.js"></script>
   <script src="../addon/fold/xml-fold.js"></script>
   <script src="../addon/fold/markdown-fold.js"></script>
+  <script src="../addon/fold/stex-fold.js"></script>
```
SERVER-MSH: http://store.woltlab.com/typhoon/

marijnh:
Hi. Thanks for the PR. But to keep the main distribution maintainable by a small team, we're tightening the criteria for what goes into the main distribution, and are encouraging people to maintain code like this in separate NPM modules. (We will, eventually, also split the current main package into smaller modules.) So it'd be great if you could put this into a separate repository and publish it to NPM. It's not in scope for the main distribution.

edemaine:
OK.  Do you have a suggested naming for the NPM package, e.g. `codemirror-stex-folding`?  Or would you prefer a more "contributed" namespace?

marijnh:
I'd go with `codemirror-fold-stex`, so that other folding modules can share the `codemirror-fold-` prefix.

edemaine:
Ah yes, that is better.  FWIW, [markdown-it](https://github.com/markdown-it/markdown-it) could be a good model here, which has several self-owned plug-ins and several contributed plug-ins, including links to the latter.  In any case, I'll keep you posted about an NPM module in case you want to assemble a list of links somewhere.

adrianheine:
We already have [a list](https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons) which definitely could be improved.

edemaine:
Cool, thanks!  I'll just add to that Wiki once I get around to packaging (maybe making a section for fold-related plugins).

PR #4510, opened 2017-01-23
Add "//" as a line comment for LESS and SCSS

### LINKED ISSUE #4509: LESS and SCSS modes don't define lineComment, so toggle comment command only ever creates block comments
Not sure if this was ever working, but I recently upgraded the version of CodeMirror we use in our LESS editor, and noticed that without some previous customisations the ability toggle a line comments using a keyboard shortcut stopped working in some cases, and created block comments in others.

Dug a bit deeper and saw that we'd customised `mode/css.js` to add a `lineComment: "//"` declaration for the LESS and SCSS modes (it only has `blockCommentStart`/`End` currently). 

The `toggleComment`command tries to do a line comment first, and since LESS and SCSS support line comments I'd expect one to be used, rather than a `/* block comment */`.

Not sure if this has been brought up before and knocked back by some arguing that block comments are more common in preprocessors or something or if it's just an oversight, but either way it'd be nice to at least have _some_ way to get this behaviour out of the box (eg. a global option), rather than needing to alter the mode file.

Adding the `lineComment` definition into the css mode for LESS and SCSS mime types restored my desired behaviour.

Tested on CodeMirror 5.23.0, in Chrome 55 on macOS Sierra.

geordiemhall:
Resolves #4509 

Adds an optional `lineComment` property in the base css mode's `parserConfig`, and sets it to `//` for `text/x-less` and `text/x-scss` mime modes.

This enables the `toggleComment` command to properly toggle single line comments, as I think most would expect when using LESS or SCSS.

Am not sure if tests are required for a change like this, but feel free to yell out if so :)

### DIFF
--- a/mode/css/css.js
+++ b/mode/css/css.js
@@ -732,6 +732,7 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
     valueKeywords: valueKeywords,
     fontProperties: fontProperties,
     allowNested: true,
+    lineComment: "//",
     tokenHooks: {
       "/": function(stream, state) {
         if (stream.eat("/")) {
@@ -774,6 +775,7 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
     valueKeywords: valueKeywords,
     fontProperties: fontProperties,
     allowNested: true,
+    lineComment: "//",
     tokenHooks: {
       "/": function(stream, state) {
         if (stream.eat("/")) {
--- a/mode/css/css.js
+++ b/mode/css/css.js
@@ -28,6 +28,7 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
       colorKeywords = parserConfig.colorKeywords || {},
       valueKeywords = parserConfig.valueKeywords || {},
       allowNested = parserConfig.allowNested,
+      lineComment = parserConfig.lineComment,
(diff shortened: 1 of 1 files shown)


vincentwoo:
I'm not sure I see the value of the option. Would it make more sense and be simpler to hardcore `lineComment` to `//` for less/scss?

geordiemhall:
Yeah if you don't need the base mode to output the option then that'd be better, I just wasn't familiar enough with how the modes worked to know if it was extended over the base or just a set of options passed to the base.

mode/css/css.js
```diff
@@ -28,6 +28,7 @@ CodeMirror.defineMode("css", function(config, parserConfig) {
       colorKeywords = parserConfig.colorKeywords || {},
       valueKeywords = parserConfig.valueKeywords || {},
       allowNested = parserConfig.allowNested,
+      lineComment = parserConfig.lineComment,
```
vincentwoo: you don't need these anymore right? as long as the hardcoded linecomments are set below for just less/scss?

marijnh:
Thanks! Merged as 15aed69a

marijnh:
(Doing this via an option in the mode config seems entirely sensible to me, so I merged it as-is.)

geordiemhall:
Either works – thanks for the quick responses too :)

PR #4511, opened 2017-01-23
[java mode] Make @interface a defining keyword

Victorystick:
Although `@interface` may look like any other annotation, it is in fact a keyword that works a lot like `class`. The placement is the same, and both define a new type. This makes highlighting more consistent.

java- interface-before

With the fix:
java- interface-after


marijnh:
Seems reasonable. Merged.

adrianheine:
> Seems reasonable.

For varying degrees of reasonable. The merged commit is 268cf9905509b87de33a00342ee1cce9a1bdc5cd, btw.

Issue #2353, opened 2014-03-11
Add Dust JS Language Mode


tonytopper:
I know there are a lot of requests for language modes but I didn't see one for Dust.  Dust is a popular JavaScript templating language.

Thanks for making the great tool that is CodeMirror!

PR #4514, opened 2017-01-25
[dust mode] Adds support for dust templates.

calmdev:
Dust was originally requested in #2353 and has been on the wish list for a few years.

http://www.dustjs.com/

### DIFF
--- a/mode/dust/dust.js
+++ b/mode/dust/dust.js
@@ -1,0 +1,66 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") { // CommonJS
+    mod(require("../../lib/codemirror"),
+        require("../../addon/mode/simple"),
+        require("../../addon/mode/multiplex"));
+  } else if (typeof define == "function" && define.amd) {// AMD
+    define(["../../lib/codemirror",
+            "../../addon/mode/simple",
+            "../../addon/mode/multiplex"], mod);
+  } else { // Plain browser env
+    mod(CodeMirror);
+  }
+})(function(CodeMirror) {
+  "use strict";
+
+  CodeMirror.defineSimpleMode("dust-tags", {
+    start: [
+      { regex: /\{!/, push: "comment", token: "comment" },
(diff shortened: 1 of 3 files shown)


marijnh:
Hi. Thanks for contributing. However, we're aiming to move language modes into separate modules in the future, and are no longer accepting new languages in the core distribution. The recommended course of action is to publish and maintain this yourself, as a module called, say, `prosemirror-dust` on NPM.

calmdev:
Do you have an example repo of how that should be done? It's not clear to me, because mode files have relative paths to other modes. If I was to place the mode into a standalone repo and require it those relative paths wouldn't resolve properly. I am seeing packages on NPM with new modes that appear to be entire forks of code mirror. Ideally the mode repos would only contain source files for the actual mode, right?

claviska:
I second this. It's disenchanting to see a solid PR tossed out without a reasonable example of how it's expected to be contributed to the community. 😭

I appreciate the idea of moving to independent modes/themes, but I really think an example repo would provide a degree of consistency and help users integrate independent modes and themes easier.

Curious, will CodeMirror's core modes/themes be moved to independent repos at some point? That would provide more than enough of an example.

Thanks for the Dust contribution, @calmdev. And thanks for all your incredible work through the years, @marijnh. 👍

almibe:
Do these help?
https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons
https://github.com/optick/codemirror-mode-elixir

marijnh:
The [elixir mode](https://github.com/optick/codemirror-mode-elixir) is a good example. You specify the core module as well as other modes you depend on as dependencies, as you would with any NPM module.

As for it being disenchanting that I don't continue to adopt and maintain everybody's code, yeah, if that's what you've come to expect I've probably set the bar a bit too high in the past. I am, however, in no way obligated to keep expanding the amount of (largely unpaid) work I'm doing.

calmdev:
Thanks guys. I'll work on setting up a repo and getting the mode published.

calmdev:
https://github.com/nodewrite/codemirror-mode-dust
https://www.npmjs.com/package/codemirror-mode-dust

marijnh:
Wonderful!

adrianheine:
I added the mode to [the wiki page](https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons#external-modes).

PR #4520, opened 2017-01-31
Better Highlighting for Blocks

riverdusty:
This enables proper variable highlighting when there are three pipes in a row such as in a block that accepts an argument, thing do: [:e | | var1 var2 | ... ].


riverdusty:
freekin lint and stuff.

adrianheine:
Hi, thanks for your pull request. Your changes are quite difficult to review currently since they include a lot of unrelated white-space changes. Could you try to update your fork's branch so that it doesn't include these changes? See [Why and how to correctly amend GitHub pull requests](http://blog.adamspiers.org/2015/03/24/why-and-how-to-correctly-amend-github-pull-requests/) for how to do this.

riverdusty:
That should be easier. Not terrible for my first pull request using github, but I've learned and will try to remember. Thanks!

mode/smalltalk/smalltalk.js
```diff
@@ -91,6 +92,7 @@ CodeMirror.defineMode('smalltalk', function(config) {
     } else if (/[\w_]/.test(aChar)) {
       stream.eatWhile(/[\w\d_]/);
       token.name = state.expectVariable ? (keywords.test(stream.current()) ? 'keyword' : 'variable') : null;
+      token.name = state.expectVariable ? (variables.test(stream.current()) ? 'variable-2' : 'variable') : null;
```
marijnh: Why are you assigning to the same property twice here? This effectively disables the previous statement, which doesn't look like it is right.

mode/smalltalk/smalltalk.js
```diff
@@ -15,6 +15,7 @@ CodeMirror.defineMode('smalltalk', function(config) {
 
   var specialChars = /[+\-\/\\*~<>=@%|&?!.,:;^]/;
   var keywords = /true|false|nil|self|super|thisContext/;
+  var variables = /Smalltalk/;
```
marijnh: You shouldn't use a mode-global variable to keep state -- modes can be restarted and resumed at different points at any time, and thus state should go into the state object.

mode/smalltalk/smalltalk.js
```diff
@@ -120,6 +122,17 @@ CodeMirror.defineMode('smalltalk', function(config) {
 
     if (aChar === '|') {
       token.context = context.parent;
+      var str = stream.string;
```
marijnh: Could whatever you're trying to do here be expressed by matching a regexp on the stream at this point? Munging the whole line without regard for where in the line we currently are doesn't seem like a good idea.

mode/smalltalk/smalltalk.js
```diff
@@ -120,6 +122,17 @@ CodeMirror.defineMode('smalltalk', function(config) {
...
+      var arr = str.split('|');
+      arr.length == 3 ? str = arr[1] : str = arr[2];
+      if(!str){ str = '' };
+      var arr2 = str.split(' ');
+      for(var i=0;i<arr2.length;i++){
```
marijnh: (Also please put spaces around operators.)

mode/smalltalk/smalltalk.js
riverdusty: Could probably be better represented like,
`    } else if (/[\w_]/.test(aChar)) {
      stream.eatWhile(/[\w\d_]/);
	  if(keywords.test(stream.current()){ 
	  token.name = 'keyword'
	  } else if(variables.test(stream.current){
		  token.name = 'variable-2'
	  } else { token.name = 'variable' }`

mode/smalltalk/smalltalk.js
riverdusty: I'm not a JS developer by trade, but I am learning. I'll give the RegEx a shot. Basically I'm trying to find whether or not there are additional variables in a block, so [:var1 | |var2 var3| ...]

mode/smalltalk/smalltalk.js
riverdusty: I feel like I'm in a lecture! heh. I will make a note of this and a) fix and b) not do it again!

PR #4524, opened 2017-02-01
Added license name

amller:
Added the name of the license in the first line inorder to let Github recognize the License.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+MIT License
+
 Copyright (C) 2017 by Marijn Haverbeke <marijnh@gmail.com> and others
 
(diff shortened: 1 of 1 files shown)


amller:
Mh. Something derpy happened before, wasn't able to see the License in the repo info, thus I added the headline. Seems like it was a derp on my side. Closing Pull Request again.

marijnh:
Good idea. Merged as d0e3b2e7

Oh, our comments crossed. Not sure about what github is doing, but naming the license seems to add clarity.

amller:
Seems like Github managed to handle this. Probably I was just derping when I said I derped and the merge already happened, so I could see the License in the info.
However license info is now visible in repo info and sorry for the confusion.

PR #4527, opened 2017-02-01
[vim bindings] regexp support (sort) fixed cursor pos (ex/search)

kevmuret:
Hi everybody,

First there is a little improvement from my point of view, in the `ex` and `search` history navigation : the cursor is sticked to his position (and to his end when some text is selected).

And for the `sort` command there is regular expression support and also :

- `n` argument support (same as `d`)
- not sorting lines that doesn't match the pattern (for regexp and arguments) like vim do

PS: sorry i forgot about the console.log and tabulation. Is it possible to check again ? i have replaced my commit.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -1245,11 +1245,14 @@
           }
         }
         function onPromptKeyUp(e, query, close) {
-          var keyName = CodeMirror.keyName(e), up;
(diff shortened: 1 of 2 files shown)


marijnh:
It seems your commit is still the old one (with trailing whitespace and console.log). Could you update the branch?

PR #4534, opened 2017-02-03
add OCL mode

Morriar:
Hey there,

This PR adds a mode for the OCL language that supports the [USE-OCL](useocl.sourceforge.net) flavor for model definition.

Example:
image

Signed-off-by: Alexandre Terrasa <alexandre@moz-code.org>


marijnh:
Nice, but we're not accepting new modes. See my response to #4535

PR #4539, opened 2017-02-03
Fix and test edge cases in bidi coordsCharInner

adrianheine:


src/measurement/position_measurement.js
```diff
@@ -427,7 +427,7 @@ export function coordsChar(cm, x, y) {
 function wrappedLineExtent(cm, lineObj, preparedMeasure, y) {
   let measure = ch => intoCoordSystem(cm, lineObj, measureCharPrepared(cm, preparedMeasure, ch), "line")
   let end = lineObj.text.length
-  let begin = findFirst(ch => measure(ch).bottom < y, end - 1, 0) + 1
+  let begin = findFirst(ch => measure(ch).bottom <= y, end - 1, -1) + 1
```
marijnh: Doesn't this mean you're giving `findFirst` permission to measure the position of character -1, which isn't something that actually has meaning?

adrianheine: No, since `to` is assumed to satisfy `pred`, it is not actually tested.

marijnh: Ah, makes sense. So this is just a somewhat kludgy way to make sure `begin` can end up being 0? Any chance we can express the search predicate in such a way that the two `-1`'s and the final `+1` disappear?

(I'm deeply suspicious of code that's littered with +/- 1's, since it breeds off-by-one errors -- which I think you mentioned you had a lot of.)

adrianheine: That should do it, shouldn't it?
```javascript
let begin = findFirst(ch => measure(ch - 1).bottom <= y, end, 0)
```
I find the current version to be a bit closer to the actual intent, though. It's the character after (`+ 1`) the last character (`first` from `length - 1`) whose bottom is not `<= y`. The simpler version looks simpler, but doesn't convey any meaning. As for off-by-one errors: I found tests to be a good solution to that problem :)

marijnh: This reads as 'get the position for which the previous character is on a different wrapped line', which I find easier to parse than the old formulation.

adrianheine: I suppose if you're used to the concept of »position«, that might be the case. I amended the last commit.

marijnh: Thanks. I'd say that (despite its name) this function is finding a _position_.

marijnh:
Merged as 407c5227...7ad2ce62

PR #4540, opened 2017-02-03
Use wrappedLineExtent in moveVisually

adrianheine:
This adds to #4539, so please review that first.


marijnh:
Trying to put the last two patches in this on top of 7ad2ce62 produces relatively complicated-looking conflicts. Could you rebase them for me, since you're less likely to screw something up?

Is cb2ebf1 intended as a relatively pleasant alternative to an infinite loop for when there are bugs in moveVisually, or are there other situations that might trigger it?

adrianheine:
Rebased. a7f80181 is intended for issues in `moveVisually`, yes. Those probably could come up with wrong bidi orders, too, though.

marijnh:
Thanks! Merged.

PR #4544, opened 2017-02-04
[vim bindings] regexp support (sort) fixed cursor pos (ex/search)

kevmuret:
Hi everybody,

First there is a little improvement from my point of view, in the ex and search history navigation : the cursor is sticked to his position (and to his end when some text is selected).

And for the sort command there is regular expression support and also :

- n argument support (same as d)
- not sorting lines that doesn't match the pattern (for regexp and arguments) like vim do

Some `sort` tests have been changed in consequence and some added for regexp support, run then with `sort_pattern*`.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -1245,11 +1245,13 @@
           }
         }
         function onPromptKeyUp(e, query, close) {
-          var keyName = CodeMirror.keyName(e), up;
+          var keyName = CodeMirror.keyName(e), up, offset;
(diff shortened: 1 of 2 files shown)


marijnh:
Thank you. Merged as f2c63b67

For next time, please split different changes/features into different commits, when possible even submit them as different pull requests. And when you get something wrong in your pull request, you can simply update the branch (also completely replacing it with a 'force push'), no need to close and open a new pull request.

PR #4551, opened 2017-02-07
Update Crystal mode

### LINKED ISSUE #3887: Failing indent for trailing if/unless in crystal
Trailing if/unless are counted for indentation. 

If user inputs,

```
a = rand if true
```

Then, the cursor is left with one extra indentation level.

The ruby mode does not suffer from this, maybe there is something to reuse from there.

version: CodeMirror 5.12.0
cc: @MakeNowJust

MakeNowJust:
- Update keywords (remove `ifdef`. add `select`, `uninitialized` and `__END_LINE__`.)
  - Add new string literals (%q string and here docs).
  - Remove global variable syntax (because it is no longer supported.)
  - Improve indentation (fix #3887, and correct indentation of `abstract class`)

Thanks.


marijnh:
Thanks. Merged as 78539df

PR #4560, opened 2017-02-10
[annotatescrollbar addon] Avoid exception with annotations past EOF

alur:
Some of our scrollbar annotations update after XHRs complete, which means that we somewhat frequently get exceptions thrown when CodeMirror tries to redraw the scrollbar annotations and the number of lines has decreased.

I think it would be nice to patch the to & from in the change handler, but I'll just leave it at this for now.

### DIFF
--- a/addon/scroll/annotatescrollbar.js
+++ b/addon/scroll/annotatescrollbar.js
@@ -72,7 +72,6 @@
     var wrapping = cm.getOption("lineWrapping");
     var singleLineH = wrapping && cm.defaultTextHeight() * 1.5;
     var curLine = null, curLineObj = null;
-    var lastLine = cm.lineCount() - 1;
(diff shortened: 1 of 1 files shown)


marijnh:
How about we just drop (don't render) annotations where `getLineHandle` returns undefined? Compressing out-of-bounds annotations into the last line doesn't seem like useful behavior.

alur:
Sounds fine to me.

alur:
Actually, wait, I need to fix that code.

alur:
There, should be good now.

marijnh:
Thanks. Merged as 0ed93b4051c26, and followed up  by 4164d2847140cc8 because documents don't necessarily start at line 0, so their last line isn't always `lineCount() - 1`

PR #4562, opened 2017-02-11
[soy mode] use jsmode for innerMode to use the correct style of comments

fenghaolw:
The 'insert comment' command currently does the wrong thing for soy files. It surrounds the highlighted text in html style comments () but soy uses c-style comments (// or /* ...*/).

This change forced the innerMode to be js (instead of localstate) so that it always uses c-style comments. It also added keymap and comment for the demo page.


marijnh:
I don't think returning nonsense from `innerMode` is a good idea, since that is also used by other pieces of the system. Attached patch adds a way to tell the comment addon not to look at the inner mode.

PR #4567, opened 2017-02-12
Added "npm run compress" option to run UglifyJS 

melloware:
I find this incredibly useful and i think others would too.  This gives users the option that the old "Compress" UI website on Codemirror gave and they can tweak to their needs.  it does not run as any part of the CodeMirror build it is purely there if users do want to execute it.

Please consider this PR as other users have asked for similar functionality in other tickets.

Note: Processes modes that depend on other modes in the correct order.


marijnh:
Sorry, no, I don't want to try and create a UI for something that's easy enough for people to do on their side with just NPM and a few commands

muetzenflo:
I also find this incredibly useful! Thanks for creating this PR @melloware

melloware:
Thanks @muetzenflo glad at least someone else finds it useful.

Alex-Sour:
i think this is Great i am using it for my Project

Thanks so much for making it, I'm really excited to see all the thrilling new content.

ettoreleandrotognoli:
I'm getting this error when I try to use https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.65.16/mode/rust/rust.min.js
Does it means that I have to stop to use CDNs and compile it for my application?

PR #4574, opened 2017-02-14
[php mode] add syntax highlight for heredoc strings

ficristo:
This tentatively fixes https://github.com/codemirror/CodeMirror/issues/4550
The tests I've added pass but in the demo the heredoc string is still highlighted as string instead of using the specific mode. Any idea of why?

### DIFF
--- a/mode/php/php.js
+++ b/mode/php/php.js
@@ -3,13 +3,22 @@
 
 (function(mod) {
   if (typeof exports == "object" && typeof module == "object") // CommonJS
-    mod(require("../../lib/codemirror"), require("../htmlmixed/htmlmixed"), require("../clike/clike"));
+    mod(require("../../lib/codemirror"), require("../htmlmixed/htmlmixed"), require("../clike/clike"), require("../meta"));
   else if (typeof define == "function" && define.amd) // AMD
-    define(["../../lib/codemirror", "../htmlmixed/htmlmixed", "../clike/clike"], mod);
+    define(["../../lib/codemirror", "../htmlmixed/htmlmixed", "../clike/clike", "../meta"], mod);
   else // Plain browser env
     mod(CodeMirror);
 })(function(CodeMirror) {
   "use strict";
+
+  function getMode(name) {
+    if (CodeMirror.findModeByName) {
+      var found = CodeMirror.findModeByName(name);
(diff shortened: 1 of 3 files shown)


ficristo:
Including the `meta.js` file fixed the demo.
Before I fix the lint error, could you verify if the approach is correct?

mode/php/php.js
```diff
@@ -111,6 +136,18 @@
           stream.eatWhile(/[\w\.]/);
           var delim = stream.current().slice(before[0].length + (quoted ? 2 : 1));
           if (quoted) stream.eat(quoted);
+
+          var mode;
+          if (!quoted && delim && delim === delim.toUpperCase() && (mode = getMode(delim))) {
+            state.delim = delim;
+            state.quoted = quoted;
+            state.curMode = mode;
+            if (state.curMode) {
```
marijnh: Isn't this always true?

mode/php/php.js
```diff
@@ -111,6 +136,18 @@
           stream.eatWhile(/[\w\.]/);
           var delim = stream.current().slice(before[0].length + (quoted ? 2 : 1));
           if (quoted) stream.eat(quoted);
+
+          var mode;
+          if (!quoted && delim && delim === delim.toUpperCase() && (mode = getMode(delim))) {
+            state.delim = delim;
+            state.quoted = quoted;
+            state.curMode = mode;
```
marijnh: I think it'd be cleaner to put these three constant-over-the-whole-string fields in an object and store that in the state, so that there's less fields to reset later on (and less properties to copy when the state is copied)

ficristo:
I didn't save the `curMode` in the new object because it is used for other things.
Actually I'm also not sure if setting to null is correct...
I have the same doubt for `curState`

marijnh:
You do realize that the `curMode` you're working with is in a different state object than the one that the rest of the mode file uses, I hope?

ficristo:
I thought the `copyState` function would have been called and so the `curMode` attribute would be copied. My bad.
I moved `curMode` in the new object as asked before, but I left the change to `copyState`.
Now I'm not sure if I also need to undo that change.

marijnh:
In general, it sounds like you might need a bit of a better grasp of how this mode works before making a tricky change like this. As the code currently is, it seems too confused to merge.

Issue #4582, opened 2017-02-16
Marking text seems inconsistent when overlapping sections


bmass02:
# Problem
When marking text that overlaps one another, some inconsistencies arise. If an encompassing text is marked after an inner text, then it seems deleting any text that includes editable text deletes the uneditable text also. However, the only uneditable text that is deleted is text that follows the editable text.

# Solution
If you change the order of marking the text (i.e. mark the encompassing text first) then the marked text acts as expected (i.e. uneditable text isn't edited).

I'm not sure if this functionality is intended. If so, maybe updating the docs on marking text would help.

# Reproduction
https://jsfiddle.net/udskry4v/19/

The first CodeMirror editor has the problem.
The second has the fix.

Thanks for the great software!

bmass02:
After further looking into it, this is only the case when both marked texts have "readOnly" set to true.

marijnh:
The problem appears to go away when I run this outside of jsfiddle, with the current version of CodeMirror. If you can reproduce it as a stand-alone HTML file (preferably without jQuery or superfluous searchCursors), that'd help.

bmass02:
Here is a gist(https://gist.github.com/bmass02/0f40a6ee3dea2236b7214bd1abedfc94) I created using CodeMirror v5.23.0 with only codemirror.js and codemirror.css. It still behaves the same. Although I did notice that the inner text does not get deleted. So maybe creating the outer marked text failed silently? Thanks for the quick response!

bmass02:
Alright, I did some more digging and the problem seems to come from the "removeReadOnlyChanges" function. When attempting to remove a section of text that starts inside an editable range but finishes past the inner marked text, there are 2 part objects returned for the failing setup and only 1 for the solution setup. I'll continue to look into exactly where the problem is.

marijnh:
Fixed via #4588

PR #4588, opened 2017-02-17
Updated offset for removeReadOnlyRanges

bmass02:
Subtract by 3 to account for the 2 extra parameters for splice and for replacing an existing part.

I ran the tests for both master and my branch. There were failing tests, but the same for both branches.

I also ran bin/lint, but it seemed to already be upset with some other syntax ('let' to be exact). Since my changes are so small, I thought not to worry about it.

Let me know if there is something I am leaving out. Thanks!

Issue: #4582


marijnh:
Thanks, that looks correct. Merged. The linter and the tests are passing for me (and for the [CI](https://travis-ci.org/codemirror/CodeMirror/builds/202654686)), so I'm not sure what you saw there.

PR #4595, opened 2017-02-22
TypeScript: broken when index signature is used

zaggino:
following code is parsed incorrectly, because of index signature (it can be seen for example here https://www.typescriptlang.org/docs/handbook/interfaces.html)
```
interface A {
  [prop: string]: any;
  prop1: any;
}
```
`prop1` is currently parsed as a `variable` instead as a `property`


marijnh:
Merged as d79935c60. Fixed in 96a1ca9f

PR #4597, opened 2017-02-22
Add U+0080 through U+009F to the specialChars default

alur:
They are all control characters.


marijnh:
Thanks. Merged as 4c9e134d

PR #4598, opened 2017-02-22
Protect from null markedSelection

JoelEinbinder:
markedSelection is set to null as the  onChange and onCursorActivity events are turned off. But inside an operation it is possible for the onChange event to be called after markedSelection is set to null. This causes a runtime error.


marijnh:
Thanks. Merged as ebfed69e

Issue #4600, opened 2017-02-22
YAML linting for multi-document YAML raises an uncaught exception


ryanpetrello:
Entering the following YAML into a CodeMirror field with YAML linting:

```
---
---
foo: bar
```

...will raise the following an exception:

`Uncaught TypeError: Cannot read property 'line' of undefined`
https://github.com/codemirror/CodeMirror/blob/fc41788dd765feaaddd07958250f315f56314574/addon/lint/yaml-lint.js#L22

Instead, it should properly surface this message to the user:

https://github.com/nodeca/js-yaml/blob/9bd2c10ba3746acb4d86c614b22094b8b2b3d05b/lib/js-yaml/loader.js#L1574

PR #4601, opened 2017-02-22
properly handle linting for multi-document YAML

ryanpetrello:
js-yaml doesn't always provide line/mark data for certain exceptions;
handle this scenario more gracefully

closes: #4600


marijnh:
Thanks. Merged as bd308e71

PR #4608, opened 2017-02-25
support for ignoring whitespace within a line in diff

sfbayhacker:
This pull request includes changes to accommodate a new option - to ignore whitespace within a line. This does not ignore new line changes.

Requires user to pass option **ignoreWS** as true to ignore white space. Default is false.

### DIFF
--- a/addon/merge/merge.js
+++ b/addon/merge/merge.js
@@ -44,7 +44,8 @@
       if (Object.prototype.toString.call(classLocation) != "[object Array]") classLocation = [classLocation]
       this.classes.classLocation = classLocation
 
-      this.diff = getDiff(asString(orig), asString(options.value));
+      var ignoreWS = this.mv.options.ignoreWS || false;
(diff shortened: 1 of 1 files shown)


adrianheine:
I think I would prefer an option that takes a regular expression, but I don't know how @marijnh generally feels about this feature.

sfbayhacker:
Appreciate your comment @adrianheine. Would you be able to point out the common use case(s) for a broader regular expression based approach? This is only to educate myself.

marijnh:
I think the formulation is fine this way, since a regexp might match newlines, which would break the plugin. I merged this as 608e3f822d24 but followed up with a3891b51ec which makes the option name `ignoreWhitespace`, and simplifies the code a little.

sfbayhacker:
Thanks @marijnh.

sfbayhacker:
@adrianheine Ah, OK. Thanks. Appreciate you letting me know.

PR #4612, opened 2017-02-27
Added Promises support to linter addon

keul:
I copied the approach done on the hinter plugin.

I didn't find any test case for this.


marijnh:
Thanks! Merged as d8d77035

PR #4614, opened 2017-02-27
Support number with underscores for Kotlin 1.1.

satamas:


marijnh:
Thanks. Merged as d45038f1

PR #4616, opened 2017-02-28
[merge addon] Stop the other orig from trailing behind while scrolling an orig editor.

alur:
Currently the other original is trailing behind a bit while scrolling in one of the originals, but not when scrolling the actual editor.


marijnh:
Looks good! Merged as 17159fec

PR #4618, opened 2017-02-28
[sql mode] Added "text/x-sqlite" mime type

crazy4chrissi:
Added the "text/x-sqlite" mime type for sql-mode. Neither existing mime-types mysql nor x-sql fit well for SQLite, e.g. important keywords like PRAGMA are missing. Therefore, I added a new mime type based on the SQLite language definition available online. I added references to all sources in comments. Feel free to remove them if you think this is too much.
SQLite says identifiers like columns should be quoted in double quotes preferably, just like in standard SQL, not backticks. Therefore, I made the necessary changes to sql-hint.js, so that a mime type declaration can define the column identifier that the language uses. I guess other SQL dialects will need this as well.

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -14,6 +14,7 @@
   var tables;
   var defaultTable;
   var keywords;
+  var identifierQuote;
   var CONS = {
     QUERY_DIV: ";",
     ALIAS_KEYWORD: "AS"
@@ -26,6 +27,15 @@
     var mode = editor.doc.modeOption;
     if (mode === "sql") mode = "text/x-sql";
     return CodeMirror.resolveMode(mode).keywords;
+  }
+
+  function getIdentifierQuote(editor) {
+    var mode = editor.doc.modeOption;
+    if (mode === "sql") mode = "text/x-sql";
+    if(CodeMirror.resolveMode(mode).identifierQuote)
+      return CodeMirror.resolveMode(mode).identifierQuote;
+    else
+      return "`";
   }
 
   function getText(item) {
@@ -86,17 +96,17 @@
   }
 
   function cleanName(name) {
-    // Get rid name from backticks(`) and preceding dot(.)
+    // Get rid name from identifierQuote and preceding dot(.)
     if (name.charAt(0) == ".") {
       name = name.substr(1);
     }
-    return name.replace(/`/g, "");
+    return name.replace(new RegExp(identifierQuote,"g"), "");
   }
 
-  function insertBackticks(name) {
+  function insertIdentifierQuotes(name) {
     var nameParts = getText(name).split(".");
     for (var i = 0; i < nameParts.length; i++)
-      nameParts[i] = "`" + nameParts[i] + "`";
+      nameParts[i] = identifierQuote + nameParts[i] + identifierQuote;
     var escaped = nameParts.join(".");
     if (typeof name == "string") return escaped;
     name = shallowClone(name);
@@ -106,13 +116,13 @@
 
   function nameCompletion(cur, token, result, editor) {
     // Try to complete table, column names and return start position of completion
-    var useBacktick = false;
+    var useIdentifierQuotes = false;
     var nameParts = [];
     var start = token.start;
     var cont = true;
     while (cont) {
       cont = (token.string.charAt(0) == ".");
-      useBacktick = useBacktick || (token.string.charAt(0) == "`");
+      useIdentifierQuotes = useIdentifierQuotes || (token.string.charAt(0) == identifierQuote);
 
       start = token.start;
       nameParts.unshift(cleanName(token.string));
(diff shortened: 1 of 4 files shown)


crazy4chrissi:
Ok sorry, I will fix these.

addon/hint/sql-hint.js
```diff
@@ -28,6 +29,15 @@
     return CodeMirror.resolveMode(mode).keywords;
   }
 
+  function getIdentifierQuote(editor) {
+    var mode = editor.doc.modeOption;
```
marijnh: It's easier to do `editor.getMode()` to get the mode (and then check `mode.name == "sql"`)

marijnh:
Could you address my comment, add a minimal test for handling double-quoted identifiers to `test/sql-hint-tests.js`, and add this new dialect to `mode/meta.js`?

crazy4chrissi:
Thanks for your review. I will try to do a test case and add it to meta.js. Regarding getting the mode: I did it the same way as  function getKeywords(editor) in the same file does. So then this should be adjusted there as well I think?
```
  function getKeywords(editor) {
    var mode = editor.doc.modeOption;
    if (mode === "sql") mode = "text/x-sql";
    return CodeMirror.resolveMode(mode).keywords;
  }
```

marijnh:
Hm, on closer look, no, it makes sense the way it is now -- you're calling `resolveMode`, not `getMode`, because you need the original mode spec, not the mode object, to be able to get at its configuration. So the code is fine as it is (though I'd simplify the end of the function to ``CodeMirror.resolveMode(mode).identifierQuote) || "`"``

crazy4chrissi:
Thank you. I will do like this. I am currently looking at the `sql-hint-tests.js` and if I am not mistaken it has some problems. This test does not fail:
```
  test("keywords2", {
    value: "SEA",
    cursor: Pos(0, 3),
    list: ["SELECT"],
    from: Pos(0, 0),
    to: Pos(0, 3)
  });
```
The only change I did from the original `keywords` test is to change SEL to SEA. Shouldn't this complain that "SELECT" is not in the list of autocompletion as required by the test `keywords2`?

marijnh:
You're right, those tests weren't failing when the expected result had more elements than the produced result. Patch b359a021 fixes this.

crazy4chrissi:
Ok all done, anything else?

marijnh:
Hi, the tests have syntax errors (missing commas) and one of them fails.

crazy4chrissi:
I am sorry. I cannot reproduce this. I just did a fresh clone and only exchanged the `test/sql-hint-test.js`.
- ran bin/lint without problems
- continuous-integration does not report any problem
- `npm test` says "All passed", but this was also the case when tests should fail!?
- test/index.html reports that 1 test called `core_lineWidgets`fails, but this is the same with a fresh clone and seems to be unrelated to my tests

Can you please explain where you found syntax errors and which test is failing for you?

test/sql-hint-test.js
```diff
@@ -99,6 +109,16 @@
     to: Pos(0, 16)
   });
 
+  test("doublequotedcolumn", {
+    value: "SELECT users.\"na",
+    cursor: Pos(0, 16),
+    tables: simpleTables,
+    list: ["\"users\".\"name\""],
+    from: Pos(0, 7),
+    to: Pos(0, 16)
```
marijnh: Missing comma here

test/sql-hint-test.js
```diff
@@ -118,6 +138,16 @@
     to: Pos(0, 11)
   });
 
+  test("schemadoublequoted", {
+    value: "SELECT \"sch",
+    cursor: Pos(0, 11),
+    tables: schemaTables,
+    list: ["\"schema\".\"users\"", "\"schema\".\"countries\""],
+    from: Pos(0, 7),
+    to: Pos(0, 11)
```
marijnh: Another one here, and the test fails (no results)

marijnh:
I'm not sure what you're doing, but since the test file is syntactically invalid, you can't have been running it (probably you missed an error on the console saying that the file failed to load, and thus the tests were skipped)

crazy4chrissi:
Okay you are right, but I still don't get it why nothing on my side nor the continuous-integration check complained about the syntax error.  Attached you will find exactly what I did:
[test.txt](https://github.com/codemirror/CodeMirror/files/810578/test.txt)
I just saw that Firefox complains about the syntax error in the console when running the test/index.html. So next time I should look there as well....
Anyway. I will soon commit a fix of the tests.

marijnh:
The linter apparently doesn't check the test scripts, and I guess the phantomjs process also failed to evaluate the broken file and thus didn't define or run the tests.

marijnh:
Thanks, looks great now. Merged as 65c70cf5d18 and c4387d6073b15

crazy4chrissi:
Great thank you for all your great work. I am going to integrate CodeMirror into phpLiteAdmin, a web-based GUI to manage SQLite databases. Therefore, I need good highlighting of SQLite in CodeMirror.

PR #4619, opened 2017-03-01
add a tsx definition to meta.js and add a basic test for tsx

zaggino:
Nothing breaks - I was just exploring a bug in Brackets with TSX and wrote these to be sure the problem is not coming out of CodeMirror.


marijnh:
Looks good. Merged as 17159fec...5e427c43

PR #4620, opened 2017-03-01
[merge addon] Re-align after swapping docs.

alur:


marijnh:
Thanks for spotting that. Merged as 8d552508

PR #4622, opened 2017-03-02
situations, when type should not be parsed as a keyword

zaggino:
```
class HasType {
  type: string;
  constructor(type: string) {
    this.type = type;
  }
  setType({ type }: { type: string; }) {
    this.type = type;
  }
}
```
tested this as a valid typescript code

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -579,6 +579,7 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
     if (type == "{") return contCommasep(proppattern, "}");
   }
   function proppattern(type, value) {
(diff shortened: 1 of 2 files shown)


adrianheine:
This also works with `string`, `any`, `number`, … Microsoft's monaco editor also highlights them as keyword.

zaggino:
@marijnh I've added code to make the test-case pass. Can you please have a look at this?
All JavaScript and TypeScript tests are passing for me. Thanks!

marijnh:
I've merged your tests as 00f59d1a50ec and pushed 486507818, which solves the problem in a different way by only making `type` a keyword when at the start of a statement.

zaggino:
Thanks!

PR #4624, opened 2017-03-02
Support setting the editor direction

adrianheine:
This is based on work by Radek Piórkowski <radek25c@gmail.com>.

### DIFF
--- a/src/model/document_data.js
+++ b/src/model/document_data.js
@@ -1,9 +1,11 @@
 import { loadMode } from "../display/mode_state"
+import { runInOp } from "../display/operations"
 import { regChange } from "../display/view_tracking"
 import { Line, updateLine } from "../line/line_data"
 import { findMaxLine } from "../line/spans"
 import { getLine } from "../line/utils_line"
 import { estimateLineHeights } from "../measurement/position_measurement"
+import { addClass, rmClass } from "../util/dom"
 import { lst } from "../util/misc"
 import { signalLater } from "../util/operation_group"
 
@@ -91,8 +93,19 @@ export function attachDoc(cm, doc) {
   doc.cm = cm
   estimateLineHeights(cm)
   loadMode(cm)
+  setDirectionClass(cm)
   if (!cm.options.lineWrapping) findMaxLine(cm)
   cm.options.mode = doc.modeOption
   regChange(cm)
 }
 
+function setDirectionClass(cm) {
+  ;(cm.doc.direction == "rtl" ? addClass : rmClass)(cm.display.lineDiv, "CodeMirror-rtl")
+}
+
+export function directionChanged(cm) {
+  runInOp(cm, () => {
+    setDirectionClass(cm)
+    regChange(cm)
+  })
+}
--- a/src/util/bidi.js
+++ b/src/util/bidi.js
@@ -74,16 +74,16 @@ let bidiOrdering = (function() {
 
   let bidiRE = /[\u0590-\u05f4\u0600-\u06ff\u0700-\u08ac]/
   let isNeutral = /[stwN]/, isStrong = /[LRr]/, countsAsLeft = /[Lb1n]/, countsAsNum = /[1n]/
-  // Browsers seem to always treat the boundaries of block elements as being L.
-  let outerType = "L"
 
   function BidiSpan(level, from, to) {
     this.level = level
     this.from = from; this.to = to
   }
 
-  return function(str) {
-    if (!bidiRE.test(str)) return false
+  return function(str, direction) {
+    let outerType = direction == "ltr" ? "L" : "R"
+
+    if (str.length == 0 || direction == "ltr" && !bidiRE.test(str)) return false
     let len = str.length, types = []
     for (let i = 0; i < len; ++i)
       types.push(charType(str.charCodeAt(i)))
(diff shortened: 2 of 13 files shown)


src/model/document_data.js
```diff
@@ -91,8 +93,19 @@ export function attachDoc(cm, doc) {
...
+function setDirectionClass(cm) {
+  (cm.doc.direction == "rtl" ? addClass : rmClass)(cm.display.lineDiv, "CodeMirror-rtl");
+}
+
+export function directionChanged(cm) {
+  runInOp(cm, function(){
+    setDirectionClass(cm);
+    regChange(cm);
+  });
+}
```
adrianheine: These functions should be somewhere else, probably `display/`?

src/util/bidi.js
```diff
@@ -198,15 +198,16 @@ let bidiOrdering = (function() {
       order.push(new BidiSpan(0, len - m[0].length, len))
     }
 
-    return order
+    order.forDirection = direction
+    return direction == "rtl" ? order.reverse() : order
```
adrianheine: I wonder if this is acceptable, performance-wise.

src/model/document_data.js
```diff
@@ -91,8 +93,19 @@ export function attachDoc(cm, doc) {
...
+function setDirectionClass(cm) {
+  ;(cm.doc.direction == "rtl" ? addClass : rmClass)(cm.display.lineDiv, "CodeMirror-rtl")
+}
+
+export function directionChanged(cm) {
+  runInOp(cm, () => {
+    setDirectionClass(cm)
+    regChange(cm)
+  })
+}
```
adrianheine: These functions should be somewhere else, probably `display/`?

src/util/bidi.js
marijnh: I wouldn't put 'wild' properties on the array -- since global direction changes are probably rare, you could clean out the `order` properties on all lines when they happen.

marijnh:
Is per-line direction something you want to do in a separate PR?

adrianheine:
Either way is fine. I think this is already a useful state that could be merged, but per-line direction isn't far away either.

src/util/bidi.js
adrianheine: Why not? I could also store a wrapper `{ order: [], direction: ''}` in `line.order`, so that the property is not on an array and doesn't leak.
Global direction changes are probably rare, but this technique would also work for line direction changes, it is simple and lazy, and it exactly implements the logic.

marijnh: Clearing the order on a line direction change would also be trivial. Since we might be generating lots of order arrays, using a cheap representation (plain array) seems a good idea. (Such estimates of how expensive something is are always debatable, in a JIT world, of course.)

adrianheine: Done.

marijnh:
Thanks! Merged as 4f091edc...4f08a5c4

adrianheine:
This does not correctly work without wrapped lines. The issue is that cm.display.scroller.scrollWidth is capped at the visible width instead of giving the content's width. I think that's intended. See [jsfiddle](https://jsfiddle.net/e6esn4je/).

marijnh:
Could you elaborate on that? `display.scroller.scrollWidth` is supposed to reflect the total width of the content, but what is the effect of direction on that? Your jsfiddle creates non-scrolling elements. If I create a scrolling, direction: rtl element with overflowing content, I do see the expected `scrollWidth`

adrianheine:
Scrolling (and `scrollWidth` is probably tied to that) only works for overflowing on one side (left in rtl, right in ltr). So, `display.scroller.scrollWidth` takes right overflow into account (because it is ltr), but the relevant overflow is on the left side.

marijnh:
So our content is overflowing on the wrong side? Is the issue that the scrollable element should have the same `direction` property as the element that contains the overflowing content?

adrianheine:
I think so. Our content overflows left, but we measure `scrollWidth` on a ltr element, and I think that doesn't take left overflow into account. So, obvious solutions would be to measure somewhere else, or make the element we measure on rtl.

marijnh:
The way things work is that the scroller divs actual scroll size must match the fake scrollbars, so that touch and wheel scrolling continue to work. As such, we probably really need to do something about the browser's scroll behavior, not just our measurement.

Even scarier is be what happens when you mix ltr and rtl blocks inside a single scrollable container -- the scroll size only takes into account the ones that match the scrollable element's direction. As in this example:

```
<div style="width: 200px; overflow: auto">
  <p style="direction: rtl">ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff</p>
  <p>ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff</p>
</div>
```

Maybe browsers just aren't able to deal with this use case at all, and we should try to use some kind of hack like positioning the rtl content in such a way that its left side is aligned with the left of the ltr content. That'd be an awful kludge which would involve knowing the maximum rtl line width and somehow offsetting all rtl lines by that minus the view width, but given what CSS's default behavior here is, we may  not have a lot of choice.

marijnh:
On second thought, since per-line direction seems to be a relatively obscure use case and clearly hard, I'd also be happy to cut that feature, and search for a solution that works for whole-document RTL content, where at least we can use the known max line length for the offsetting hack, and offset the whole content at once, rather than doing so per line.

Or we could try to support scrollbars on the left, but the current set of [hacks](http://marijnhaverbeke.nl/blog/a-pathological-scrolling-model.html) that let us maintain our own custom scrollbars and hide the native ones are already overconstrained and fragile, and I don't look forward to adding another complicating factor there.

PR #4626, opened 2017-03-04
[sql-hint] identifier quote is not escaped if identifiers contain it - or spaces

crazy4chrissi:
There are several bugs in sql-hint. The following table / column names don't work correctly. Assume the identifier quote is ", then the identifier `some"name`  needs to be escaped as `"some""name"`.

Also, spaces cause problems: Assume a table is called `table name` and we start auto completion after `table na`, we don't get `table name` but all `na*`hints. 

I wrote several failing tests that show this problem contained in this pull request. Currently, this pull request fixes all except two of the failing tests. Maybe fixing the remaining two is more easy for you than me. It seems it would require me to dig deep into CodeMirror to understand what's going on.

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -97,13 +97,21 @@
     if (name.charAt(0) == ".") {
       name = name.substr(1);
     }
-    return name.replace(new RegExp(identifierQuote,"g"), "");
+    // replace doublicated identifierQuotes with single identifierQuotes
+    // and remove single identifierQuotes
+    var nameParts = name.split(identifierQuote+identifierQuote);
+    for (var i = 0; i < nameParts.length; i++)
+      nameParts[i] = nameParts[i].replace(new RegExp(identifierQuote,"g"), "");
+    return nameParts.join(identifierQuote);
   }
 
   function insertIdentifierQuotes(name) {
     var nameParts = getText(name).split(".");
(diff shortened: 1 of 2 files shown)


crazy4chrissi:
(continuous-integration fails because this pull request introduces two new tests regarding bugs that it does not fix yet)

marijnh:
Yeah, sql-hint is a mess and I regret ever having merged it. Oh well. Do you intend to work on fixing the other failures? I'd prefer not to have our CI red because of this.

crazy4chrissi:
Haha, okay. I can't believe the phpMyAdmin developers use codemirror with sql-hint and don't bother about such issues. Anyway, I will try to fix the other two issues. But it will take more time I guess because I need to better understand the way the hinter works to find where the problem is.
Maybe you can already merge the fixes provided in this pull request without the failing tests. Then I will create a new pull request once the other 2 are fixed.

marijnh:
> I need to better understand the way the hinter works to find where the problem is.

It might, depending on how much effort you want to put into this, be more productive to rewrite the thing than to try and salvage it.

I've merged 7c87848067517907d3a71cb75369b0eebfbe52b9 (minus the trailing whitespace) as 3c816659

diegogarciaa:
Please, consider this issue:

https://github.com/codemirror/CodeMirror/issues/4712

I already have a working code, I just do not know how to subit my changes.

Thanks

PR #4629, opened 2017-03-04
Sefgit patch x86asm mode

sefgit:
initial release of x86 assembly mode


marijnh:
Hi. Nice, but we're not accepting new modes into the distribution. The recommended way to publish a new mode is to put it into a separate repository and distribute it as an NPM module -- see for example the [elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir)

PR #4630, opened 2017-03-05
[vim mode] Basic implementation of special mark '

adrianheine:


marijnh:
Looks great. The reaching into the history might break at some point if we change the representation, but I don't want to add a public API for this, so it's fine, we'll notice when the test fails. Merged.

PR #4631, opened 2017-03-05
Fix soy indentation and let it restore previous local states.

axellew:
Changes in this pull request:
- Added a stack for the local states to make it possible to restore the previous state. This fixes problems when using `<tagName kind="some mode"> </tagName>`, which didn't return to the previous state after the end tag.
- Added hideIndentation() to StringStream. The hideFirstChars(), which already exists, does almost the same thing but hideIndentation does not change the behaviour of sol(), which makes the inner modes (like htmlmixed, js and css) of soy mode to indent correctly.
- Fixed indentation for local modes.

Before (the last two html tags are wrongly marked as error and indentation is wrong)
before changes
After
after-changes

The indentation might be possible to solve in some different way, and any suggestions how to mix two modes where both can change affect the indentation, is appreciated. Currently the idea is to use two counters, one for the outer mode (soy mode) and one for the inner mode (htmlmixed, js, css or text) and then add them to create the total indentation.


src/util/StringStream.js
```diff
@@ -73,9 +74,14 @@ class StringStream {
   current(){return this.string.slice(this.start, this.pos)}
   hideFirstChars(n, inner) {
```
marijnh: So now `lineStart` appears to be completely ignored, yet you somehow still left this function in. Why?

axellew: It's just not visible in the diff, there is still one line left using lineStart:
sol() {return this.pos == this.lineStart}
There is only one more mode using hideFirstChars, but it doesn't have any test cases so I'm not quite sure how just changing "sol() {return this.pos == this.lineStart}" to "sol() {return this.pos == 0}" would impact that mode, so I added hideIndentation.

marijnh: Ah, I see, I expanded the diff, but apparently not far enough, sorry. Since string stream objects are being created by the thousands during editing, adding an extra field is a non-trivial cost. Could you take a rough look at the pug mode to see if just changing the behavior of `hideFirstChars` would cause any damage there? I'd prefer to simply do that, and not introduce a new method.

axellew: I can't see any difference in the pug mode when changing sol() to "sol() {return this.pos == 0}" and looking through the code I can't see anything which should be affected, but I can't say I'm very familiar with the pug syntax or how which test cases would a be a good fit. 

Another solution is adding this to the soy mode, where it eats indentation before entering the inner mode to set it to the correct stream position:

      if (stream.sol()) {
        for (var indentEaten = 0; indentEaten < state.indent; indentEaten++) {
          if (!stream.eat(/\s/)) break;
        }
        if (indentEaten > 0) return null;
      }

marijnh: > Another solution is adding this to the soy mode

If things can be fixed without touching the StringSteam API, I'd much prefer that.

marijnh:
Thanks! Merged as c96aae3d

Issue #2411, opened 2014-03-27
';' and '\n' are not 'electricChars' in emacs (C) mode


popinet:
When using the emacs key bindings, typing ';' does not automatically indent the line in C mode. This can be tested here:

http://codemirror.net/demo/emacs.html

This works fine in straight C mode:

http://codemirror.net/mode/clike/index.html

I would also expect new lines to be indented automatically. This is the case in straight C mode but not when using emacs bindings.

Note that 'electric mode' in emacs (the editor) behaves as the straight C-mode in CodeMirror.

PR #4634, opened 2017-03-07
keymap/emacs.js: update to match Emacs >= 24.4's RET/C-j mappings.

fischman:
Starting with Emacs 24.4 the default meanings of Enter and Ctrl+J was swapped, so that
RET does newline-and-indent while C-j is the simple non-electric newline.  This
change makes codemirror match Emacs.

Related:
- Emacs 24.4 was released 20 Oct 2014, FWIW
- #2411 previously visited this partially, but predated the Emacs change by 7
  months, so was rejected.


marijnh:
Makes sense. Merged as d3e68b6e

PR #4636, opened 2017-03-08
Treat {@inject} same as {@param}

vrana:
`{@inject}` is used for injected parameters instead of normal parameters. It behaves the same as `{@param}`.


marijnh:
Thanks, merged as 438079cb

PR #4644, opened 2017-03-16
Continue markdown task lists when formated with * and + bullets

brrd:
The continuelist addon currently supports the `- [ ]` syntax only. 

Even if `* [ ]` and `+ [ ]` syntaxes are not clearly mentioned in [GFM documentation](https://guides.github.com/features/mastering-markdown/#GitHub-flavored-markdown) they are supported by most markdown parsers (such as Github's one). This PR aims to support them in the continuelist addon.

### DIFF
--- a/addon/edit/continuelist.js
+++ b/addon/edit/continuelist.js
@@ -11,8 +11,8 @@
 })(function(CodeMirror) {
   "use strict";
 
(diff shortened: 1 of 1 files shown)


marijnh:
Seems like a good idea. Merged as f583ab93

PR #4647, opened 2017-03-16
More screenreader improvements

schanzer:
Screenreaders treat all DIVs as `role="group"` by default, which impacts depth calculation. An element which is visually at the "top level" of CodeMirror, such as a widget, should have a depth of zero.

Since CM has a nest of DIVs, this throws off the depth calculation significantly. This PR overrides the role of several DIVs with `presentation`, thereby fixing the depth calculation.

@marijnh , I can't help but wonder if the `elt()` function shouldn't just assign this by default. Virtually everything from the `<pre>` nodes on up to just before the wrapper should be probably have `role="presentation"`

### DIFF
--- a/src/display/Display.js
+++ b/src/display/Display.js
@@ -1,5 +1,5 @@
 import { gecko, ie, ie_version, mobile, webkit } from "../util/browser"
-import { elt } from "../util/dom"
+import { elt, eltP } from "../util/dom"
 import { scrollerGap } from "../util/misc"
 
 // The display handles the DOM integration, both for input reading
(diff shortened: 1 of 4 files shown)


marijnh:
If you want, you can define an `eltP` function which calls `elt` and then adds this role, and use that for all places that currently explicitly set role to presentation. Doing it by default for all elements created seems a little too magical/unexpected.

schanzer:
Good call. Done.

marijnh:
Great. Merged as 0b00c0e1

PR #4653, opened 2017-03-22
Addon: Swift hint

TheSoundDefense:
This PR adds the swift-hint.js addon, adding a custom hint function to Swift mode. The addon contains a large number of Swift keywords that are built into the language, along with additional optional features.

The addon has some level of context-awareness, and is able to use the contents of the current line to discern which words would be more appropriate to suggest. It can determine if a given word/token is a property (and what type it's a property of), a type declaration, an attribute, a compiler instruction, or something else, and suggest different word lists appropriately.

The addon also can scan the current document and extract user-defined variables, types or properties and add them to the suggested word lists. These user-defined keywords also are context-aware, so a user-defined variable will not appear if a type declaration is more appropriate.

The addon includes the following new options:

- `contextAware` - a boolean that turns context-awareness on or off. The default value is `false`.
- `extractVariables` - a boolean to indicate if the addon should scan the document to extract user-defined variables. The default value is `false`.
- `extractLineLimit` - a number that represents the maximum number of lines the document should scan when looking for user-defined variables. Once the document exceeds this number of lines, the addon will stop doing new scans of the document, and instead use the set of user-defined variables that it last extracted. The default value is 2000. (Currently this scans the entire document with each character typed, which starts slowing the document down after several thousand lines depending on how many tokens are in each line, which is why it is turned off by default. Trying to figure out how to improve this without introducing bugs.)
- `keywords` - an array of strings, indicating other keywords to add to the suggested word list.


marijnh:
Hi. Nice work. But we're moving to a distribution model where we don't accept any further new modules into the main distribution, and instead encourage people to put it them on NPM as separate packages. (At some point, the main package will also be split into smaller pieces.) So please create a separate repository and NPM package for this, and mention it on [the wiki](https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons).

TheSoundDefense:
Will do, thanks!

PR #4658, opened 2017-03-27
Adding a diff viewer wrapper over merge module

abhijit-paul:
Parameters:
DOM component,
Left text,
Right text

Takes care of avoiding re-initialization of diff same DOM model, just replaces the text if the diff DOM model already exists


marijnh:
This is too special-purpose for inclusion in the main distribution, sorry.

PR #4659, opened 2017-03-27
Modularising the usage of CodeMirror library instead of using global

abhijit-paul:


marijnh:
In my comment on the previous pull request, I meant that I'm simply not interested in including this in the distribution, not that the code has to change.

PR #4665, opened 2017-03-30
Support editor rtl mode without wrapped lines

adrianheine:
This seems to work pretty well, even with changing `fixedGutter` and `direction`.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -27,6 +27,10 @@
   background-color: #f7f7f7;
   white-space: nowrap;
 }
+.CodeMirror-rtl .CodeMirror-gutters {
+  border-right: none;
+  border-left: 1px solid #ddd;
+}
 .CodeMirror-linenumbers {}
 .CodeMirror-linenumber {
   padding: 0 3px 0 5px;
@@ -165,9 +169,21 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket {color: #f22;}
   outline: none; /* Prevent dragging from highlighting the element */
   position: relative;
 }
+
+.CodeMirror-rtl .CodeMirror-scroll {
+  /* 30px is the magic margin used to hide the element's real scrollbars */
+  /* See overflow: hidden in .CodeMirror */
+  margin-left: -30px;
+  margin-right: 0;
+}
+
 .CodeMirror-sizer {
   position: relative;
   border-right: 30px solid transparent;
+}
+.CodeMirror-rtl .CodeMirror-sizer {
+  border-right: 0;
+  border-left: 30px solid transparent;
 }
 
 /* The fake, visible scrollbars. Used to force redraw during scrolling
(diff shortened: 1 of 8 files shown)


marijnh:
Nice. Some things jump out thought (tested in Chrome):

 - When doing `editor.setOption("direction", "rtl")` on the main website (/index.html), selecting with the mouse seems to completely stop working -- so I guess finding the position at coordinates broke.

 - Setting the direction back with `editor.setOption("direction", "rtl")` leaves an empty gap to the left of the code.

adrianheine:
Both should be fixed now.

marijnh:
A few more:

When you have a long line in an rtl editor, the left end of the line ends up below the scrollbar.

Also, anything beyond the first screen width, horizontally, doesn't get selection drawn (the selection rectangles stop at the original width) and doesn't properly handle mouse clicks (they are treated as if they are inside the original width)

marijnh:
(This kind of stuff is why my first impulse was to forget about putting the scrollbar on the left, but I applaud your effort to do things right.)

adrianheine:
> When you have a long line in an rtl editor (which also has a vertical scrollbar), the left end of the line ends up below the scrollbar.

Chromium?

> Also, anything beyond the first screen width, horizontally, doesn't get selection drawn (the selection rectangles stop at the original width) and doesn't properly handle mouse clicks (they are treated as if they are inside the original width)

Should be better now.

marijnh:
Yes, the content under the scrollbar happens in Chrome (57).

> Should be better now.

Great!

Also, in the editor on index.html (which auto-resizes up to a certain point), when you type a long line the horizontal scrollbar initially doesn't appear, both in Firefox and Chrome. Note that many parts of the scrolling code are written the way they are because they need to handle both the auto-resize-through-CSS and the scrollbar model, and you'll have to test both when you touch it. That stuff could use more automated tests I guess, especially since different browsers behave slightly differently with, for example, how they handle the way a margin on an absolutely positioned element influences the parent container's size.

Then I got this error a few times when clicking around in an RTL editor:

```
Error: Broke out of infinite loop in coordsCharInner  codemirror.js:2838:49
```

And clicking on the rightmost part of the document in RTL mode when there's a horizontal scrollbar (which is scrolled all the way to the right) often (but not always) causes the editor to scroll to the left, in Firefox and Chrome.

I saw the gutter in the wrong place a few times, directly after setting the direction option, or during scrolling. I can reliably reproduce this by opening mode/javascript/index.html in Chrome and running `editor.setOption("direction", "rtl")` in the console.

adrianheine:
> I saw the gutter in the wrong place a few times, directly after setting the direction option, or during scrolling. I can reliably reproduce this by opening mode/javascript/index.html in Chrome and running editor.setOption("direction", "rtl") in the console.

Fixed.

> And clicking on the rightmost part of the document in RTL mode when there's a horizontal scrollbar (which is scrolled all the way to the right) often (but not always) causes the editor to scroll to the left, in Firefox and Chrome.

I think this should only happen in Chrome, not in Firefox. I used Epiphany for Webkit testing, which as I realized today handles `scrollLeft` in rtl content like Firefox, not like Chrome.

PR #4670, opened 2017-03-31
Add tag parameter to xml-fold findEnclosingTag

mce2:
Add tag parameter to xml-fold findEnclosingTag to find enclosing tag of particular type.


coolziroonezero:
coolziro

PR #4672, opened 2017-04-03
Added wrapping span to search dialog labels to make them easier to style

Maloric:
At present it is impossible to style the label of the search dialog without affecting the other elements also.  For instance, if I want to position the buttons to the right and the label to the left, I cannot do this, even with flex box, as CSS doesn't let me target a text node.

Adding a wrapping element around the label text allows them to be styled easily without changing any existing functionality.

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -99,7 +99,7 @@
   }
 
   var queryDialog =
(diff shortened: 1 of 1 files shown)


marijnh:
That seems useful and harmless. Merged.

PR #4675, opened 2017-04-05
Added configurable search specific closeOnBlur setting

Maloric:
It seems that search and replace will often require that the dialog stays open on blur (for instance when copying the text to replace from another window).  However, rather than changing the behaviour of all dialogs in the editor, this allows the search and replace dialogs to be configured differently.

As part of this, I added a similar behaviour to the confirm dialog, which did not use the closeOnBlur setting.

Because it is desirable to have an intuitive way to close the confirm dialog (other than assigning that behaviour to a button), I also added a close behaviour if the user presses escape while any of the buttons have focus.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 .idea
 *.iml
 /lib/codemirror.js
+/.vscode
--- a/addon/dialog/dialog.js
+++ b/addon/dialog/dialog.js
@@ -123,13 +123,11 @@
         CodeMirror.on(b, "focus", function() { ++blurring; });
       }
       CodeMirror.on(b, "keydown", function(e) {
-        if (options && options.onKeyDown && options.onKeyDown(e, inp.value, close)) { return; }
         if (e.keyCode == 27) {
           b.blur();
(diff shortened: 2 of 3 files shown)


marijnh:
The regular search dialog intentionally closes on blur, since it covers part of the editor and would interfere with editing if it stayed open. If you want to create a better search dialog, please fork this one and maintain it independently.

Maloric:
In my case I have restyled the dialog with position:fixed so that it is attached to the top of the view port rather than the top of the editor.  This was mainly because if the search matched something on line 1, the dialog would obscure it.

Surely the use case I mentioned merits consideration?  Our users will often want to copy/paste from a separate window.  While I recognise that it is possible to copy the text beforehand, it still leads to a frustrating user experience that is easily remedied by a config setting that is entirely **optional**.

At the end of the day it's your product and if I need to maintain my own copy then so be it.

marijnh:
I simply don't have the bandwidth to maintain further extensions to the search addon, so yes, you'll have to maintain it yourself.

PR #4679, opened 2017-04-07
[mode soy] Don't require colon in @params

Victorystick:
If the colon `:` character was missing from a declaration like `{@param foo: type}`, the Soy mode would break the highlighting of all following code until it encountered a valid `@param` declaration. This can be quite annoying when editing templates, and it doesn't add any value to the experience. The TypeScript doesn't break highlighting due to missing `:` in argument lists. Neither should the Soy mode.

### DIFF
--- a/mode/soy/soy.js
+++ b/mode/soy/soy.js
@@ -152,8 +152,8 @@
             return null;
 
           case "param-def":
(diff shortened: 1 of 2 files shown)


marijnh:
Thank you. Merged as 0597eeef

PR #4685, opened 2017-04-11
[soy mode] Guard function calls

aparticka:


marijnh:
Thank you. Merged as 5aa64cc4

PR #4687, opened 2017-04-12
Remove duplicate declaration of 'style' variable

MatrixFrog:


marijnh:
Your commit's parent commit is from 2013, and it touches a file that hasn't existed in the repository for quite a while (it is automatically built from ES6 sources now, so it probably won't have duplicate variable definitions anymore). As such, this isn't something we can use.

PR #4690, opened 2017-04-14
Update oz.js

ffelten:
add "div" keyword


marijnh:
Thanks, merged as 7ae9e757

PR #4692, opened 2017-04-14
Fast coordsChar for bidi lines, closes #4691

### LINKED ISSUE #4691: coordsChar is slow with long bidi lines
When moving the cursor in a line with rtl characters and greater than ~5,000 lines, coordsChar is very slow. The behavior appears after 864a681b4d57823687c2

JoelEinbinder:
This passes all of the tests, and is much faster. But I don't quite understand everything the old code was doing so there is a good chance I missed something important.

### DIFF
--- a/src/measurement/position_measurement.js
+++ b/src/measurement/position_measurement.js
@@ -1,4 +1,3 @@
-import { moveVisually } from "../input/movement"
 import { buildLineContent, LineView } from "../line/line_data"
 import { clipPos, Pos } from "../line/pos"
 import { collapsedSpanAtEnd, heightAtLine, lineIsHidden, visualLine } from "../line/spans"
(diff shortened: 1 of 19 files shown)


adrianheine:
Hi, thanks for your work and for trying to understand my code! I had a quick glance to find the main difference performance-wise. I think it's that you are not measuring after every `moveVisually` step, but rather after half the logical distance. I'll have a deeper look now.

src/measurement/position_measurement.js
```diff
@@ -457,23 +457,21 @@ function coordsCharInner(cm, lineObj, lineNo, x, y) {
...
       }
-      diff = cursorCoords(cm, pos, "line", lineObj, preparedMeasure).left - x
-    } while ((dir < 0) != (diff < 0) && (Math.abs(diff) <= Math.abs(prevDiff)))
-    if (Math.abs(diff) > Math.abs(prevDiff)) {
-      if ((diff < 0) == (prevDiff < 0)) throw new Error("Broke out of infinite loop in coordsCharInner")
-      pos = prevPos
-    }
+      lastX = cursorCoords(cm, pos, "line", lineObj, preparedMeasure).left
+      let adjPos = moveVisually(cm, lineObj, prevPos, 1)
+      if (adjPos && adjPos.ch == pos.ch) break
```
adrianheine: If I understand these two lines right, they are for breaking out of the infinite loop alternating between the two cursor positions closest to the coordinates. I would prefer an approach that makes this intent obvious, instead of a side-effect. I also think this implementation is not correct for ltr spans, see the test I just pushed as d56a5f95b009f4cfa0ee0b4b4b4710348e34728b.

adrianheine:
For a 50000 character mixed-content line, this runs `coordsChar` in roughly a seventh of the time `master` takes. Nice!

JoelEinbinder:
I tried to wrap my head around more of the surrounding code. I think we can use the same logic for lines with and without bidi by converting between "visual" character locations, and internal character locations. 

Added a test for lineWrapping.

adrianheine:
I'm pretty sure your new approach doesn't correctly work with rtl content and line-wrapping. Also, the case of ltr-only content was intentionally optimized.

marijnh:
This was superseded by #4757

PR #4693, opened 2017-04-16
[clike mode] Improve destructor highlighting for C++

r-e-d:
Destructor are now parsed the same way as constructor.


marijnh:
Thanks. Merged as 1e4f5b8db174b0 and generalized a little in 78498753f9

PR #4695, opened 2017-04-17
Add the CodeMirror-focused class when scroller is focused, closes #4694

### LINKED ISSUE #4694: On clicking selected text, selection should go away immediately when it is not drag operation.
Steps to reproduce : 
1. Select some text in the editor
2. Click on the selected text to deselect it

Results : It shows the selection as disabled first and then remove it

Expected : Selection should be removed immediately without the intermediate state when the action is not drag and drop.

JoelEinbinder:


marijnh:
I prefer to keep the invariant that the editor only looks focused when it is focused as an input field, so that seeing the focused style means the user can start typing into it. As such, I prefer the workaround in f821ebabd064

PR #4697, opened 2017-04-18
Fix missing docs for passing config object to show-hint

teleclimber:
There is nothing in the docs that explains how to pass options to show-hint (or not that I could find at least). I finally found the answer here: https://discuss.codemirror.net/t/how-to-support-global-object-mode-autocomplete/536/3 and used that to update the manual.


marijnh:
Thanks. The manual did mention that you can pass an argument to `showHint`, but failed to refer to `hintOptions`. I've pushed a somewhat different patch that mentions both of these, and their relationship in 41e3936e

PR #4700, opened 2017-04-20
Add 'covariant' keyword to dart.js

srawlins:
Dart 1.22 added a new keyword: covariant. http://news.dartlang.org/2017/02/dart-122-faster-tools-assert-messages.html


marijnh:
Thanks! Merged as 63999b89

PR #4702, opened 2017-04-21
Fix error in linkRE regular expression literal of markdown mode

avonwyss:
The original character class `[^\\\[\\]]` escapes `\\` instead of `\]` (e.g. it would be the character class `[^\\\[]` followed by a `]` which seems wrong).


marijnh:
I agree that that looked wrong. Merged as 322a80b7

PR #4703, opened 2017-04-23
Factor mode improved

kerabromsmu:
More precise Factor code coloring


marijnh:
Would it be possible to make these improvements without writing a completely new interpreter for this style of mode? I really don't want to maintain a second copy of such an interpreter. What precisely does your patch improve, and why did you decide to not use `defineSimpleMode`?

kerabromsmu:
Sorry - I was sure I was using the simple mode. Fixed it now.

mode/factor/factor.js
```diff
@@ -22,50 +22,62 @@
...
     ],
     stack: [
-      {regex: /\)/, token: "meta", next: "start"},
+      {regex: /\)/, token: "keyword", next: "start"},
       {regex: /--/, token: "meta"},
-      {regex: /\S+/, token: "variable-3"},
+      {regex: /\S+/, token: "comment"},
       {
-        regex: /./,
+        regex: /(?:\s*)|./,
```
marijnh: This will match the empty string when the next character isn't whitespace. Maybe `+` instead of `*` is more appropriate? Also, the parentheses appear to not be needed, so please leave them off. Same remarks apply to two similar rules above.

mode/factor/factor.js
```diff
@@ -22,50 +22,62 @@
...
         token: null
       }
     ],
     string: [
-      {regex: /(?:[^\\]|\\.)*?"/, token: "string", next: "start"},
-      {regex: /.*/, token: "string"}
+        {regex: /(?:[^\\]|\\.)*?"/, token: "string", next: "start"},
```
marijnh: Please follow the indentation of the rest of the code -- two spaces per level.

mode/factor/factor.js
```diff
@@ -22,50 +22,62 @@
...
     ],
     string3: [
-      {regex: /(?:[^\\]|\\.)*?"""/, token: "string", next: "start"},
-      {regex: /.*/, token: "string"}
+        {regex: /(?:[^\\]|\\.)*?"""/, token: "string", next: "start"},
+        {regex: /.*/, token: "string"}
     ],
     stack: [
-      {regex: /\)/, token: "meta", next: "start"},
+      {regex: /\)/, token: "keyword", next: "start"},
```
marijnh: Style punctuation as keyword is unusual. Why did you make that change?

mode/factor/factor.js
```diff
@@ -22,50 +22,62 @@
       {regex: /#?!.*/, token: "comment"},
       // strings """, multiline --> state
       {regex: /"""/, token: "string", next: "string3"},
-      {regex: /"/, token: "string", next: "string"},
```
marijnh: Why did you remove this rule?

kerabromsmu: It's actually improved on line 26 in the new version. In Factor, there are several words that define strings. One of them is just a quote ("). The others are words ending in a quote, which is reflected as /\S*?"/.

marijnh: Ah, that makes sense.

mode/factor/factor.js
kerabromsmu: This requires a bit of explanation of the language. Factor is a sort of a modern incarnation of Forth. It has no punctuation. In fact, there are no keywords either, but some words fit better as keywords when you want to make more sense of the program. For the most part, those are the defining words - they parse the compiler input. And, Factor allows any symbols in the words except whitespace. And so, many words in standard Factor library look like puntuation.

marijnh: I get that a case can be made for seeing these as keywords, but visually, in terms of the code matching the general coloring patterns that themes are designed for, I don't think using the keyword style is a good idea here. Meta is kind of faded, which is probably also not great -- how about just styling them as `punctuation`?

mode/factor/factor.js
kerabromsmu: I'll just remove this and only leave the dot. Must have been some workaraound for some silly mistake I made while doing this, which is already fixed.

mode/factor/factor.js
kerabromsmu: It looks like punctuation is the same color as identifiers. This doesn't really help read this code. It would be better to visually separate stack effect in some way. Btw, can I find a list of all defined token styles (comment, meta, variable etc.) somewhere?

marijnh: The [theme css](https://github.com/codemirror/CodeMirror/blob/master/lib/codemirror.css#L113) is the only place where they are listed. Maybe `bracket` would be a good choice for these.

kerabromsmu: Thank you! Changed the parantheses and "--" to "bracket", stack items to "meta".

mode/factor/factor.js
kerabromsmu: So, is it good to go now?

marijnh:
Merged as 37d59ff4

PR #4705, opened 2017-04-25
Re-use setScrollLeft in operations

adrianheine:
This drops `val = Math.max(0, val)`.


marijnh:
That makes sense. Is there a reason not to do the same with the vertical scrolling and `setScrollTop`?

adrianheine:
`setScrollTop` does `updateDisplaySimple` (nice gecko hack! 🙄 ) and `startWorker`. I suppose that's not necessary in `endOperation_finish`? I could extract the three lines of actual scroll position updating and call them in `operations`, though.

marijnh:
Ah, let's just merge this as it is then. Done as d8a95221

PR #4709, opened 2017-04-26
[verilog mode] Fix js object names being highlighted as keywords.

alur:


marijnh:
Thanks, merged as 9ab45baa

PR #4711, opened 2017-04-26
Moving around scrolling-related code

adrianheine:
This moves scrolling-related code from `src/edit/methods.js` and `src/display/scroll_events.js` to `src/display/scrolling.js`. In the end, `cm.curOp.scroll(Top|Left|ToPos)` is only set in `src/display/scrolling.js`.


marijnh:
Looks good. Merged!

Issue #4355, opened 2016-10-28
[markdown] List indentation with tab


joel-porquet:
Hi,

I've started the Laverna note-taking webapp, which uses CodeMirror as its markdown editor. I noticed a behavior that seems to come from CodeMirror since I can reproduce it in the markdown demo.

Basically, when I'm writing a bullet list, pressing `<enter>` at the end of an item generates a new item below, which also starts with a bullet. Eg:

```
* my first item<enter>
* <cursor>
```

Currently, pressing `<tab>` only pushed the cursor further, without touching the bullet item:

```
* my first item
*    <cursor>
```

Now, it would be great if pressing `<tab>` could indent this new bullet item so that it actually becomes a sub-item:

```
* my first item
    * <cursor>
```

And conversely, pressing `<shift-tab>` would unindent the bullet item.

Thanks!

sbruchmann:
The [continuelist addon](https://codemirror.net/addon/edit/continuelist.js) should be a good starting point for an implementation of the requested behavior.

marijnh:
Indeed, you can add this as optional behavior to continuelist if you want, and submit a pull request, but I'm not taking feature requests, so I'm closing this issue.

PR #4714, opened 2017-04-27
List indentation addon for markdown

joel-porquet:
Hi,

A few months ago, I suggested the idea of a feature that would be quite convenient for Markdown: see #4355 

I finally found the time to work on it and came up with the implementation of a new addon, based on the continuelist addon (they could potentially be merged in the same file I imagine). This indentlist addon provides two functions to indent or unindent list items when pressing Tab or Shift-Tab. As far as I can tell, it works flawlessly in the markdown demo (mode/markdown/index.html), even with multiple selections!

Let me know if you have any questions or any improvement requests,
Thanks,
Joël

### DIFF
--- a/addon/edit/indentlist.js
+++ b/addon/edit/indentlist.js
@@ -1,0 +1,70 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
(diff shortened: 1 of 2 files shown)


marijnh:
Hi Joel, nice work. But in order to reduce the maintenance burden of this project, we're not accepting new addon functionality into the main distribution, so I encourage you to put this into a separate repository and distribute it as its own NPM package. It's not going to be merged into this repository.

joel-porquet:
FYI, just submitted https://www.npmjs.com/package/codemirror-markdown-list-autoindent for those who would be interested in using this patch ;)

PR #4716, opened 2017-04-27
Adds mode for Laravels Blade templates

bjoern-tantau:
Adds support for Laravel 4 Blade templates. https://laravel.com/docs/4.2/templates#blade-templating
Does not yet support Laravel 5 as I didn't need that. But it shouldn't be hard to add that.


marijnh:
Hi. We aren't accepting new modes into the main distribution, to keep the maintenance burden under control. The recommended way to distribute new modes it to put them on NPM as separate packages -- see [the elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir) as an example.

PR #4721, opened 2017-04-30
Removed suspicious "!" from indent-fold.js

MarkusBordihn:


marijnh:
Thanks for spotting that. Merged as 41da8714

PR #4724, opened 2017-05-02
Restore any selections widgets had after patching the display.

alur:


marijnh:
Clearing and restoring the selection like that can have all kinds of side effects (scrolling it into view, resetting IME state, messing with spell checking), so I don't think this is a good idea.

alur:
That's fair.  Let me explain what I'm actually trying to solve.

We have some contenteditable divs inside linewidgets. Right now we do this storing/restoring of the selection in response to blur & focus events on the divs. It has worked fine in the past, but now (and I think this is something new in Chrome 58) it seems like the selection offsets get set to 0 prior to the blur event firing. Unfortunately, the .focus() call just places the cursor at the beginning of the text area.

Would you be open to something like adding an event instead, so that we can store the selection before the element gets removed from the DOM?

mtaran-google:
Do any of the above proposals sound better? From our usage, this recent change (per above, originally due to Cr58) is a bug in how line widgets work. Would appreciate a response as soon as possible :)

marijnh:
Does attached patch address this? It conservatively restores selections only when they were inside the editor and the redraw changed them. Also makes sure to keep selection direction intact.

alur:
Yes, that works great. Thank you!

PR #4731, opened 2017-05-07
Fix check for undefined

robertbrignull:
When checking for definedness the `typeof` operator returns a string. Unfortunately comparing it to `undefined` will always yield false. I expect this was just a typo.

I tested this in Firefox 16 and confirmed this specific case.

This bug was found using [lgtm.com](https://lgtm.com/projects/g/codemirror/CodeMirror/), where we analyse many open source repositories including CodeMirror, and in fact we use and rely on CodeMirror every day.


marijnh:
Thanks for spotting that. Merged as 51707934

PR #4732, opened 2017-05-07
remove a double semicolon

robertbrignull:
Highly trivial I know but this PR removes a double semicolon which is technically a statement with no effect and also an unreachable statement as it comes after the `return`. Not harmful at all but easy to fix at least.


marijnh:
Thanks. Merged as e29b193

PR #4733, opened 2017-05-07
Fix a bug with the vb mode test reporting

robertbrignull:
There was a problem with the callback from the `test` function in that it tried to set the `ok` value after returning and hence all tests would always pass. You could see this by introducing a test that should have failed but didn't.
Secondly what would happen when a test failed was it would print "undefined", which wasn't very friendly, and this could be improved a bit.

This bug was found using [lgtm.com](https://lgtm.com/projects/g/codemirror/CodeMirror/).


marijnh:
Thanks for noticing that hot mess. Mode tests shouldn't actually be part of the mode's demo page, and they definitely shouldn't be rolling their own test framework. I started porting these to the regular test suite, but they were so entirely trivial that I decided to just drop them instead, in 98f2d9b2.

PR #4737, opened 2017-05-10
add handel-c langage https://en.wikipedia.org/wiki/Handel-C

FAIG2014:
a commit in the other repo will add the langage into : src/editor/classes/Languages.js


marijnh:
Hi. New language modes will have to be distributed as separate modules -- we are not accepting any new ones into the main distribution. See the [elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir) as an example.

FAIG2014:
I am not sure I understand how this project is structured.

each new languages will be in a different repo?

marijnh:
Yes, that's the idea, to make it easier to distribute the maintenance burden.

PR #4738, opened 2017-05-10
Add CaseConversion commands to the sublime keymap.

WillOursler:
Hi! First time trying this sort of thing, apologies if I've screwed something up.

I talked to Maksym Taran internally about how best to add this feature to codemirror. He suggested the sublime keymap was most appropriate. Please let me know if I should restructure it another way.

Thanks! ~Will

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -562,7 +562,7 @@
   cmds['upperCamelCaseAtCursor'] = function(cm) {
     modifyWordOrSelection(cm, function(str) {
       return splitStringToRedoCase(str)
-        .map(function(word, index) {
+        .map(function(word) {
           // If it not the first word only upper case the first char and
           // lowercase the rest.
           return word.charAt(0).toUpperCase() + word.slice(1).toLowerCase();
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -480,6 +480,112 @@
     modifyWordOrSelection(cm, function(str) { return str.toLowerCase(); });
   };
 
+  // Breaks up a given string into a list of "words" regardless of its original
+  // case.
+  //
+  // This function isn't very clever, but it does the right thing for all
+  // of the following standard cases:
+  //
+  // * snake_case
+  // * SCREAMING_SNAKE_CASE
+  // * lowerCamelCase
+  // * UpperCamelCase
+  // * Space seperated words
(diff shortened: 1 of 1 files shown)


keymap/sublime.js
```diff
@@ -480,6 +480,112 @@
...
+  // repeated--delimiters--are--ignored -> [repeated, delimiters, are, ignored]
+  function splitStringToRedoCase(str) {
+    return str
+      // This part splits up camelCased portions of the string,
+      // seperating them with whitespace.
+      .replace(/([a-z])([A-Z])/g, '$1 $2')
+      .replace(/([A-Z])([a-z])/g, ' $1$2')
+      // Trim out any leading or trailing whitespace
+      // The replaces above sometimes produce some.
+      .trim()
```
marijnh: We're still targeting ES3 runtimes, so I'm afraid you'll have to do the trimming manually.

keymap/sublime.js
```diff
@@ -480,6 +480,112 @@
...
+
+  function toUpperCaseFxn(word) {
+    return word.toUpperCase();
+  }
+
+  // These commands are designed to imitate the popular
+  // Case Conversion sublime plugin.
+  cmds['snakeCaseAtCursor'] = function(cm) {
+    modifyWordOrSelection(cm, function(str) {
+      return splitStringToRedoCase(str).map(toLowerCaseFxn).join('_');
```
marijnh: Again, ES3, so you'll need to write a `map` helper function.

marijnh:
I added a few small comments, and I have one more general reservation -- this doesn't seem to be functionality that Sublime Text actually provides, so I don't think it belongs in the Sublime Text bindings. I guess the `modifyWordOrSelection` is helpful to have access to, but I don't think that's enough justification to put this here.

Additionally, we have a rule of not accepting new functionality into the distribution that could be created as an addon. Is copy-pasting `modifyWordOrSelection`, along with your new code, into a separate addon and putting it on NPM as, say, `codemirror-case-conversion`, an option for you?

mtaran-google:
I think it'll be easier to have this be internal to our project (since publishing an open source repo involves some bureaucracy. This way we can also use new ES features easily :)

Thanks for taking a look, though!

marijnh:
Great, thanks for understanding.

PR #4746, opened 2017-05-15
[css mode] Update CSS Box Alignment properties and values

mrego:
These properties have been implemented in Firefox, Chrome and Safari.
In addition, the CSS WG has just resolved to move the spec to CR.

Spec: https://drafts.csswg.org/css-align/#property-index


marijnh:
Great. Merged as 2fb70c6d

PR #4747, opened 2017-05-15
[soy mode] Don't highlight functions in the start of tags as keywords

Victorystick:
## Before
soy-function-before

## After
soy-function-after


marijnh:
Looks good. Merged as 4c625b7a

PR #4753, opened 2017-05-18
Fix tabs chars in Stylus mode being treated like 1 space

opl-:
Fixes the bug by replacing the tab characters with an indent unit equivalent of spaces and counting those as the line indent.


marijnh:
Merged as d599d656

PR #4754, opened 2017-05-19
TypeScript: complex return type breaks function body parsing

zaggino:
```
function A() {
  return this.property;
}
function B(): Promise<{ [ key: string ]: any } | null> {
  return this.property;
}
```
In A() `property` is parsed correctly as `property`
In B() `property` is parsed as a `variable`


marijnh:
See attached patch.

zaggino:
Thanks for the quick fix!

PR #4756, opened 2017-05-22
Fix lgtm alerts

xiemaisi:
Fixed a minor issues identified by [lgtm](https://lgtm.com), see https://lgtm.com/projects/g/codemirror/CodeMirror/alerts/ for more details.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -2815,7 +2815,6 @@
         var range = {anchor: new Pos(line, baseCh), head: new Pos(line, headCh)};
         selections.push(range);
       }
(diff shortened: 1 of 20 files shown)


marijnh:
Thanks! I've merge these, except for the changes to pug.js -- there it looks like these functions are all part of a 'family' of function that _may_ use the state argument, and it'd be an easy refactoring mistake to make one of them dependent on state again, which will break if the, technically spurious, argument isn't passed.

xiemaisi:
Makes sense, thanks.

PR #4757, opened 2017-05-22
Faster coordsChar for bidi content

### LINKED ISSUE #4691: coordsChar is slow with long bidi lines

adrianheine:
This closes #4691 and is based on an idea by Joel <joel.einbinder@gmail.com> (see #4692).

### DIFF
--- a/src/measurement/position_measurement.js
+++ b/src/measurement/position_measurement.js
@@ -456,20 +456,30 @@ function coordsCharInner(cm, lineObj, lineNo, x, y) {
     if (cm.options.lineWrapping) {
       ;({begin, end} = wrappedLineExtent(cm, lineObj, preparedMeasure, y))
     }
-    pos = new Pos(lineNo, begin)
+    pos = new Pos(lineNo, Math.floor(begin + (end - begin) / 2))
     let beginLeft = cursorCoords(cm, pos, "line", lineObj, preparedMeasure).left
     let dir = beginLeft < x ? 1 : -1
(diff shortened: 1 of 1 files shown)


marijnh:
Do I understand correctly that the idea here is to do a kind of binary-subdivision-sized linear scan, where steps start large but shrink as the code overshoots its goal?

Why did we stop doing a regular binary search (in visual positions) for this again? (I think we talked about this at some point, but I forgot.)

adrianheine:
What I did before is: Move one step, measure, continue until overstepped, then choose the closer one. This patch introduces a first phase of moving several steps before measuring until the diff is pretty small, since measuring is quite expensive. In `/demo/bidi.html`, this nearly always performed 3 iterations for a click, which I consider quite good.

I don't know how a binary search over visual positions should work. This function is about mapping visual coords to logical positions. It's probably easy to guess how many columns into a line the visual coordinate is, but knowing which logical position it has is really difficult. It should be possible to really quickly move if you take bidi spans and column widths into account, but that would require having logic similar to `moveVisually`.

marijnh:
>  but that would require having logic similar to moveVisually.

The old code just called `moveVisually` X times to move X characters, which worked. So you'd do the binary search not in string-index space but in visual-position space. Probably the fact that visual order is now dependent on rendering makes this impractical.

So yeah, merging this!

PR #4759, opened 2017-05-23
TypeScript: add failing complex type casting test case

zaggino:
Hi, current code fails to parse with latest CodeMirror properly:
```
const giftpay = config.get('giftpay') as { [platformUuid: string]: { version: number; apiCode: string; } };
```


marijnh:
See c3bf7122

Issue #1909, opened 2013-10-25
CodeMirror in readonly state traps keyboard tabs


ajvincent:
If I have CodeMirror configured with the readOnly option set to true or "nocursor", then the user cannot tab over or out of the CodeMirror editing box.  A tab keystroke gets sucked in and eaten by CodeMirror.

This is an accessibility issue.

PR #4760, opened 2017-05-23
Fixes keyboard trap when readOnly option is 'nocursor'

40thieves:
Fixes a keyboard trap when readOnly option is set to `'nocursor'`. I think this has regressed from #1909

### DIFF
--- a/src/input/TextareaInput.js
+++ b/src/input/TextareaInput.js
@@ -353,7 +353,8 @@ export default class TextareaInput {
   }
 
   readOnlyChanged(val) {
-    if (!val) this.reset()
+    if (val == "nocursor") this.textarea.disabled = true
+    else if (!val) this.reset()
   }
 
   setUneditable() {}


marijnh:
What do you mean by keyboard trap?

40thieves:
Sorry, that wasn't a good PR description was it.

A keyboard trap is where you are unable to "tab out" of the input, i.e. pressing the Tab key to move focus to another input. As I understand it, the `nocursor` option allows this by preventing focus on the CM textarea. As far as I can tell, this is broken on master because the hidden textarea still receives focus. Here's a [small repro](http://jsbin.com/sekobahiqu/2/edit?html,js,output)

marijnh:
Ah, that makes sense. I've pushed an alternative patch (3091a34) since yours wouldn't re-enable the textarea when `readOnly` is set to a different value again.

40thieves:
Great! Thanks a lot

Issue #4763, opened 2017-05-24
Add Hg to the command set for the Shell Mode


KyleMcNutt:

marijnh:
Could you elaborate? As it is, I have no idea what you mean.

KyleMcNutt:
@marijnh I mean, that it would be nice to have `hg` be a highlighted command in the shell.js mode

KyleMcNutt:
@marijnh Opened PR #4764 to address this

marijnh:
Solved via #4764

PR #4764, opened 2017-05-24
Adding hg and svn as highlited commands to shell

KyleMcNutt:
Addresses #4763


marijnh:
Makes sense. Merged as 5600b5ed

Issue #4771, opened 2017-05-31
Request: more graceful Vim error handling


vincentwoo:
cc @mightyguava 

In CodeMirror's bundled Vim mode, users aren't given a lot of guidance on what options will work or not. Users on my site often just try various options. Unfortunately, whenever this happens, CodeMirror will throw an exception: https://github.com/codemirror/CodeMirror/blob/c81cba4d351ee4a51a5994b446decc4c5438ac31/keymap/vim.js#L414.

I think this is probably not the right thing to do. I don't think attempting to get/set an unavailable option counts as a runtime error, and this definitely blows up our error reporting. We could try to install a try/catch, but it's not clear where that should go. Also, the Vim mode doesn't throw a specific kind of error, so you'd have to parse the messages on all errors caught.

I propose that the Vim mode instead simply just display a message to the user that the option is not yet implemented and leave it at that.

marijnh:
This sounds like a good idea, but you'd probably have to implement it yourself and submit a PR, since no one is actively maintaining the vim bindings.

marijnh:
Solved via #4775

PR #4775, opened 2017-06-01
[vim] Don't throw errors to signal missing options

vincentwoo:
Addresses #4771

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -411,11 +411,11 @@
       cfg = cfg || {};
       var scope = cfg.scope;
       if (!option) {
-        throw Error('Unknown option: ' + name);
+        return new Error('Unknown option: ' + name);
       }
       if (option.type == 'boolean') {
         if (value && value !== true) {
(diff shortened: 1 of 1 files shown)


keymap/vim.js
```diff
@@ -411,11 +411,11 @@
       cfg = cfg || {};
       var scope = cfg.scope;
       if (!option) {
-        throw Error('Unknown option: ' + name);
+        return new Error('Unknown option: ' + name);
```
adrianheine: Why do you not use `throw` and `catch` for this?

vincentwoo: Doesn't really seem necessary. We're just trying to signal that an option was missing. We could return a hash signalling error, but this changes the least of the original code without requiring try/catch blocks.

marijnh: I agree -- exception control flow, especially in JavaScript, is quite a lot harder to deal with and easier to mess up, so a regular return seems sensible.

marijnh:
Merged as 27de924e

PR #4777, opened 2017-06-01
[scala mode] Add enum

MasseGuillaume:
Enums are available in Dotty and soon will be available in Scala.


marijnh:
Thanks, merged.

PR #4783, opened 2017-06-07
Fix invalid CodeMirror module path in PowerShell mode

craftzdog:
It caused the `powershell` mode not loading correctly.
Just fixing `require` & `define` statements to be same as other modes.


marijnh:
Thank you. Merged as 782fd40e

PR #4790, opened 2017-06-10
[yacas mode] fix the issue with # used as operator

grzegorzmazur:


marijnh:
Thanks, merged as 735ed37

Issue #4794, opened 2017-06-12
Yank and Record are written to the same register


hamir:
It seems that recordings are being written to the same register as yank.

You can repro by going to https://codemirror.net/demo/vim.html and in the editor, trying the following:

1. go to line 5
2. hit q + ' (you should now see "(recording)[']" at the bottom of the demo widget)
3. left arrow
4. p
5. you'll notice that "LEFT" gets printed (seems recordings are being written to the same register as yank)

heppe:
I captured this in a vim unit test

testVim('macro_yank_tick', function(cm, vim, helpers) {
  cm.setCursor(0, 0);
  // Start recording a macro into the \' register.
  helpers.doKeys('q', '\'');
  helpers.doKeys('y', '<Right>', '<Right>', '<Right>', '<Right>', 'p');
  helpers.assertCursorAt(0,4);
  eq('the tex parrot', cm.getValue());
}, { value: 'the ex parrot'});

The next step is to actually fix it.  Working on that now.

heppe:
a backslash on one of the quotes in the first call to helpers.doKeys got swallowed.

heppe:
This is complete.  Am I able to close it, or do I need to have someone do it for me?

marijnh:
You can close issue if you consider them solved. Closing this one.

PR #4795, opened 2017-06-13
[vim] avoid entering macro record mode with invalid register name

heppe:
relates to
https://github.com/codemirror/CodeMirror/issues/4794


marijnh:
Thanks, merged as 227ffed

PR #4797, opened 2017-06-13
Use document.getElementsByClassName directly

vincentwoo:
Sometimes attemping to call `document.body.getElementsByClassName` results in an error like: `Cannot read property 'getElementsByClassName' of null`. I'm not sure what causes `document.body` to be null, but the MDN standard seems to suggest that calling `getElementsByClassName` directly on `document` is standard now:  https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName.


marijnh:
Makes sense, merged

PR #4799, opened 2017-06-14
[vim] remove special handling for blank line from linewise delete.

heppe:
I'm not sure why that handling was ever there in the first place.  Removing it fixes my newly added test and breaks none of the unit tests.  Also verified manually in the vim demo that deleting and pasting seems to be doing the right thing.


marijnh:
I also have no idea why that was being done, so I've merged your patch as ed012e60

PR #4802, opened 2017-06-15
fix an off-by-one error in bracket matching.

heppe:
Parenthesis matching highlight is off by one.  In my example, my cursor is at the end of the line (last close parenthesis), yet the highlight is showing the matching for the previous parenthesis pair. It should be highlighting the parenthesis pair under the cursor. 

I'm not yet sure how to add a unit test for this.

### DIFF
--- a/addon/edit/matchbrackets.js
+++ b/addon/edit/matchbrackets.js
@@ -17,8 +17,8 @@
   var matching = {"(": ")>", ")": "(<", "[": "]>", "]": "[<", "{": "}>", "}": "{<"};
 
   function findMatchingBracket(cm, where, strict, config) {
-    var line = cm.getLineHandle(where.line), pos = where.ch - 1;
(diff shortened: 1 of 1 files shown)


marijnh:
There is no 'under the cursor' in CodeMirror -- the cursor is defined as being _between_ two characters. And the bracket matching prefers the character _before_ the cursor when matching (but will fall back to the one after it if the one before isn't a bracket). This is entirely intentional and is the behavior shared by, for example, Emacs' bracket matching.

heppe:
That may be, but vim, when not in insert mode, visually represents the cursor as a box highlighting the character after the cursor.  See this screen shot: https://goo.gl/photos/vrH2RAYkDn4hbVUi9

Having the bracket matching operate on any character other than what is highlighted in that case looks pretty wrong.  It's made even worse by the fact that the next character position isn't even reachable via the keyboard (though with mouse clicks it is).  Can you think of a better solution for this?

marijnh:
You could add a configuration parameter to matchbrackets that controls this, and call it with the right configuration from the vim bindings.

PR #4803, opened 2017-06-16
Parenhighlight

heppe:
minor cleanup to the interface for findMatchingBracket
    add a paren matching test to vim_test
    add a fatCursor option to config object passed to findMatchingBracket

### DIFF
--- a/addon/edit/matchbrackets.js
+++ b/addon/edit/matchbrackets.js
@@ -16,12 +16,27 @@
 
   var matching = {"(": ")>", ")": "(<", "[": "]>", "]": "[<", "{": "}>", "}": "{<"};
 
-  function findMatchingBracket(cm, where, strict, config) {
-    var line = cm.getLineHandle(where.line), pos = where.ch;
-    var match = (pos >= 0 && matching[line.text.charAt(pos)]);
+  function findMatchingBracket(cm, where, config) {
+    var line = cm.getLineHandle(where.line);
+    var pos = where.ch - 1;
+    var match;
+    // A cursor is defined as between two characters, but in in vim command mode
+    // (i.e. not insert mode), the cursor is visually represented as a
+    // highlighted box on top of the 2nd character. Otherwise, we allow matches
+    // from before or after the cursor.
+    if (config && config.fatCursor) {
+      match = matching[line.text.charAt(++pos)];
+    } else {
+      match = (pos >= 0 && matching[line.text.charAt(pos)]) ||
+              matching[line.text.charAt(++pos)];
+    }
     if (!match) return null;
     var dir = match.charAt(1) == ">" ? 1 : -1;
-    if (strict && (dir > 0) != (pos == where.ch)) return null;
+    if (config && config.scanStrict) {
+      if ((dir > 0) != (pos == where.ch)) {
+        return null;
+      }
+    }
     var style = cm.getTokenTypeAt(Pos(where.line, pos + 1));
 
     var found = scanForBracket(cm, Pos(where.line, pos + (dir > 0 ? 1 : 0)), dir, style || null, config);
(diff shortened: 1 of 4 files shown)


addon/edit/matchbrackets.js
```diff
@@ -106,13 +121,15 @@
...
     if (val) {
       cm.state.matchBrackets = typeof val == "object" ? val : {};
+      cm.state.matchBrackets.fatCursor = 
+          cm.getWrapperElement().classList.contains('cm-fat-cursor');
       cm.on("cursorActivity", doMatchBrackets);
     }
   });
 
   CodeMirror.defineExtension("matchBrackets", function() {matchBrackets(this, true);});
-  CodeMirror.defineExtension("findMatchingBracket", function(pos, strict, config){
```
marijnh: This would be a breaking change, please leave the argument list the same or at least provide backwards-compatibility code that detects and handles the old signature.

keymap/vim.js
```diff
@@ -247,15 +247,23 @@
     }
 
     function detachVimMap(cm, next) {
-      if (this == CodeMirror.keyMap.vim)
+      if (this == CodeMirror.keyMap.vim) {
         CodeMirror.rmClass(cm.getWrapperElement(), "cm-fat-cursor");
+        if(cm.state.matchBrackets) {
+          cm.state.matchBrackets.fatCursor = false;
```
marijnh: Mutating a random object that isn't owned by this code isn't a good idea. Doesn't the code that checks for the existance of a fat cursor in the dom already handle this?

addon/edit/matchbrackets.js
```diff
@@ -16,12 +16,27 @@
...
+      match = matching[line.text.charAt(++pos)];
+    } else {
+      match = (pos >= 0 && matching[line.text.charAt(pos)]) ||
+              matching[line.text.charAt(++pos)];
+    }
     if (!match) return null;
     var dir = match.charAt(1) == ">" ? 1 : -1;
-    if (strict && (dir > 0) != (pos == where.ch)) return null;
+    if (config && config.scanStrict) {
+      if ((dir > 0) != (pos == where.ch)) {
```
marijnh: I'd prefer if this remained one `if` with an `&&` in it -- nesting ifs like this seems overly verbose

addon/edit/matchbrackets.js
```diff
@@ -106,13 +121,15 @@
       cm.off("cursorActivity", doMatchBrackets);
     if (val) {
       cm.state.matchBrackets = typeof val == "object" ? val : {};
+      cm.state.matchBrackets.fatCursor = 
+          cm.getWrapperElement().classList.contains('cm-fat-cursor');
```
marijnh: Could you make this so that, instead of mutating this object, you check for `null` when actually running `findMatchingBracket`, and only if the user did not provide an option for `fatCursor`, you default it based on the presence of the class name.

Also, we still support old browsers that don't have `classList`, so a regexp check on `className` is preferred.

marijnh: Additionally, please call the option `afterCursor` instead of `fatCursor`, since it may also be useful in cases that don't involve a fat cursor.

heppe:
made changes according to your comments, but still need to fix the unit tests

addon/edit/matchbrackets.js
heppe: done * 3.  I wasn't sure if checking the DOM every time the cursor moves was kosher, but maybe it's ok

addon/edit/matchbrackets.js
heppe: done

keymap/vim.js
heppe: that code is run only once on startup.  This code is run every time we attach/detach the vim keymap (but too early for the startup code).  Per your other comments, I reverted this change.

addon/edit/matchbrackets.js
heppe: reverted.  I felt that it was cleaner having all the configs coming from the same object, but needign a compatibility layer would be even worse.

marijnh:
Oops, I didn't read your last comment properly and merged this (dbb2645c3d7d2). Looks like some vim commands _need_ the character in front of the cursor to be considered.

marijnh:
Attached patch fixes the tests again

marijnh:
It'd be great if you could submit a separate PR with a documentation update for this new option

PR #4805, opened 2017-06-17
Add proper MIME type for Rust

danilaml:
leave the old one for now (maybe some old code depends on it).

See:
https://bugs.freedesktop.org/show_bug.cgi?id=90487
https://github.com/rust-lang/kate-config/pull/6


marijnh:
Thanks! Merged as f1d002ff

PR #4808, opened 2017-06-19
docs only change

heppe:
Added a couple of sentences describing the options to findMatchingBracket


marijnh:
Merged as fdff25dfa76d7, but that text won't help people actually control these options, so I've followed up with f11978387aa and 1c186477f5bcf

PR #4810, opened 2017-06-19
Export style in package.json

toddself:
This way it can be included more easily with javascript bundlers like
browserify/sheetify


marijnh:
Oops, I thought we were already doing this. Thanks for fixing it! Merged as 1cbb7b73

PR #4814, opened 2017-06-21
Added new SQL keywords supported by Postgres 10

mzabuawala:
Ref: https://www.postgresql.org/docs/10/static/sql-keywords-appendix.html


marijnh:
Thanks. Merged as ee3a8f72

Issue #2254, opened 2014-02-13
scrollCursorIntoView() doesn't always show entire character following cursor


redmunds:
This only applies when word wrapping is turned off.

When using right-arrow key to navigate a file, `scrollCursorIntoView()` doesn't always scroll right to ensure that character following cursor is entirely in view. Here's a screen shot showing this in a CodeMirror demo:

![cursor3](https://f.cloud.github.com/assets/1197144/2162085/00f23a08-94ce-11e3-8f05-a9a6c1db0408.png)

This is a problem because you need to know what the character following the cursor is for certain operations such as Delete key.

This seems to be worse in Brackets than with the CodeMirror.net demos, but I think this is because the viewport in Brackets can easily be adjusted to be any width (i.e. not necessarily an exact multiple of char width) where the viewport width of the online demos is fixed.

This issue recently became more important in Brackets because support was added for a different (underscore) cursor for overwrite mode:

![cursor1](https://f.cloud.github.com/assets/1197144/2162146/bbe8211a-94ce-11e3-9587-b759f2c78018.png)

So, in some case, the entire cursor cannot be seen:

![cursor2](https://f.cloud.github.com/assets/1197144/2162150/d77efa20-94ce-11e3-86a4-47408a06bdf0.png)

The `cursorScrollMargin` setting was increased to `3`, but that setting is only used when determining vertical (not horizontal) scroll position, so this only fixes problem with down-arrow scrolling.

I looked at the code in `scrollCursorIntoView()` but I don't understand all of the cases that are being handled, so I'm not sure of the best way to fix it.

PR #4823, opened 2017-06-26
Use cursor stickiness in ensureCursorVisible

adrianheine:
This way, only one char's left and right edges are used, instead of its
(logical) neighbour's right edges.

In bidi content, this avoids trying to scroll to two logically close,
but visually far apart positions in some situations.

Refer to #2254 for the original reason for this code.

### DIFF
--- a/src/display/scrolling.js
+++ b/src/display/scrolling.js
@@ -35,6 +35,13 @@ export function maybeScrollWindow(cm, rect) {
 export function scrollPosIntoView(cm, pos, end, margin) {
   if (margin == null) margin = 0
   let rect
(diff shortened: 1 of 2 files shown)


marijnh:
Good idea. Is there a reason we can't set up `scrollToPos` itself to be intelligent about position stickiness, so that we can simply store the same position as `from` and `to` here, and let a more primitive operation handle it?

adrianheine:
I'll first add a test for the misbehavior I saw and then I look into that. Does make sense, though :)

adrianheine:
I moved it and added a test.

marijnh:
Looks great! Merged.

PR #4824, opened 2017-06-27
[search] Apply normalization in adjustPos, too

adrianheine:
This prevents the test case search_expandingCaseFold from hanging in
Firefox 54 on Linux. Also enables the test in phantom.


marijnh:
Merged, and followed up with f1319c8d, which makes sure `adjustPos` is using the right conversion function.

adrianheine:
Right, thanks :)

PR #4828, opened 2017-06-29
Provide better error messages if the source library is missing.

georgestephanis:
This should help developers building this functionality out that may not
realize they need to source the linting library separately.

I ran into this after having it fail silently before I realized I needed the external library.


marijnh:
Thanks, merged as 03d99e5

PR #4842, opened 2017-07-08
[julia mode] Julia 0.6 syntax updates

pabloferz:
This is an update to properly handle some syntax changes from Julia 0.6 (also includes some other fixes).

### DIFF
--- a/mode/julia/julia.js
+++ b/mode/julia/julia.js
@@ -23,24 +23,27 @@ CodeMirror.defineMode("julia", function(config, parserConf) {
   var uChar = "([^\\u0027\\u005C\\uD800-\\uDFFF]|[\\uD800-\\uDFFF][\\uDC00-\\uDFFF])";
 
   var operators = parserConf.operators || wordRegexp([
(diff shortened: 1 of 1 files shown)


pabloferz:
C.c. @marijnh

marijnh:
Thanks, merged.

danilaml:
@pabloferz Btw, is there a reason for `in` not to be included in keywords? (`for i in 1:42 ...`)

pabloferz:
It is not included as a keyword but it is treated as an operator (in the example page over the codemirror site operators are not highlighted, but they are for example in jupyter notebooks).

danilaml:
@pabloferz thanks for answering! Was confused since [here](https://docs.julialang.org/en/stable/manual/control-flow/#man-loops-1) it's referred as `the alternative (but fully equivalent) keyword in or ∈`

jasonsun0310:
I think so, too. ''''struct''' is a recently added keyword to Julia to replace '''type'''. Perhaps code mirror didn't update this change in time.

PR #4845, opened 2017-07-11
Expose startOperation/endOperation in CodeMirror API.

asolove:
This PR makes it possible for users to call `startOperation` and `endOperation` to batch changes that occur asynchronously. It also adds docs for those methods making clear that they are not for normal use cases and require some care. ([As discussed in this Discuss thread](https://discuss.codemirror.net/t/expose-startoperation-and-endoperation/1301/2))


marijnh:
Thanks!

PR #4846, opened 2017-07-11
Add N-Quads support to N-Triples mode

BigBlueHat:
[N-Quads](https://www.w3.org/TR/n-quads/) is a format based on [N-Triples](https://www.w3.org/TR/n-triples/) with an additional fourth element for stating which graph the triples are/were encoded in.

Consequently, the color coding for N-Triples works just fine for N-Quads documents.

I've updated the N-Triples demo to state that it also supports N-Quads and added an example document to prove it. 😄 

Also, I've added `application/n-triples` as a supported media type for N-Triples (leaving the old, though incorrect `text/n-triples` for backwards compatibility.

Lastly, I've updated `mode/meta.js` to properly handle the `application/n-triples` and `application/n-quads` media types as well as the `.nq` file extension.

Hope it's useful!
🎩


BigBlueHat:
I'd also be happy to provide a similar PR for adding [`text/n3`](https://www.w3.org/TeamSubmission/n3/) and [`application/trig`](https://www.w3.org/TR/trig/) as available formats for the current `text/turtle` mode--as both of those formats are compatible with Turtle document highlighting and have [registered media types](https://www.iana.org/assignments/media-types).

I'll make sure I'm doing it right via this commit first, though. 😄 Thanks!

marijnh:
Seems like a good idea. Merged as c4174e0569e...d9e926358854f

Issue #4849, opened 2017-07-12
[meta] add .editorconfig


dwelle:
http://editorconfig.org/

Would make working on forks much easier. I can set up a basic config and issue a PR, if that's ok?

marijnh:
Sure, that'd be useful.

PR #4852, opened 2017-07-12
add .editorconfig

dwelle:
closes https://github.com/codemirror/CodeMirror/issues/4849

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,8 @@
-*.txt   text
-*.js    text
-*.html  text
-*.md    text
-*.json  text
-*.yml   text
-*.css   text
-*.svg   text
+*.txt   text eol=lf
+*.js    text eol=lf
+*.html  text eol=lf
+*.md    text eol=lf
+*.json  text eol=lf
+*.yml   text eol=lf
+*.css   text eol=lf
+*.svg   text eol=lf


dwelle:
on second thought... the EOL seems not really consistent throughout the project

dwelle:
At least some files in `bin/` folder seems to have LF only.. IMO I'd normalize the EOLs in that folder to CRLF

dwelle:
speaking of which, made the editorconfig itself `LF`-only by mistake due to my default settings... :man_facepalming:

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = crlf
```
marijnh: The repository most certainly does not use crlf line endings. It may be that git adds those when you check out on windows, but this should not be in the editor config.

dwelle: in that case we can either ignore it (I'll remove the eol from the editorconfig), or we can ensure files are checked out with `lf`, because currently it's forced to autocrlf on windows due to the `.gitattributes` config (which takes precedence over global git config, where I have `autocrlf=false`).

We could tweak the `.gitattributes` as follows:

```
*.txt   text eol=lf
*.js    text eol=lf
*.html  text eol=lf
*.md    text eol=lf
*.json  text eol=lf
*.yml   text eol=lf
*.css   text eol=lf
*.svg   text eol=lf
```

marijnh: I think the reason that .gitattributes is there at all is that people kept committing CRLF files and I had to clean them up. Would your proposed change cause git to clean them up on commit?

dwelle: should ensure `LF` both on check-out and check-in (commit)

dwelle:
force-pushed change to `.editorconfig` commit, and added the `.gitattributes` changes

marijnh:
Thanks, merged. Eof is _supposed_ to be all `\n`, but it's very well possible that that hasn't been enforced everywhere.

dwelle:
yea, `CRLF` can burn

PR #4853, opened 2017-07-13
[sublime keymap] "Add to selection" bindings

### LINKED ISSUE #4850: Ctrl-Shift-Up/Down hotkey for adding cursor
I noticed there is no hotkey yet for adding a cursor one line up or down like in Sublime. Here is a quick implementation that can be added to the `extraKeys` option:
```
'Shift-Alt-Down': (codeMirror) => {
    var selections = cm.listSelections();
    var lastSelection = selections[selections.length - 1];
    cm.setSelection(lastSelection.anchor, lastSelection.head);
    cm.execCommand('goLineDown');
    var newSelection = cm.listSelections()[0];
    cm.setSelections(selections);
    cm.addSelection(newSelection.anchor, newSelection.head);
}
```
It would be nice if something like this could be added to the Sublime keymap. There may be a better way to do it, but I couldn't find a method to get the position immediately above the cursor, so that's why I used the `goLineDown` command.

nickasd:
These bindings clone the cursors by moving them up or down by one line (Ctrl-Alt-Up/Down on Windows, Cmd-Shift-Up/Down on Mac). Closes #4850.

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -164,7 +164,7 @@
     if (fullWord)
       cm.state.sublimeFindFullWord = cm.doc.sel;
   };
-  
+
   function addCursorToSelection(cm, dir) {
     var ranges = cm.listSelections(), newRanges = [];
     for (var i = 0; i < ranges.length; i++) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks, merged as 3073bdd9

adrianheine:
There seems to be an issue with this change, see #5109. @nickasd, could you have a look?

PR #4854, opened 2017-07-14
Remove midnight theme top border and bottom border

tylerlong:
Because the borders don't make much visual sense. And they cause dual scroll bar issue when the editor takes 100% of page height.

You can reproduce the issue here: http://mdp.tylingsoft.com/  Press <kbd>CMD-,</kbd> to bring up the preferences panel and change the theme to midnight.

### DIFF
--- a/theme/midnight.css
+++ b/theme/midnight.css
@@ -12,8 +12,6 @@
     color: #D1EDFF;
 }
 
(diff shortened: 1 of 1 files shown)


tylerlong:
Dual scroll bar issue:

image

Other themes don't have this issue.

marijnh:
Agreed, merged as 9a223d81

Issue #2813, opened 2014-09-18
[gfm mode] no gfm highlight after header immediately followed by code block


cben:
Look at last section of http://codemirror.net/mode/gfm/ - SHAs, github issue refs and the naked URL on last line are all unhighlighted.

Some experiment shows this happens after seeing a header immediately followed by a fenced code block.  In such a case, `overlay.codeBlock` stays false after 1st fence and becomes true after 2nd fence, which prevents further highlighting.  After some digging, I have no clue why this only happens directly following a header (header > text > code block is fine)

test case: 
https://github.com/cben/CodeMirror/commit/b3057ea4985a2474e4f7003f24e3e98a0a1e0349
https://cdn.rawgit.com/cben/CodeMirror/b3057ea4985a2474e4f7003f24e3e98a0a1e0349/test/index.html#gfm*
(first failed test is the bug; second without header works, failed artificially to let you [display states] and compare)  

cc @bfrohs

PR #4856, opened 2017-07-14
[markdown] make list tokenizer more CommonMark compliant

dwelle:
Basically a rewrite of list tokenization (mostly) to make _more_ (but still not fully) CommonMark compliant. Further improvements in the future (see todos).

### improvements

- better nested/un-nested list parsing
- correctly recognizes numbered-lists' base indentation
- should work better with tabs (though no one should use tabs with markdown anyway)

### todos (future)

- recognize line continuation when there's single EOL and next line is not block markdown
- recognize setext header inside lists

### notes

- removed `forceBlankLine` hack coz both https://github.com/codemirror/CodeMirror/issues/2813 and https://github.com/codemirror/CodeMirror/commit/4ef6a904c6312a8ca3d7ba1d81c88a78bbcba076 should work out of the box now (+ current algo didn't work well if we didn't cancel tokenization of blank lines)
- kept the `true`/`null`/`false` behavior for lists

image


dwelle:
rebased to latest master and amended

dwelle:
force pushed once again to align with project's styleguide

marijnh:
Thanks for working on this. Merged as 0f8d7593

PR #4860, opened 2017-07-17
Line numbers should be placed before dropmarkers

684sigma:
Firefox bug 1381523:
Line numbers are created after dropmarkers in some cases. They should be placed before dropmarkers.
https://bugzilla.mozilla.org/show_bug.cgi?id=1381523

### DIFF
--- a/src/display/gutters.js
+++ b/src/display/gutters.js
@@ -26,7 +26,7 @@ export function updateGutters(cm) {
 export function setGuttersForLineNumbers(options) {
   let found = indexOf(options.gutters, "CodeMirror-linenumbers")
   if (found == -1 && options.lineNumbers) {
-    options.gutters.unshift(["CodeMirror-linenumbers"])
+    options.gutters.unshift("CodeMirror-linenumbers")
   } else if (found > -1 && !options.lineNumbers) {
     options.gutters = options.gutters.slice(0)
     options.gutters.splice(found, 1)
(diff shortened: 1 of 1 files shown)


marijnh:
I don't want to change the current behavior here. If you want control over the place of the line number gutter, just include it explicitly in your `gutters` option.

684sigma:
> If you want control over the place of the line number gutter
I don't want to have any customization of that kind, I want it to work exactly as you intended, but it seems there's still a misunderstanding.
Example.

1 Open https://codemirror.net/demo/folding.html
2 Execute `editor.setOption("lineNumbers",false); //disable line numbers` in web console
3 Execute `editor.setOption("lineNumbers",true); //enable line numbers` in web console

So, before I changed line numbers option, line numbers were placed before dropmarkers, but then the order changed.

Please confirm that after toggling line numbers twice gutters should change their order.

Also, if codemirror instance initially created without line numbers, and later they were enabled, line numbers appear after dropmarker. Please confirm that it is intended order of gutters.

marijnh:
The demo sets `gutters: ["CodeMirror-linenumbers", "CodeMirror-foldgutter"]`. If you want to toggle line numbers while keeping the gutters in the same order, you have to make sure you also update the gutters option when re-enabling line numbers. The same happens with your patch, if you happen to want them on the right.

684sigma:
I see, it works as intended. Bug 1381523 is completely on Firefox side.

It seems that codemirror doesn't have or care about any "default" order of gutters. Just saying, to avoid mistakes of codemirror users (like Bug 1381523), it would be nice to have a warning about the actual behavior in the manual:
https://codemirror.net/doc/manual.html#option_lineNumbers
For example:
> Before enabling line numbers, make sure that you specified correct order of gutters, using [gutters](https://codemirror.net/doc/manual.html#option_gutters) option. (If `"CodeMirror-linenumbers"` isn't listed, it will default to be to the right of all other gutters)

PR #4861, opened 2017-07-18
[lint plugins] Allow lint plugins to output HTML formatted messages

nijel:
This is opt-in, the plugin needs to produce message_html instead of
message to be rendered as HTML.

We were using this with custom patch at phpMyAdmin, but I think it's better to integrate such feature upstream.

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -112,7 +112,11 @@
     if (!severity) severity = "error";
     var tip = document.createElement("div");
     tip.className = "CodeMirror-lint-message-" + severity;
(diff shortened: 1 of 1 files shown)


marijnh:
Looks good. Could you use `messageHTML` as a property name instead to make it follow the general capitalization style of the rest of the code?

nijel:
Good point, changed to ``messageHTML``.

marijnh:
Thank you. Merged!

nijel:
Thanks, that was fast!

PR #4862, opened 2017-07-19
SQL line comments start with "--"

antimatter15:
SQL line comments start with "--" unless it's explicitly something else (i.e. `//` or `#`). Currently `lineComment` in the SQL mode is set to `null` whenever there isn't an explicit exception, which forces typical SQL modes to only do block comments.

### DIFF
--- a/mode/sql/sql.js
+++ b/mode/sql/sql.js
@@ -193,7 +193,7 @@ CodeMirror.defineMode("sql", function(config, parserConfig) {
 
     blockCommentStart: "/*",
     blockCommentEnd: "*/",
-    lineComment: support.commentSlashSlash ? "//" : support.commentHash ? "#" : null
+    lineComment: support.commentSlashSlash ? "//" : support.commentHash ? "#" : "--"
   };
 });
 


adrianheine:
Thanks for your pull request! [As far as I see](https://stackoverflow.com/questions/17554061), `--` is indeed one of the most common and portable ways to have a line comment in SQL. It already seems to be supported in Line 74ff, though. Did you try it, for example on the [CodeMirror website](https://codemirror.net/mode/sql/index.html)?

antimatter15:
This change affects `lineComment` as consumed by the official `comment` plugin. 

Right now, hitting "Cmd+/" toggles a block comment (even on languages which support `-- ` line comments). This pull request toggles the line comment.

adrianheine:
I see! I suppose the white-space padding for MySQL is not necessary since it is automatically added by the comment plugin. I think it makes sense to untangle what a dialect supports and what is used for `lineComment`, or at least make `lineComment` configurable. Since `-- ` is so portable I would think it makes sense to use it per default in all dialects.

marijnh:
I think this is fine as it is. Merged as cb01fa18

Issue #4858, opened 2017-07-16
SUGGESTION: (MarkDown mode) Colorize Emoji Names (or better display a emoji)


jlchereau:
Emojis names are displayed within colons as in `:...:`.
`:smile:` actually gives :smile:

dwelle:
Sounds like a good idea. Should not be part of plain markdown, but gfm [specifies](https://guides.github.com/features/mastering-markdown/) emojis so I'd put it there.

I can look at it next week.

/cc @marijnh

dwelle:
The basic functionality has been implemented, so can be closed - I don't think the replace-emoji-markup-with-actual-emoji is applicable for markdown mode, at this time.

javisperez:
Hey @dwelle ! 👋Thanks for this update, I'm trying to implement the `replace-emoji-markup-with-actual-emoji` part, do you have any help on how to replace the text on the mode itself? my idea so far is to read all lines on the content, match everything between `:` (e.g. `:thinking_face:`) and replace it with the proper unicode character, but that sounds slow.

Is there anyway to replace the text when the token is matched? like, token by token?

dwelle:
I don't think syntax mode is made for replacing text. If you register a `change` handler and scan only what's been written (current line), or pasted (block of lines), then perf shouldn't be an issue.

javisperez:
Thanks @dwelle yeah, i imagined the mode limitations but I didn't know you could actually scan only what's been written from the `change` event, will dig a bit more into that as my first idea was to use `regexp` to match emoji "names" and replace the whole content, which sounds like not ideal for me. 

Thank you for the tip, any other tip is more than welcome :)

Have a great day!

PR #4864, opened 2017-07-20
[gfm mode] add emoji rule

dwelle:
notes:

- the token regex allows for `\w!?+-` characters to be between the `:` markup. Since there's no emoji spec that I know of (and because e.g. `markdown-it` allows pretty much anything the user specifies), I chose this character set to both 1) cover the most common emojis 2) minimize probability the regex will match non-emoji text.
- didn't add default `.cm-emoji` to `codemirror.css` to not bloat the file since `gfm` is the only mode right now to utilize it (added a note for users in the `index.html`)
- :question:  not sure if to disable by default

closes https://github.com/codemirror/CodeMirror/issues/4858

### DIFF
--- a/mode/gfm/gfm.js
+++ b/mode/gfm/gfm.js
@@ -115,7 +115,8 @@ CodeMirror.defineMode("gfm", function(config, modeConfig) {
   var markdownConfig = {
     taskLists: true,
     fencedCodeBlocks: '```',
-    strikethrough: true
+    strikethrough: true,
+    emoji: true
   };
   for (var attr in modeConfig) {
     markdownConfig[attr] = modeConfig[attr];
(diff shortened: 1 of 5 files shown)


adrianheine:
> the token regex allows for \w!?+- characters to be between the : markup. Since there's no emoji spec that I know of (and because e.g. markdown-it allows pretty much anything the user specifies), I chose this character set to both 1) cover the most common emojis 2) minimize probability the regex will match non-emoji text.

GitHub links to [this cheat sheet](https://www.webpagefx.com/tools/emoji-cheat-sheet/). All emojis in there match `[a-z_\d+-]`.

dwelle:
Yea, I'm aware of that -- not sure if that's really authoritative. I started with that regex but decided to include `?`/`!` so it covers `:?:` etc. coz it seemed it might be common. I can revert.

marijnh:
Thank you! Merged as b11179d1f6e1a. I've added patch 4292c2ff1bad0, which changes the default token type to `"builtin"`, so that regular themes will style these (token types in the markdown mode aren't terribly semantically-meaningful at the moment, because at the moment the widely themed set of types doesn't cover a lot of prose-specific types)

PR #4873, opened 2017-07-24
Use background-color for cm-searching

barrucadu:
At work we use a spell-checker plugin for CodeMirror, and give misspelled words a wiggly red underline with this CSS:

```css
.cm-spell-error {
    background-image: image-url('resources/red_underline.png');
    background-position: bottom;
    background-repeat: repeat-x;
}
```

Unfortunately, misspelled things don't get underlined when you search for them, because the `cm-searching` style sets the `background` property, rather than just the `background-color`, so it overrides everything.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -319,8 +319,8 @@ div.CodeMirror-dragcursors {
 .CodeMirror-line::-moz-selection, .CodeMirror-line > span::-moz-selection, .CodeMirror-line > span > span::-moz-selection { background: #d7d4f0; }
 
 .cm-searching {
-  background: #ffa;
-  background: rgba(255, 255, 0, .4);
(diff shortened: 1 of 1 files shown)


adrianheine:
Hm, I'm not sure if there's a universally right way to do this. I can imagine that someone uses `background-image` for highlighting and actually prefers it to be overridden when searching instead of not having highlighting. In your case, adding `!important` to the spelling error's style would be a workaround if I'm not mistaken.

barrucadu:
Ok, that's a good point.  We'll just set `!important` then.

marijnh:
I think change is harmless and a slight improvement, since if you want to override the whole background, you're likely to set a fully-covering background image, which will hide the background-color anyway. I've gone ahead and merged this as 3ec7088b

PR #4877, opened 2017-07-25
added minification docs

swill:
Now that the `compress.html` convenience file has been removed by https://github.com/codemirror/CodeMirror/commit/518fb67418baeec4d3eef41976af79cc0ce45e93#commitcomment-23283614, it is important that we at least document some details around how to produce compressed JS files for people who would like to deployed minified files to production.

I think it is reaching that everyone is using a build pipeline to build and compress all the files going into deployment, so I have tested and documented an approach for compressing the files manually without any requirements on a local environment.

Cheers...

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -32,6 +32,7 @@
   </ul>
   <ul>
     <li><a href="#usage">Basic Usage</a></li>
+    <li><a href="#compress">Minify / Compress</a></li>
     <li><a href="#config">Configuration</a></li>
     <li><a href="#events">Events</a></li>
     <li><a href="#keymaps">Key maps</a></li>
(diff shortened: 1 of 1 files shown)


marijnh:
Hi, I'm sorry, but I think this is something users can research outside of our docs, and I don't want to tie our docs to some specific compression page that may stop working in the future, so I'd rather not include this section in the manual.

swill:
If nothing else, please consider adding documentation to guide that the compression or minification must be done outside of what you guys offer and what files need to be added to the minification.  I spent like two hours yesterday trying to find the old `compress.html` file to figure out how to get the code in a minified state.  I did not upgrade from 5.12 for months simply because it was not clear how to produce the minified version. 

The current documentation is not clear how to upgrade from previously minified code to a new process.  I think this is a major problem.  AT LEAST call out that you have to minify on your own since you dropped the feature.  That alone would have saved me like 2 hours of searching yesterday...

Issue #4151, opened 2016-08-07
GFM & Markdown mode get wrong syntax highlight when using code block with blockquote


jackycute:
1. Go http://codemirror.net/mode/gfm/
2. Input below codes

``````
> ```
> asdasdasdasd
> ```

asdasdasdasd
``````

You'll see only the first `>` get correct color.
And the syntax highlight not turning back after the code block end.

![image](https://cloud.githubusercontent.com/assets/2639151/17463207/b1aec2ea-5cf3-11e6-9eb5-2e2843132feb.png)

Thank you for helping! Really appreciated!

**Comment by ryancui-** [comment]
Also have a wrong syntax hightlight when typing this:

```
# Title
## title 1

> - aaa

## title 2
```

using list in quote cause a mistake. Thanks for fixing this.

dwelle:
Gonna look into this.

dwelle:
associated commit doesn't really "fix" this, but for now IMO better than nothing  :blush:

PR #4880, opened 2017-07-27
[markdown mode] disallow lists and fencedCode inside blockquote

dwelle:
Semi-fixes https://github.com/codemirror/CodeMirror/issues/4151 where fencedBlock in blockquotes was badly styled and lists leaked outside.

This is not a CM-compliant fix but is better than the previous buggy behavior.

Sometimes in the future I may look into making it CM-compliant (i.e. properly allow lists and codeBlocks inside blockquotes).


marijnh:
Thanks, I've merged this -- I agree it's a reasonable way to control damage for now

PR #4881, opened 2017-07-28
Add additional MIME types

BigBlueHat:
Found a few more useful additions.

Most of these findings are from using the https://github.com/jshttp/mime-db/ project which attempts to aggregate the IANA, Apache, and nginx type lists.

I've not made it through their whole JSON file (yet! :wink:):
https://github.com/jshttp/mime-db/blob/master/db.json

I'll send a few more of these (most likely) as I trip over ones I need. Hope that's OK!

Tnx!
🎩

### DIFF
--- a/mode/asciiarmor/asciiarmor.js
+++ b/mode/asciiarmor/asciiarmor.js
@@ -68,6 +68,7 @@
   });
 
   CodeMirror.defineMIME("application/pgp", "asciiarmor");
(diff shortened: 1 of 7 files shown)


marijnh:
Thanks, that's useful!

BigBlueHat:
Happy to help! Thanks for all you're building. :heart:

PR #4883, opened 2017-07-31
Fix broken links

hokaccha:
- before: http://codemirror.net/mode/demo/vim.html
- after: http://codemirror.net/demo/vim.html


marijnh:
Thanks! Merged.

Issue #2143, opened 2014-01-15
Markdown header issue.


lukasoppermann:
Hey,

I think I found a "bug" in markdown.

When I type the following, only the `=====` gets classes for being a header, and the text `Header` does not. I think either both should get this class or none of the two, or maybe the `======` could get an additionally class, if neither of the above work. Otherwise it makes it hard to style correctly. Thanks.

```
Header
======
```

brondsem:
As a workaround, you can style the previous lines with this snippet in your own code.  I'm not sure if there would be an appropriate place to include this as part of the built-in markdown mode.

```javascript
function updateSectionHeaderStyles(cm, change) {
  var lines = cm.lineCount();
  for (var i = Math.max(0, change.from.line-1); i <= Math.min(change.to.line+1, lines-1); i++) {
    var line = cm.getLineHandle(i);
    cm.removeLineClass(line, 'text', 'cm-header');
    cm.removeLineClass(line, 'text', 'cm-header-1');
    cm.removeLineClass(line, 'text', 'cm-header-2');
    var lineTokens = cm.getLineTokens(i);
    var tok = lineTokens[0];
    if (!tok || !tok.type || tok.type.indexOf('header') === -1) {
      // first token could be some spaces, try 2nd
      tok = lineTokens[1];
    }
    if (tok && tok.type && tok.type.indexOf('header') !== -1
      && tok.string !== '#') { // not ATX header style, which starts with #
      var classes = tok.type.
        split(' ').
        filter(function(cls) { return cls.indexOf('header') === 0; }).
        map(function (cls) { return 'cm-' + cls; }).
        join(' ');
      var prev_line = cm.getLineHandle(i-1);
      cm.addLineClass(prev_line, 'text', classes);
    }
  }
}
myCodeMirror.on("change", updateSectionHeaderStyles);
updateSectionHeaderStyles(myCodeMirror, {from: {line: 0}, to: {line: myCodeMirror.lineCount()}});
```

tobloef:
Any possibility of this issues getting reopened as a feature request? It would be very nice if the workaround posted by brondsem was native to CodeMirror.

dwelle:
Made a quick naive fix [over here](https://github.com/dwelle/CodeMirror/commit/7cb4e1588119835c37fb4b8e981625ac4ea95cdc). It's still not 100% CM-compliant, but it fixes styling for the line immediately before the `------` setext line.

[...]

PR #4884, opened 2017-07-31
[markdown mode] improve CommonMark compliance for several rules

dwelle:
# Notes

- improving CommonMark compliance
- aimed at common cases, thus some fixes are still not 100% CommonMark-compliant
- separated PR into several commits to bunch related changes together for easier review
- __SUGGESTION:__ I'd make `fencedCodeBlocks` default in `markdown` to align with CommonMark. It'll simplify the code + solve some bugs which I didn't fix for this purpose.

# Commit notes

- __commit 1__ - _improve setext & hr tokenization_
    - `setext` tokenization
        - semi-fixes [2143](https://github.com/codemirror/CodeMirror/issues/2143) and other improvements
        - __NOT-CM-COMPLIANT:__ looks ahead just one line, thus:
            ```
            heading-line-1
            heading-line-2
            -----
            ```
            will parse as `header-2` only line `[1,2]`, but I aimed at most common cases.
        - shuffled the code a bit due to precedence + took a bit of liberty with stylyguide in the setex conditional expression because it was getting crazy
    - `hr` tokenization
         - recognizes few cases such as multiple hrs after each other, and other improvements
         - __NOT-CM-COMPLIANT:__ require empty line before `hr` if inside list (IMO better to have false-negatives than false-positives -- will lead people to cleaner code, e.g. blank lines around hrs) -- will be fixed in the future.
- __commit 2__ - _improve header, list & fencedCode behavior around blockquote &_ indentation
    - fixes some cases such as lists/fencedCode directly after blockquotes
    - fixes some cases where the block should result in `codeBlock` instead of `header`/`list`...
    - disabled `header` in blockquotes, too, until we implement proper block tokenizing inside blockquotes.
- __commit 3__ - _auto-terminate fencedCode after exiting list_
    - as per CommonMark. Works only for `gfm`/`fencedCodeBlocks`, see notes above.
- __commit 4__ - _improve blockquote indentation behavior_
    - ensure indented blockquotes correctly infers blockquote-levels
    - other indentation fixes

Notable changes:

image

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -270,14 +270,17 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
   }
 
   function local(stream, state) {
-    if (state.fencedChars && stream.match(state.fencedChars)) {
+    var hasExitedList = state.indentation < state.listStack[state.listStack.length - 1];
+    if (state.fencedChars && (hasExitedList || stream.match(state.fencedChars))) {
       if (modeCfg.highlightFormatting) state.formatting = "code-block";
-      var returnType = getType(state)
+      var returnType;
+      if (!hasExitedList) returnType = getType(state)
       state.localMode = state.localState = null;
       state.block = blockNormal;
       state.f = inlineNormal;
       state.fencedChars = null;
       state.code = 0
+      if (hasExitedList) return switchBlock(stream, state, state.block);
       return returnType;
     } else if (state.fencedChars && stream.skipTo(state.fencedChars)) {
       return "comment"
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -139,6 +139,7 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
 
   function blockNormal(stream, state) {
     var sol = stream.sol();
+    var firstTokenOnLine = stream.column() === state.indentation;
     var prevLineLineIsEmpty = lineIsEmpty(state.prevLine);
     var prevLineIsIndentedCode = state.indentedCode;
     var prevLineIsHr = state.hr;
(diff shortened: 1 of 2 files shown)


dwelle:
Btw, do we frown upon adding custom props on `stream` object? I'd like to either cache some line-specific data, which would be much easier than putting it in `state` where we'd have to initialize/copy/reset manually.

Also, would like to cache `stream.prevLineState` on the `stream` object.

marijnh:
> Btw, do we frown upon adding custom props on stream object?

Yeah, I'd prefer if you didn't do that.

> I'd like to either cache some line-specific data

A kludge I've used in the past is to cache stuff using two mode-local variables, one with the cached value and one with the stream object for which the cache is known to be valid. That way, you relatively convenient caching without doing weird things.

marijnh:
Thanks for the patches! Going by the tests, those are some great improvements. I was of the opinion that the Markdown mode would have to be rewritten from scratch at some point, but I guess moving it towards CommonMark compliance one step at a time might also work.

Merged as 4d448b2976...c81346e2c235961292c

dwelle:
Do you think we should get rid of the `gfm` fencedCode and move it to base markdown as suggested in OP?

marijnh:
The `gfm` mode _definitely_ needs to be rewritten, to either be a proper wrapping mode around the Markdown mode, or just a different configuration of the Markdown mode (i.e. moving all its logic into conditionally enabled features in markdown.js). It's currently keeping redundant state, which is very error prone, and is generally a pile of hacks that doesn't work very well.

dwelle:
> The gfm mode definitely needs to be rewritten

Fore sure. But before then, what I'm suggesting is to move `fencedCodeBlocks` over to base `markdown` since they belong to CommonMark by default.

The base `markdown` mode would always be non-compliant and buggy if we keep `fencedCodeBlocks` in `gfm`.

I.e., as of now, this makes using base `markdown` mode not very useful.

marijnh:
> Fore sure. But before then, what I'm suggesting is to move fencedCodeBlocks over to base markdown since they belong to CommonMark by default.

Oh, right, that's a very good idea.

dwelle:
Cool. I'll make a PR later.

PR #4888, opened 2017-08-01
[markdown mode] meta stuff

dwelle:
- __commit 1__ - _support fencedCodeBlocks in base markdown as per CommonMark_
    - Base markdown first-class support.
    - Aligned with CommonMark, thus not disable-able nor customizable `fenceChars`.
    - Moved fencedCode tests to `markdown/test.js` with minor tweaks.
    - Updated doc, (auto) removed trailing whitespace so a bit mess there.
    - Thought I'd simplify `gfm.js` after this change, but it turns out that's not yet easily possible.
- __commit 2__ - _allow to disable xml and fencedCodeBlock highlighting_
    - Added `xml` option to disable inline xml highlighting. This is a good idea to support when one doesn't want to support HTML for security reasons, while want to retain ability to highlight xml in fenced code blocks.
    - Added `fencedCodeBlockHighlighting` to enable previous behavior of not highlighting fenced code blocks even if they have language specified.
- __commit 3__ - _update doc with available options_

### DIFF
--- a/mode/gfm/test.js
+++ b/mode/gfm/test.js
@@ -14,11 +14,6 @@
   FT("doubleBackticks",
      "[comment&formatting&formatting-code ``][comment foo ` bar][comment&formatting&formatting-code ``]");
 
-  FT("codeBlock",
-     "[comment&formatting&formatting-code-block ```css]",
-     "[tag foo]",
-     "[comment&formatting&formatting-code-block ```]");
-
   FT("taskList",
      "[variable-2&formatting&formatting-list&formatting-list-ul - ][meta&formatting&formatting-task [ ]]][variable-2  foo]",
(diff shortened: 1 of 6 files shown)


mode/gfm/test.js
```diff
@@ -167,11 +137,6 @@
      "foo asfd:asdf bar");
 
   MT("notALink",
-     "[comment ```css]",
```
dwelle: permanently removed this test coz it's not a valid CommonMark codeBlock terminated... will need to update tokenizer later.

mode/gfm/test.js
```diff
@@ -167,11 +137,6 @@
...
-     "[tag foo] {[property color]:[keyword black];}",
-     "[comment ```][link http://www.example.com/]");
```
dwelle: permanently removed this test coz it's not a valid CommonMark codeBlock termination... will need to update tokenizer later.

marijnh:
Thank you!

PR #4893, opened 2017-08-03
[verilog mode] add .sv and .svh extensions

yesnault:


marijnh:
Having the same extension associated with multiple languages isn't a good idea -- I see that was already the case for .v, but that should probably be fixed too. The idea of this metadata is being able to map from an extension to a (single) mode.

yesnault:
".v" is attached to two mime, but one mode only: verilog

yesnault:
PR updated, I add sv and svh only for SystemVerilog

marijnh:
Thanks! Merged as d02b1198

Issue #4897, opened 2017-08-06
Changing editor default direction in RTL HTML document does not work


wajnberg:
In the bidi.html demo file, Change the HTML direction to RTL (<html dir="rtl">)

Then try to change the default directio of the editor.
The direction remains RTL

tomerm:
@BAMIR. FYI.

tomerm:
@amirbrans FYI.

tomerm:
@wajnberg thanks for fixing. Based on the comments in PR it looks like the fix for this issue will be available with CodeMirror 5.29.0 (expected to be released in Aug 2017).

PR #4898, opened 2017-08-06
Fix(bidi): Fix for bug /codemirror/CodeMirror/issues/4897

wajnberg:
Fix for bug https://github.com/codemirror/CodeMirror/issues/4897

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -5,6 +5,7 @@
   font-family: monospace;
   height: 300px;
   color: black;
+  direction: ltr;
 }
 
 /* PADDING */
--- a/demo/bidi.html
+++ b/demo/bidi.html
@@ -61,6 +61,11 @@
     <input type="radio" id="rtl" name="direction"/><label for="rtl">RTL</label>
   </fieldset>
   <fieldset>
+    HTML document direction:
+    <input type="radio" id="htmlltr" name="htmldirection"/><label for="htmlltr">LTR</label>
+    <input type="radio" id="htmlrtl" name="htmldirection"/><label for="htmlrtl">RTL</label>
+  </fieldset>
+  <fieldset>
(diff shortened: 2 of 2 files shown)


wajnberg:
@TOMERM FYI

tomerm:
Thanks, @wajnberg.  @amirbrans FYI.

lib/codemirror.css
```diff
@@ -270,6 +270,7 @@ div.CodeMirror span.CodeMirror-nonmatchingbracket {color: #f22;}
 .CodeMirror-widget {}
 
 .CodeMirror-rtl pre { direction: rtl; }
+.CodeMirror-ltr pre { direction: ltr; }
```
marijnh: I think you can just style `.CodeMirror pre`, _before_ the other rule, and don't need an additional class. Would that work? I'd prefer not to have another always-present class on the main container.

wajnberg:
Hello @marijnh 

I removed the rtl class following your advice.

Could you please review ?

Thank  you

marijnh:
If I set the outer page to RTL and the editor to LTR in Chrome, I'm seeing an extra, badly behaving scroll bar on the left of the editor, overlapping the gutter. That could be fixed by moving the default `direction: rtl` to the top container (`.CodeMirror`). Does that sound reasonable?

(Eventually, we'd like to support scrollbars on the left, but we're not there yet.)

wajnberg:
Hello @marijnh ,

You are right. Sorry I didn't notice that. I made the change you requested.

Could you please review ?

Thank you

marijnh:
Much better! Merged.

wajnberg:
Hello @marijnh ,

Thank you very much for having merged this change
Could you please tell me in which CodeMirror release this fix would be available?

Thank you

marijnh:
The next release (5.29.0) will be out in two weeks (I do regular releases monthly, around the 20th)

PR #4904, opened 2017-08-09
Make html-lint play more nicely with node/webpack.

geophree:
* Pass in and use the `htmlhint` module required.
* Check for `verify()` and try `HTMLHint.HTMLHint` if missing because of module nesting. it's `require('htmlhint').HTMLHint` in node (for some reason).

I'd be happy to do similar fixes to the rest of the linting, but I'm not planning on using them anytime soon.


marijnh:
Makes sense. Merged as 2add03c7

PR #4907, opened 2017-08-12
Added 'includeOrigins' configuration option

ckosmowski:
Added a configuration option to configure if origins should be included in the tern servers responses. This is useful if the results of the server should be sorted / filtererd or otherwise be affected based on which def file te completion comes from.


marijnh:
Could you generalize this to a set of properties to be added to the request (or a function that's called on the request object), so that we don't have to keep adding new options for different query parameters?

PR #4909, opened 2017-08-12
[jsx mode] Fix string interpolation with HTML string

### LINKED ISSUE #4908: [JSX] JavaScript string interpolation breaks on HTML string
Interpolated strings containing HTML seem to confuse JSX mode if an HTML tag comes immediately after the interpolated section. For example, the string ``var str = `<div>${anExpression}</div>`` ` is parsed incorrectly. 

Consider the following example:

```jsx
var React = require("react");
var ReactDOMServer = require("react-dom/server");

var innerHTML = `<pre>${JSON.stringify({ a: 123 })}</pre>`;
var myDivElement = <div 
                     className="foo" 
                     dangerouslySetInnerHTML={{ __html: innerHTML }} 
                   />;

ReactDOMServer.renderToStaticMarkup(myDivElement);
```

CodeMirror 5.28.1 on Chrome 60.0.3112.90 renders this with a syntax error on line 4: 

screen shot 2017-08-12 at 2 41 21 pm

The `</pre>` in ``var innerHTML = `<pre>${JSON.stringify({ a: 123 })}</pre>` `` should be parsed as a normal `string-2`, but because of [this line](https://github.com/codemirror/CodeMirror/blob/2add03c7efbbb1c685e7da0c4c3733778c6f35d9/mode/jsx/jsx.js#L106), CodeMirror switches from JavaScript to XML Mode and interprets the string as a dangling closing tag (and thus, a syntax error).

I've gone ahead and found [a way](https://github.com/pouwerkerk/CodeMirror/commit/63670540cd7150f6c1b243b78c0eeb7f36a65189) to stop this from happening, though maybe there's a better solution? 

You can see the `</pre>` renders correctly here on my branch:

screen shot 2017-08-12 at 2 44 36 pm

I tested and was able to reproduce this in CodeMirror 5.28.1 (and 5.15.2) on Chrome 60.0.3112.90 and Safari 10.1.2 (MacOS 10.12.6).

Thanks for looking at this—I'm a huge fan of CodeMirror!

pouwerkerk:
Fixes #4908 

This change fixes a bug where interpolated strings containing HTML in CodeMirror's JSX can cause CodeMirror to parse a string in XML mode rather than JavaScript mode.

This PR also adds a test to prevent regressions on this case in the future.

### DIFF
--- a/mode/jsx/jsx.js
+++ b/mode/jsx/jsx.js
@@ -103,7 +103,7 @@
     }
 
     function jsToken(stream, state, cx) {
-      if (stream.peek() == "<" && jsMode.expressionAllowed(stream, cx.state)) {
+      if (stream.peek() == "<" && (cx.state.tokenize.name !== "tokenQuasi" && jsMode.expressionAllowed(stream, cx.state))) {
         jsMode.skipExpression(cx.state)
         state.context = new Context(CodeMirror.startState(xmlMode, jsMode.indent(cx.state, "")),
                                     xmlMode, 0, state.context)
--- a/mode/jsx/test.js
+++ b/mode/jsx/test.js
@@ -33,6 +33,9 @@
   MT("preserve_js_context",
      "[variable x] [operator =] [string-2 `quasi${][bracket&tag <][tag foo][bracket&tag />][string-2 }quoted`]")
 
+  MT("string_interpolation",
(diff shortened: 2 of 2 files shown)


marijnh:
(This was superseded by c42322890d75a)

PR #4910, opened 2017-08-13
XQuery: Keyword lists updated (support for XQ31, XQFT, XQUP)

ChristianGruen:
Thanks for this nice and efficient library!

We are using CodeMirror in the Database Administration UI of BaseX. I have updated the keywords lists by adding all keywords of the most recent W3 XQuery 3.1, XQuery Full Text and XQuery Update recommendations, and I have removed some few obsolete/non-standard keywords.

### DIFF
--- a/mode/xquery/xquery.js
+++ b/mode/xquery/xquery.js
@@ -29,12 +29,7 @@ CodeMirror.defineMode("xquery", function() {
 
     // kwObj is what is return from this function at the end
     var kwObj = {
-      'if': A, 'switch': A, 'while': A, 'for': A,
-      'else': B, 'then': B, 'try': B, 'finally': B, 'catch': B,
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. Merged as c49bb5fed2a and followed up with 22861a03a517 (it'd be nice if you could run the tests or address failing continuous integration next time).

ChristianGruen:
Thanks for both merging the code and hinting me to the CI tests.

marijnh:
Oh, there's another test failing. You seem to have remove `private` as a keyword. Is that intentional?

marijnh:
Also `this`, `throws`, `finally`, and `null`

ChristianGruen:
There are actually no such keywords in XQuery. In XQuery 3.1, annotations (`%private`, `%public`, `%updating`), but I think these are not considered so far in the highlighter.

marijnh:
Adjusted the tests in 404d869

PR #4912, opened 2017-08-14
add greenplum mime type as gpsql

geekbosszhang:
Greenplum is based on postgresql. We added a new mime type that can recognize GPDB keywords.


marijnh:
Thanks, merged as a3970d6f

PR #4919, opened 2017-08-18
Add definitions for Spark SQL

helfer:
This PR adds a MIME definition for [Apache Spark SQL](https://spark.apache.org/sql/). Spark SQL is based on HiveQL but has some keywords that are different.

We use CodeMirror for our data science notebooks at Databricks, and having this PR merged would let us stop using our own fork and instead contribute directly to this awesome project :tada:


marijnh:
Thanks! Merged as f82a09a2

PR #4928, opened 2017-08-24
Move instantiation of diff_match_patch into getDiff()

scipper:
Move instantiation into getDiff to make merge.js bundleable. 
With var dmp = new diff_match_patch(); outside the function, it will be called during require, before diff_match_patch could be exposed to global.

### DIFF
--- a/addon/merge/merge.js
+++ b/addon/merge/merge.js
@@ -644,9 +644,11 @@
   }
 
   // Operations on diffs
-  
+  var dmp = undefined;
   function getDiff(a, b, ignoreWhitespace) {
-    var dmp = new diff_match_patch();
+    if(!dmp) {
+       dmp = new diff_match_patch();
+    }
     var diff = dmp.diff_main(a, b);
     // The library sometimes leaves in empty parts, which confuse the algorithm
     for (var i = 0; i < diff.length; ++i) {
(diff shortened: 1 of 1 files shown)


marijnh:
I'm not sure how much work the library is doing when initializing that object -- I'd prefer not to do that over and over again for every diff call.

Also, if you're going to bundle merge.js, doesn't it make sense to also bundle diff_match_patch?

scipper:
Yes, of course bundling diff_match_patch makes sense. That's what is did. But merge.js is expecting diff_match_patch to be global, and the expose of it is done after every requirement is resolved. 
an alternative to my pull request might be to call:
`var dmp = require('diff-match-patch')`
That would be the proper way of handling bundles libs, IMO.

marijnh:
> an alternative to my pull request might be to call:
> var dmp = require('diff-match-patch')

Unfortunately, diff_match_patch isn't distributed as a CommonJS module, only as a global-declaring file. (In fact, it seems like it has been unmaintained for ages, but I haven't been able to find another implementation that comes close to it in speed.) So yeah, that does complicate bundling.

Maybe update the pull request to lazily initialize the object, but only once, reusing the previous value when it exists?

scipper:
that sounds like a good solution for now. I will fix this.

marijnh:
Great. Merged as 91817cc

scipper:
Thanks!

scipper:
Can you tell, when I might expect a new release? :)

PR #4930, opened 2017-08-25
Folding for Ruby

francocatena:


marijnh:
Thank you. Merged as 35a5139d.

PR #4936, opened 2017-08-27
Fire correct hint selection event when hints are displayed. Fixes #4935.

### LINKED ISSUE #4935: Hints addon fires incorrect event when initial selected hint != 0
When displaying hints, it is possible to set the initially-selected hint by passing, for example `data.selectedHint = 2`. This works perfectly, but the final line of the `Widget` constructor fires a `select` event, unconditionally stating that the first hint was selected. This looks like a pretty straightforward mistake.

daviesian:
This fixes the issue described in #4935.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -302,7 +302,7 @@
       setTimeout(function(){cm.focus();}, 20);
     });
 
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks, merged as 883fd008

Issue #4198, opened 2016-08-24
[lint addon / javascript] Produces extraneous messages when option is specified


orionlee:
For example, for the supplied [demo](http://codemirror.net/demo/lint.html), if I modify the lint config to use lint option onUpdateListing:
```
  var editor = CodeMirror.fromTextArea(document.getElementById("code-js"), {
    lineNumbers: true,
    mode: "javascript",
    gutters: ["CodeMirror-lint-markers"],
    lint: { onUpdateLinting:  function onUpdateLinting(annotationsNotSorted, annotations, cm) {
      console.log(annotationsNotSorted);
    } }
  });
```
The `annotationsNotSorted` returned will have an extraneous annotation `"Bad option: onUpdateLinting"` that does not match the editor.

The underlying JSHINT is returning the annotation because CodeMirror addon javascript-hint is passing on non-JSHINT option to JSHINT.  See [startLinting() in lint.js](https://github.com/codemirror/CodeMirror/blob/5.18.2/addon/lint/lint.js#L137) .

Would there be any interest to accept a patch for this?

I am working on a fix, filtering out lint addon's own option before passing to mode-specific helper.

marijnh:
Don't mix the jshint options with the options to the lint addon. Instead, put them in an `options` property in the outer options object, and only those will be passed to jshint.

orionlee:
@marijnh  Ah, in that case:

- in the example above, when lint only has lint-addon option (`onUpdateLinting`), but no `options` propterty, the current [lint.js](/codemirror/CodeMirror/blob/5.18.2/addon/lint/lint.js#L137) behavior is to pass the lint-addon option object to javascript-hint helper, and ultimately to jshint.

In example code:
```
   lint: { onUpdateLinting:  function onUpdateLinting(annotationsNotSorted, annotations, cm) {
      console.log(annotationsNotSorted);
    } }
```

In [lint.js:137](/codemirror/CodeMirror/blob/5.18.2/addon/lint/lint.js#L137)
```
    var passOptions = options.options || options; // Support deprecated passing of `options` property in options
```


-  The comment in [lint.js:137](/codemirror/CodeMirror/blob/5.18.2/addon/lint/lint.js#L137)  is outdated. It states passing of `options` property is deprecated. From what you stated, it is the other wary around.

marijnh:
Yes, that's to preserve backwards-compatible behavior. Add an `options` property to work around it.

JulianPinzaru:
[...]

PR #4944, opened 2017-08-30
[lint plugins] Undeprecate passing linter rules via lint.options

westonruter:
Linters like JSHint will complain if rules are mixed with the options for the linter addon, like `onUpdateLinting` and `delay`. If the options are not separated out like this, then JSHint complains:

screen shot 2017-08-29 at 20 48 41

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -138,7 +138,11 @@
 
   function startLinting(cm) {
     var state = cm.state.lint, options = state.options;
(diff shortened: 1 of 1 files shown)


westonruter:
I believe this is advised in https://github.com/codemirror/CodeMirror/issues/4198#issuecomment-242700776

marijnh:
What a mess. I'm not sure why I deprecated that anymore. But yeah, undeprecating is probably the right call. Merged as 5ef6dfc3

PR #4945, opened 2017-08-30
Add basic STAN support

Arvinds-ds:
Adding basic syntax highlighting for the stan programming language (mc-stan.org)


marijnh:
Hi. We're not accepting new modes into the main distribution anymore, to keep the maintenance burden manageable. The recommended way to distribute a new mode is to put it in on NPM in a separate module -- see the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir) as an example.

PR #4948, opened 2017-09-01
PromQL (prometheus.io) language

DeyV:
Very simple, based on Go, PromQL language implementation


marijnh:
We aren't accepting new modes into the main distribution anymore. Please distribute modes as separate NPM packages, as the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir) does.

PR #4950, opened 2017-09-03
Added gobstones lang mode

aguspina:
Added [Gobstones](http://www.gobstones.org) lang mode support.


marijnh:
We aren't accepting new modes into the main distribution anymore. Please distribute modes as separate NPM packages, as the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir) does.

Issue #4953, opened 2017-09-05
Closebrackets with custom keys


blittle:
I might be missing something, but it doesn't look like the `closebrackets` plugin supports custom characters because it doesn't update the `keyMap` with custom pairs (https://github.com/codemirror/CodeMirror/blob/master/addon/edit/closebrackets.js#L27)

Essentially I'm trying to make codemirror work with a custom language where the comments exist between two `#` characters. I want to add `#` as a closebracket pair for more easily entering text like `# my comment #`

marijnh:
Right, the set of keys to bind is static at the moment, because (due to nested modes) it is not possible for the plugin to see in advance which characters are defined for every mode it will encounter. One solution might be to add a utility function to the addon that allows user code to add new characters to the keymap. Do you want to implement that?

blittle:
@marijnh sure, I'll work out a PR in the next few days.

blittle:
Fixed with https://github.com/codemirror/CodeMirror/commit/e716c17e3bdfd10da477d92bd3fa2fb66cabbaa1

PR #4956, opened 2017-09-08
Add an extension for extending the keymap for matchbrackets

blittle:
Fixes https://github.com/codemirror/CodeMirror/issues/4953 with a new extension (`addBracketsKeyMap`) for adding new characters to the key map.

### DIFF
--- a/addon/edit/closebrackets.js
+++ b/addon/edit/closebrackets.js
@@ -18,6 +18,8 @@
   var Pos = CodeMirror.Pos;
 
   CodeMirror.defineOption("autoCloseBrackets", false, function(cm, val, old) {
+    var keyMap = getKeyMap();
+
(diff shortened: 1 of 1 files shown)


addon/edit/closebrackets.js
```diff
@@ -28,17 +30,27 @@
     }
   });
 
+  CodeMirror.defineExtension("addBracketsKeyMap", function(extraPairs) {
```
blittle: I might need to be doing a `this.removeKeyMap` before `this.addKeyMap`

marijnh:
Yeah, this will leak keymaps. Could you take a look at patch e716c17e, which allows you to pass a set of pairs as an option, and the addon will make sure they are bound, works for you?

blittle:
LGTM, Thank you!

PR #4957, opened 2017-09-08
fixed bug in mark-selection

callodacity:
Found a minor issue with the `mark-selection` addon.

### To reproduce:

* Go to http://codemirror.net/demo/markselection.html
* Enter at least 14 lines of text
* Create a selection starting inside line 6, ending inside line 14
* Press <kbd>Shift</kbd> + <kbd>←</kbd> (i.e., move selection one char to the left)

An error will appear in the console at this stage, and the selected text will also go from white to black.


marijnh:
Thank you! Merged as 850a92fb

PR #4958, opened 2017-09-11
Add functionality to provide aria-label

ssanaul:
The textarea generated by CodeMirror should be attributed with an aria-label to support people with disabilities and comply with web accessibility standards. 
If there is a more elegant way to fix the problem, please let me know.

### DIFF
--- a/src/edit/fromTextArea.js
+++ b/src/edit/fromTextArea.js
@@ -53,7 +53,7 @@ export function fromTextArea(textarea, options) {
       }
     }
   }
-  
+
   let labelNewTextarea = cm => {
     let cmTextarea = cm.display.input.textarea
     if (textarea.id) cmTextarea.setAttribute('aria-label', textarea.id)
--- a/src/edit/fromTextArea.js
+++ b/src/edit/fromTextArea.js
@@ -53,9 +53,15 @@ export function fromTextArea(textarea, options) {
       }
     }
   }
+  
+  let labelNewTextarea = cm => {
+    let cmTextarea = cm.display.input.textarea
+    if (textarea.id) cmTextarea.setAttribute('aria-label', textarea.id)
+  }
 
   textarea.style.display = "none"
   let cm = CodeMirror(node => textarea.parentNode.insertBefore(node, textarea.nextSibling),
     options)
+  labelNewTextarea(cm)
   return cm
 }


westonruter:
Should this be `aria-owns` or `aria-controls` or `aria-labelledby` instead of `aria-label`?

ssanaul:
@westonruter The label is derived from the id of the old textarea, which is hidden by CodeMirror. 

`aria-labelledby` is meant for indicating a visual relationship between a label and form component, so it doesn't apply in this case. 
`aria-controls` would indicate to a screen reader that keyboard focus could be moved to the hidden textarea, but input should really go into the newly generated textarea. 
Similarly `aria-owns` would indicate that the new textarea is a parent of the old textarea, when really, it replaces it (visually, at least, which is what we're concerned with).

marijnh:
This will break when the editor uses `contenteditable` [input style](http://codemirror.net/doc/manual.html#option_inputStyle) (which is recommended for situations like this anyway, since it's much more accessible). Does patch bde68f12 work for you?

ssanaul:
@marjinh Ah, yes- that's much cleaner.

marijnh:
Great!

afercia:
Sorry to bother but this should be reverted. The [aria-label property](https://www.w3.org/TR/wai-aria/states_and_properties#aria-label) takes a string (not an ID) that should be the name of the textarea. It basically works like a `<label>` element, and gives the textarea a name.

Instead, `aria-labelledby` takes an ID and it's handy when a visible element already present in the page can be referenced as "name" of a form control.

Thus, `aria-label` should be a name that identifies the purpose of the textarea. Since the textarea can be used for anything, I'm afraid there isn't a name that covers all the possible uses. Maybe, consider to pass the aria-label value from the config so users can set a meaningful name that fits their use case.

For reference, making screen readers announce the textarea ID is not so useful:

screen shot 2017-09-11 at 14 46 17

afercia:
Worth noting, authors can always use a standard `<label>` element to give the textarea a name:

```
<label for="demotext">Edit your code</label>
```

marijnh:
Reading the spec, yes, I definitely agree. Reverted in attached patch.

jongund:
I would like to see three labeling options supported by codemirror.  Note only one of these options should be used for a form control.

1. Add an explicit label using by passing in a string to generate ```aria-label="_label text_"``` on the visible textarea.
1. Add a reference from a ```<label for="id_label1">_label text_">``` and the visible textarea now has ```id="id_label1"```
1. Add a reference to some text on the page using that has a ID and the visible textarea would have ```aria-labelledby="id_text1"```

phuongkt:
.

PR #4967, opened 2017-09-14
Update gfm.js

ParthKolekar:
Remove SHA hash from being treated specially.

Now this will require 

// User/Project@SHA
// User@SHA

The existing regex matches arbitrary code that is a series of `[a-f0-9]{7,40}` not matter how they appear in a sentence.

### DIFF
--- a/mode/gfm/gfm.js
+++ b/mode/gfm/gfm.js
@@ -81,10 +81,9 @@ CodeMirror.defineMode("gfm", function(config, modeConfig) {
       if (stream.sol() || state.ateSpace) {
         state.ateSpace = false;
         if (modeConfig.gitHubSpice !== false) {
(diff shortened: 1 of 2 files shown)


ParthKolekar:
Oh. Apparently github has this *issue* too where it marks anything with `[a-f0-9]{7,40}` as special.

ParthKolekar:
Github uses CodeMirror. So it will obviously exhibit this behaviour. That was stupid.

marijnh:
GitHub does allow linking of stand-alone hashes (as in 67379b827d111c97), so I don't think this is correct.

ParthKolekar:
Agreed. This pull request is incorrect.

I didn't know about this feature that allows commit messages to become
valid links.

Issue #4926, opened 2017-08-22
Incorrect/no selection of Arabic(or Arabic with Latin) text


nethip:
Upon selecting Arabic Text, the selections are not getting shown/sometimes wrong selections are getting shown. Please refer to the GIF video below. I think the problem is with the selections not been shown rather than the selections not happening at all. Also what I have observed is that this is happening more with word wrap on.

![screenrecording](https://user-images.githubusercontent.com/1577184/29563002-5a9a3da4-8759-11e7-9cad-ecd3de45336d.gif)

marijnh:
Could you test with attached patch?

nethip:
Thanks @marijnh for the patch. We have tested your patch and the selections are much better. However, we have seen some selection issues. 

1. Unable to select the last character without using the keyboard. Please refer to the video below.

![point 1 - unable to select the last character without using the keyboard](https://user-images.githubusercontent.com/1577184/29968030-673dc63a-8f37-11e7-8bc7-b22501522421.gif)

2. Unable to select text from right side. 

![point 2 - unable to start the selection from the right side](https://user-images.githubusercontent.com/1577184/29968041-785e449e-8f37-11e7-89ca-22ad2a60f328.gif)

3. Unable to select just text without selecting the opening and closing tags

![point 4 - unable to select just text without selecting the opening and closing tags](https://user-images.githubusercontent.com/1577184/29968127-db44dfa0-8f37-11e7-9729-167f385a92c5.gif)

marijnh:
These all seem to boild down to the fact that many screen positions will correspond to two logical text positions, and if you have text like RRRLLL (where R is rtl and L ltr), selecting from before the first R to after the second will span those two R's, but selecting from that position (which, in the string order, also touches the first L) to in between the L's will get you a region that only covers the Ls.

Some editors seem to do something clever in the first case, apparently swapping the anchor and head of the selection during selecting in some circumstances, but it isn't clear to me how that would generally work.

The second and third seems to work that way in native textareas as well, and are really not avoidable -- a selection is a continuous range in the _string index_ space, not the screen space, so yeah, if you cross bidi boundaries you're not necessarily going to see a coherent selected block on the screen. CodeMirror is definitely not going to change how selections work (and I'm now aware of any tool that does this, and as far as I know this is the expected behavior for RTL users).

nethip:
Thanks @marijnh for detailed explanation.

[...]

PR #4968, opened 2017-09-14
Rewrite coordsCharInner

marijnh:
Hi @adrianheine, do you have time to take a look at this? Since you were the last to think about this stuff, you might be able to spot mistakes that I didn't notice.

I basically made `coordsCharInner` for bidi lines run two binary searches, one on the bidi spans and one on the actual characters of a single span. It's conceptually simpler, and probably faster (though I didn't benchmark). Since the other algorithm was found to be running into its infinite loop guard in some cases (see issue #4926), I figured it'd be better to start over rather than try to salvage that.

The main question is whether doing a binary search on the bidi spans makes sense on wrapped lines. I think it does, because as we also found with `wrappedLineExtent`, even when an rtl span is wrapped, as a whole it remains inside the computed ordering.

### DIFF
--- a/src/measurement/position_measurement.js
+++ b/src/measurement/position_measurement.js
@@ -382,7 +382,7 @@ export function cursorCoords(cm, pos, context, lineObj, preparedMeasure, varHeig
   if (!order) return get(sticky == "before" ? ch - 1 : ch, sticky == "before")
 
   function getBidi(ch, partPos, invert) {
-    let part = order[partPos], right = (part.level % 2) != 0
+    let part = order[partPos], right = part.level == 1
     return get(invert ? ch - 1 : ch, right != invert)
   }
   let partPos = getBidiPartAt(order, ch, sticky)
@@ -440,10 +440,10 @@ export function coordsChar(cm, x, y) {
 }
 
 function wrappedLineExtent(cm, lineObj, preparedMeasure, y) {
-  let measure = ch => intoCoordSystem(cm, lineObj, measureCharPrepared(cm, preparedMeasure, ch), "line")
+  y -= widgetTopHeight(lineObj)
   let end = lineObj.text.length
-  let begin = findFirst(ch => measure(ch - 1).bottom <= y, end, 0)
-  end = findFirst(ch => measure(ch).top > y, begin, end)
+  let begin = findFirst(ch => measureCharPrepared(cm, preparedMeasure, ch - 1).bottom <= y, end, 0)
+  end = findFirst(ch => measureCharPrepared(cm, preparedMeasure, ch).top > y, begin, end)
   return {begin, end}
 }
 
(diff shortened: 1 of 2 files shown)


adrianheine:
Whoa, I'll look into it. I don't really understand your last sentence, though. With wrapped lines, a single visual line always contains a _logically_ continuous number of chars. Inside a visual line, all bidi spans that are present are __visually__ ordered as in the bidi order. There might be spans missing, though. I was thinking about using the bidi spans instead of `moveVisually` yesterday as well.

src/measurement/position_measurement.js
```diff
@@ -446,66 +452,106 @@ export function wrappedLineExtentChar(cm, lineObj, preparedMeasure, target) {
...
+    ltr = part.level % 2 == 0
+    // If this isn't the first part, the part's start is also after
+    // the coordinates, and the coordinates aren't on the same line as
+    // that start, move one part back.
+    if (index > 0) {
+      let start = cursorCoords(cm, Pos(lineNo, ltr ? part.from : part.to, ltr ? "after" : "before"),
+                               "line", lineObj, preparedMeasure)
+      if (boxIsAfter(start, x, y, true) && start.top > y) {
+        part = order[--index]
+        ltr = part.level % 2 == 0
```
adrianheine: This is wrong in cases with lvl 2, I think. Example:

http://unicode.org/cldr/utility/bidi.jsp?a=abcd+%D9%85+1+%D8%B1&p=LTR&hack=on

If your line fits 6 characters, according to [UBA](http://unicode.org/reports/tr9/#Reordering_Resolved_Levels), it will be  
```
012345
------
abcd E 
N F
```

But the bidi ordering is:

```
"{"level":0,"from":0,"to":5}, {"level":1,"from":8,"to":10}, {"level":2,"from":7,"to":8}, {"level":1,"from":5,"to":7}"
```

The first bidi part in bidi ordering order that's after, say, the right end of `E` will be `order[1]`. I suppose you are looking for `order[3]` though?

marijnh: The unicode.org link isn't loading right now, so I'm not sure what you mean, but I guess you're saying that, in the presence of wrapped lines with level 2 spans in them, the visual ordering may not correspond to the ordering in `order`? Experiments seem to conform that. That's bad!

I'll see if I can work around this by doing the clipping to a single wrapped line _before_ finding the appropriate bidi part (discarding those outside of the line's range). I guess that makes applying a binary search a lot harder, but I suppose falling back to a full search in the wrapped case (and discarding all spans that aren't in the line) should be okay.

adrianheine: Hm, it's loading for me. Anyway, what I am saying is: There are cases in which a span `A` that comes before a span `B` _in the bidi order_ comes after the span `B` _in top-down order_.

marijnh:
Yeah, I see your point, I've added a test case for that and fixed it the way I described (different search for bidi spans when in wrapped content).

The total complexity of this is now at least as bad as the old algorithm (though I added more comments so the line count isn't representative), but I think the new code is a little easier to reason about and hopefully more robust.

marijnh:
I've gone ahead and merged this as 68b4f44e3c6e, but if you have any further concerns definitely let me know.

PR #4971, opened 2017-09-14
[mode/simple addon] Allow capturing with single tokens

### LINKED ISSUE #4970: mode/simple should ignore capture groups if tokens is a string
hello,

using the [simple mode](http://codemirror.net/demo/simplemode.html), one of my regexes needed capture groups for [Backreferences](http://www.regular-expressions.info/backref.html), however, the entire regex should've counted as a single token.

example:
```js
{regex: /(["'])((?:(?!\1).)*)\1/, token: "string"}
```

what I expected:

since `token` is not an array, the entire regex (`matches[0]`) should be classified as `string`.

what actually happened:

`"` (first quote) classified as `cm-s`, content as `cm-t`. basically for `matches[n]`, the nth character of the string is taken as the token.

my suggestion:

change line [139](https://github.com/codemirror/CodeMirror/blob/master/addon/mode/simple.js#L139) of `simple.js` to check for type of token:
```diff
139c139
<           if (matches.length > 2) {
---
>           if (matches.length > 2 && typeof(rule.token) !== "string") {
```

this fixes the problem without breaking the multi token functionality, and IMHO checking if `token` is an array or a string is a better way than checking if regex contains capture groups for this.

adrianheine:
Closes #4970. Even if you beat me to it I still like my commit better :P And I have arguments: It also works with `token: null` and fixes the documentation.

### DIFF
--- a/addon/mode/simple.js
+++ b/addon/mode/simple.js
@@ -136,7 +136,7 @@
             state.indent.pop();
           var token = rule.token
           if (token && token.apply) token = token(matches)
-          if (matches.length > 2 && typeof rule.token != "string") {
+          if (matches.length > 2 && rule.token && typeof rule.token != "string") {
             state.pending = [];
             for (var j = 2; j < matches.length; j++)
               if (matches[j])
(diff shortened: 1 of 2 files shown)


marijnh:
Oh, docs are always nice, but why the extra check? `typeof` definitely won't return `"string"` for null or undefined.

adrianheine:
Exactly. `typeof` won't return `string`, so `null` or `undefined` will be considered an array.

marijnh:
Hah, yeah, I should stay away from my laptop after 20:00. Thanks for the fix, merged!

PR #4974, opened 2017-09-16
gfm: Fix false detection as a hexadecimal number.

aero31aero:
In cases like `defamed` or while typing `feedbac`, gfm-mode recognizes the token as hex. This was because of the setting to highlight such patterns of length 7-40, most likely to easily identify commit hashes. However, this causes a false detection as faced by some users at https://github.com/jbt/markdown-editor/issues/72 .


marijnh:
GFM does not use hash marks to identify commits, so this breaks highlighting of commit hashes. See patch e5a13e62 instead.

PR #4978, opened 2017-09-18
[soy mode] Support @param declarations in comments

vrana:
@param specified in a template comment is older but still supported way of
declaring a parameter. The comment must directly precede {template}.

Reference: https://developers.google.com/closure/templates/docs/concepts#namespace-decl


marijnh:
Thank you. Merged as ed645438!

PR #4982, opened 2017-09-19
Support esper SQL dialect

frbor:
Hi,

I see that you do no accept new modes, but is it possible to merge this addition to the sql-mode (x-esper)?

Esper is a language for complex event processing, but the language conforms to the SQL standard:

http://www.espertech.com/esper/faq_esper.php#how-does-it-work-overview

I hope you will consider this pull request as it will be a lot more work to create a separate package for this mode, since most of the logic is already included in sql.js.

best regards,

   Fredrik


marijnh:
Sure, that's small enough to integrate. Merged as 282cf366

PR #4983, opened 2017-09-19
Vue mode: support more style and script types

ficristo:
I added the second commit mostly to support `<script type="text/ecmascript-6">` which, from what I understood, was necessary in some old version of Webstorm.
Feel free to merge only the first commit.


marijnh:
Looks harmless enough. Merged as 60f39516

ficristo:
Thank you.

PR #4987, opened 2017-09-21
Tags option for PHP mode

nabab:
Sends the property tag from parserConfig to mixedhtml parser


marijnh:
That forces the HTML mode to be htmlmixed rather than whatever text/html is bound to, removing flexibility.  I was thinking something more along the lines of ad9732ae

PR #4992, opened 2017-09-23
Add dot custom mark in vim mode

jkaplon:
Add tick-dot and backtick-dot functionality to vim mode.


marijnh:
Thanks! Merged as bb0e1c3f

PR #5001, opened 2017-09-27
[vim] Fix crash on exiting insert mode.

alur:
Happened if you entered visual block mode in one editor and then tried to exit insert mode in another, since lastChange comes from vimGlobalState.


heppe:
looks good to me

PR #5003, opened 2017-09-29
Remove `package` from Kotlin def keywords

andrewda:
`package` should be treated like `import`, which is not a def keyword.

| Before | After |
| :-----: | :----: |
| before | after |

Note that the package at the top in the `After` image is correctly highlighted.


marijnh:
Thanks, merged as 5fd2133b

PR #5017, opened 2017-10-09
Add language mode for the ICU message format

Janpot:


marijnh:
We are not accepting new language modes into the main distribution. The recommended way to distribute new modes is to create a separate repository and publish them as separate NPM modules with a name like codemirror-mode-icu.

Janpot:
Thanks, Seems like I didn't read the contributing guidelines thoroughly enough. That is a great idea, it already seemed an unmaintainable practice to me to keep it in the main repo. And it will allow me to keep a coding style that is more comfortable to me.

Janpot:
@marijnh Would love to be able to use `test/mode_test.js` in my own package. Any way this can be exported somehow, or be unignored, or published as a separate module?

marijnh:
Good question. It's not very well-factored or easy to reuse at the moment, and we're hoping to move to a proper test running in the near future, so I'd say just copy it and change it to match your tests. Eventually, it'd be nice to have this as a reusable module, I agree.

PR #5031, opened 2017-10-16
Editor creates red NL characters on Windows when typing [ or {

guiguan:
screen shot 2017-09-04 at 12 23 51 pm
screen shot 2017-09-04 at 12 24 04 pm


marijnh:
Thanks. Merged as e5e2aefd. I took out the call to `repeat` since we're still targeting older JS runtimes that don't have this method.

PR #5034, opened 2017-10-17
Include CodeMirror.innerMode in node runmode

niik:
The markdown mode [uses `CodeMirror.innerMode`](https://github.com/codemirror/CodeMirror/blob/master/mode/markdown/markdown.js#L268) which isn't defined in the stripped down node runmode.

Since markdown is the only mode (AFAICT) that uses it I'm not sure if it would make more sense to rewrite the mode to not use it but this seemed like the least risky option to me.

### DIFF
--- a/addon/runmode/runmode.node.js
+++ b/addon/runmode/runmode.node.js
@@ -163,6 +163,18 @@ exports.getMode = function(options, spec) {
 
   return modeObj;
 };
+
+exports.innerMode = function(mode, state) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks, merged as deaa842d. Now that the core library is modular we really shouldn't be duplicating things like this anymore, but no one has gotten around to setting up a build process using the main sources for runmode.node.js yet

PR #5035, opened 2017-10-17
[FIX] wrong representation of table hinting in sql demo page html

vtripolitakis:
The current JS hint options for table hinting, raised errors. This was misleading for newcomers like me, as there's no documentation on that, other than the code. The proper format is like this:

```
hintOptions: {tables: {
      users: ["name", "score", "birthDate"],
      countries: ["name", "population", "size"]
    }}
```


marijnh:
Thanks. Merged as 42a26328

vtripolitakis:
Thanks 🎉

PR #5037, opened 2017-10-18
Update my name in AUTHORS

pjdelport:
Just a small change; I go by Pi now. :)


marijnh:
Merged, and followed up with cf799958 to make sure the script that adds all names from the git history to that file on release doesn't re-add the old name.

pjdelport:
Ah, I didn't see the generation script; thank you for updating that too!

Issue #5030, opened 2017-10-15
Markdown List - Update Following Numbers


stuajnht:
Hi,

As mentioned in [issue #959 from Boostnote](https://github.com/BoostIO/Boostnote/issues/959), when a new item is added to the middle of a Markdown list, the following numbers are not modified. Having the numbers auto-update would be useful.

I am happy to have a look at this, if it is something you think would be useful for CodeMirror.

marijnh:
I'd be interested in that as a pull request, but since Markdown doesn't care, I'm not going to work on it myself. (Also, if you implement this, make sure it leaves lists that use `1.` everywhere intact—if people are intentionally not numbering, the addon probably shouldn't get in the way.)

PR #5044, opened 2017-10-21
[continuelist addon] Increment numbers when item is added to the middle

stuajnht:
Per #5030, when adding new items to the middle of a Markdown list, the remaining list numbers should automatically increment.


marijnh:
Looks good. Merged as 97290a68

PR #5045, opened 2017-10-21
[ada mode] Add

gusthoff:


marijnh:
Hi. We're no longer accepting new modes into the main distribution. The recommended way to distribute new modes is to maintain them in their own repository and publish them to NPM as separate packages—see for example the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir).

PR #5060, opened 2017-10-31
Bug in search with case folding

neochief:
I think the change is self-explanatory. It's probably a bug caused by the recent refactoring.


marijnh:
Thank you. Merged as a7907d37

PR #5069, opened 2017-11-06
Remove broken links in the realworld users page

jayaprabhakar:
Removing the broken links in the realworld users list. I checked the backlinks using,

https://www.deadlinkchecker.com/website-dead-link-checker.asp

and manually verified these urls to be invalid.


marijnh:
Thank you!

PR #5076, opened 2017-11-08
Fully qualify import paths

klebba:
By adding the file extension to every imported module in CodeMirror's source, native ES6 imports will work properly in a browser without a build pipeline. This will have no impact on the current build process.

Working example:

```
<!doctype html>
<html>
  <head>
    <meta charset="UTF-8">
  </head>
  <body>
    <script type="module">
    import CodeMirror from 'https://klebba.github.io/CodeMirror/src/codemirror.js';
    console.log(new CodeMirror());
    </script>
  </body>
</html>
```

ES6 imports are currently supported in Chrome and Safari, and in development for Firefox.


marijnh:
@adrianheine Do you have an opinion on this?

adrianheine:
I'm in favor of this.

marijnh:
Great. Merged!

adrianheine:
btw, I know we are not using eslint in this project, but if anybody ever wants to do something like this in a project that uses eslint, [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import/blob/HEAD/docs/rules/extensions.md) can check this.

marijnh:
Yeah, that is valuable. Ingvar got me to adopt it in Acorn, but it is so slow. Ideally, I'd want to add import checking to blint, but I'm not sure when I'll get to that.

klebba:
Thanks for merging 👍

klebba:
@marijnh -- are the `mode/*` and `addon/*` files generated from some other original source which uses `import` syntax? I would be happy to change them if so; otherwise I am forced to copy them to my project and remove this boilerplate by hand:

```
(function(mod) {
  if (typeof exports == "object" && typeof module == "object") // CommonJS
    mod(require("../../lib/codemirror"));
  else if (typeof define == "function" && define.amd) // AMD
    define(["../../lib/codemirror"], mod);
  else // Plain browser env
    mod(CodeMirror);
})(function(CodeMirror) {
"use strict";

//...

});
```

adrianheine:
They are not generated. We are not accepting new modes and addons into the repo and might package the existing ones separately, but as far as I know we have no current plans to do so but rather keep them as they are.

klebba:
Ok, understood. So if I were to attempt any kind of refactoring for modes and addons it would need to be completely transparent to current users. Would you be open to a PR of this nature?

marijnh:
> native ES6 imports will work properly in a browser without a build pipeline

As an aside, if CodeMirror had even a single NPM dependency, this would fall apart. And the library is likely to start using external modules in the future. Are you aware of any solution to address that?

klebba:
In this scenario (adding a project dependency which uses `require`) I would currently have to build a bundled artifact that has `import` compatibility and use that in order to use native ES6 loading. Not a great scenario because I can't deduplicate your child dependencies with my own once `codemirror` is packed.

For what it's worth, the fact that `codemirror` has no external dependencies is a strength of the library in my opinion.

Either way, I'm not sure what it will take to converge nodejs and browser module loading, but the current situation is ugly. Here's some more info: http://2ality.com/2017/09/native-esm-node.html

PR #5083, opened 2017-11-10
[soy mode] Support comments in all contexts

vrana:
Soy supports comments in all contexts but CodeMirror parses them only outside tags. This adds the support to the tags too.


marijnh:
Thanks! Merged.

PR #5090, opened 2017-11-17
[javascript mode] Highlight type in generic call

### LINKED ISSUE #5048: [typescript] type not highlighted in <>s
```typescript
this.a<Type>(foo);
```
`Type` should be highlighted

adrianheine:
Closes #5048.


marijnh:
This looks like it _completely_ overrides the meaning of the `<` operator in TypeScript mode—which is probably not good, since as far as I know that operator still exists in that dialect.

adrianheine:
Thanks for taking over :)

Issue #4784, opened 2017-06-07
Searching for arabic words


Ashamandi:
Hi,
Searching function does not find all matched words in case of Arabic words, it fails to match some letters like " ا  , أ , آ" and "ى ,  ي " which are considered equal and have to be matched

![image](https://user-images.githubusercontent.com/17547411/26867525-00c380e4-4b66-11e7-8095-e77e698724dc.png)

I have tested through Chrome v 58.0 and firefox developer edition v 54.0 (64 bit)

Thank you!

marijnh:
Do I understand correctly that this is about unicode string normalization? I.e. two different strings denoting the same text, which should be [normalized](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) before comparing them?

Ashamandi:
Yes, right that`s what I meant

marijnh:
It appears that most browsers support [`String.normalize`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) nowadays, so it might be possible to make the search code normalize strings. The difficult thing there is associating offsets into the normalized string (which is what you'd get from searching after normalizing) with offsets in the original string (which you need to point at the match in the editor). There's already something like that happening for case-folding, but that can be done one character at a time, which isn't the case for unicode normalization.

If somebody wants to work on this, I'd be very grateful.

marijnh:
(Maybe, though I'm not sure yet, normalizing to `NFD` form always decomposes code points, never combines them, so that might be safe to do one character at a time.)

marijnh:
Attached patch introduces normalization using `String.prototype.normalize`, but while that does help with things like precomposed characters vs extending characters, it doesn't appear to have any effect on the cases you describe -- it doesn't normalize these characters, and it doesn't appear that JavaScript comes with a built-in way to do that.

One option would be to allow code to pass a custom normalization function, and to write language-specific normalizers (or maybe a good one already exists -- I wasn't able to find it). But it wouldn't be a drop-in fix -- you'd have to explicitly set it up. Would that help for you?

Ashamandi:
Yes, you are right. we should add a custom normalization function for that I will try to implement it as npm package, is it will be ok?

marijnh:
[...]

PR #5099, opened 2017-11-28
Searching for Arabic words issue - #4784

Ashamandi:
Hello @marijnh, this PR to fix #https://github.com/codemirror/CodeMirror/issues/4784
at this commit I added two functions one for normalize Arabic strings and the other to check ifArabic

I am expecting to apply the normalization on the search input and the doc itself .. Could you help me figure out where exactly can I apply it to normalize the search input and the whole doc itself before start searching?

### DIFF
--- a/addon/search/searchcursor.js
+++ b/addon/search/searchcursor.js
@@ -114,11 +114,43 @@
       }
     }
   }
+  
+  //Normalization for Arabic
+  var normalizeArabicChars = function (s) {
+    function filter(c) {
+      // ALEF Chars
+      if ((c =='\u0625') || (c =='\u0623') || (c =='\u0622') || (c =='\u0675') ||
+          (c =='\u0673')|| (c =='\u0672') || (c =='\u0671') || (c =='\u0670') ||
+          (c =='\u0674') )
+        return '\u0627'
+      // TAAA MARBOTA Chars
+      else if ((c =='\u06c3') || (c =='\u06c2') || (c =='\u06c0') || (c =='\u06c1') ||
+            (c =='\u0647') )
+        return '\u0629'
+      // YAAA Chars
+      else if ((c =='\u0649') || (c =='\u064a') || (c =='\u0678') || (c =='\u064a') )
+        return '\u0620'
+      else if (c =='\u0626')
+        return '\u0620\u0621'
+      else
+        return c
+    }
+    var normalized = "", i, l
+    for (i = 0, l = s.length; i < l; i = i + 1) {
+      normalized = normalized + filter(s.charAt(i))
+    }
+    return normalized
+  }
+
+  function isArabic(str) {
+    var pattern = /[\u0600-\u06FF\u0750-\u077F]/;
+    return pattern.test(str);
+  }
 
   var doFold, noFold
   if (String.prototype.normalize) {
-    doFold = function(str) { return str.normalize("NFD").toLowerCase() }
-    noFold = function(str) { return str.normalize("NFD") }
+    doFold = function(str) { return isArabic(str)? normalizeArabicChars(str) : str.normalize("NFD").toLowerCase() }
+    noFold = function(str) { return isArabic(str)? normalizeArabicChars(str) : str.normalize("NFD") }
   } else {
     doFold = function(str) { return str.toLowerCase() }
     noFold = function(str) { return str }


marijnh:
> Could you help me figure out where exactly can I apply it to normalize the search input and the whole doc itself before start searching?

I don't understand this question—why do you need to normalize before you start searching?

It appears that the normalization is strickly 'expanding' (never shortens the length of a string) after all, so that's good. The code looks good except that you're turning off case- and unicode-normalization when there's Arabic in the string. The linter appears to be complaining about trailing whitespace.

Ashamandi:
> I don't understand this question—why do you need to normalize before you start searching?

what I think about is normalizing both search input and the whole doc to put them into same char shaping before start searching so the searching can detect all matches 

as an example :

searching

What I imagine solving that is 
1- Normalize search input أيه to "ايه"
2- Normalize the whole doc so all its 4 words will be converted to "ايه" as well
3- Run searching function 
the output will be as shown before.

but when I applied it on noFold and doFold only this didn`t solve the problem, the normalization done but not all matches detected

marijnh:
> so the searching can detect all matches

You mean for the purpose of highlighting them? That is done in `searchOverlay` in `addon/search/search.js`, which currently doesn't really do any normalization beyond case, for which it is relying on the regexp `i` flag. To support normalization there, we'd have to use a different approach.

Can we start by landing this change just for the search cursor, and then look into the highlighting issue separately?

Ashamandi:
> Can we start by landing this change just for the search cursor, and then look into the highlighting issue separately?

Sure ... no problem for that.


> The code looks good except that you're turning off case- and unicode-normalization when there's Arabic in the string. 

Sorry didn`t get what you mean here, Do you mean that you prefer chars instead of its unicodes like that 

```
//Normalization for Arabic
  var normalizeArabicChars = function (s) {
    function filter(c) {
      // ALEF Chars
      if ((c =='إ') || (c =='أ') || (c =='آ') || (c =='ٵ') ||
        (c =='ٳ')|| (c =='ٲ') || (c =='ٱ'))
        return 'ا'
    }

```

marijnh:
> Sorry didn`t get what you mean here

In this line (and the one after it)...

    doFold = function(str) { return isArabic(str)? normalizeArabicChars(str) : str.normalize("NFD").toLowerCase() }

... `toLowerCase` won't be called for a string like `"XYايه"`, but it should (in addition to `normalizeArarabicChars` and `.normalize`)

Ashamandi:
ooooh ... I got what you mean, so I think there is no need to check isArabic before applying the normalization  
```
if (String.prototype.normalize) {
    doFold = function(str) { return normalizeArabicChars(str.normalize("NFD").toLowerCase()) }
    noFold = function(str) { return normalizeArabicChars(str.normalize("NFD")) }
```

Will it be okey?

marijnh:
Since `normalizeArabicChars` looks somewhat expensive, I'd prefer if you kept the check. Maybe something like `str = str.normalize("NFD").toLowerCase(); if (isArabic(str)) str = normalizeArabicChars(str); return str`?

addon/search/searchcursor.js
```diff
@@ -115,13 +115,61 @@
...
+        case 'ۃ' :
+        case 'ہ' :
+          return 'ة'
+        // YAAA Chars
+        case 'ى' :
+        case 'ي' :
+        case 'ٸ' :
+          return 'ي'
+        case 'ئ':
+          return 'ي ء'
```
marijnh: Is this space in the returned string intentional?

marijnh: Also, `.normalize("NFD") already separates this character into \u064a and \u0654, which seems similar to what you're doing, and might already cover this?

Ashamandi: Thanks for comment .... yes, you are right, it seems that when I converted each Unicode into its char I added a space by mistake .... sorry for that.
About normalize("NFD"), Actually I tested it but it didn`t work , but for sure if you test it, I can remove this case to reduce checks.

marijnh:
Is there a standard or document that your normalizations are based on? Whether something works or doesn't work depends on what you expect it to do—and if we're going to implement a normalization, I'd prefer for it to be based on some standard or at least widespread convention.

marijnh:
Does the problem you are trying to solve also occur with diacritics encoded as separate characters (for example `\u0654` "hamza above")? If so, we probably need something more general than these replacements—filtering out a whole character categories, such as accents and diacritics. I notice a lot, but not all, of the characters in your code are already decomposed to a base character plus an extending character when you call `.normalize("NDF")` on them, so that could be part of the solution.

But, I guess, not the entire solution. Is there some Unicode concept that covers the rest?

Ashamandi:
> Does the problem you are trying to solve also occur with diacritics encoded as separate characters (for example \u0654 "hamza above")? 

No, my problem is limited to some chars that are considered equaled but has different Unicode chars
and these chars are AL-ALEF, AL-TAA AL-MARBOTA and AL-YAA.
and if we talk about Al-ALEF as an example:   ا, أ , آ, إ ,.. we found that all the previous ALEF shapes should be detected when searching for any one of them. 
in the same way when we take about TAAA MARBOTA and YAA different shapes.

marijnh:
I asked around about whether these characters should be considered equal, and opinions differ (see [this](https://mobile.twitter.com/MarijnJH/status/938719243687944197) and [this](https://mobile.twitter.com/aGharbeia/status/940101031219757056)).

I'd like to add some kind of generalized stripping of extending characters (so that 'e' matches 'é' and 'ٵ' matches 'ا'), which would address some of the equivalences in this pull request. For things like 'ي' and 'ى', which are similar but not technically equivalent, I prefer not to have built-in normalization (since the set of these characters, across the various scripts, is huge, and there doesn't appear to be a standard that defines how to do this generally).

I am open to having an extension mechanism through which user code can add additional normalizations, which you could use to add the ones not covered by the extending-character-removal. Does that sound reasonable?

(It might be a while until I get around to implementing all this, unfortunately.)

PR #5107, opened 2017-12-01
[css mode] Case-insensitive parsing of grammar tokens

tophf:
Grammar tokens in CSS is case-**in**sensitive as per https://www.w3.org/TR/css-syntax-3/#rule-defs
This PR makes css mode addon respect that.

A quick test in devtools timeline profiler doesn't show any perceivable difference on a relatively complex and big CSS.

Not sure how to add tests (the contributing guide suggests that)... Probably in `mode_test.js` but it's not obvious. It might make sense to add some uppercase tokens in `mode_test.css` like `@MEDIA` and a couple of random properties.


marijnh:
Thank you! Merged as b1bf7b3ad5394. Tests for the CSS mode live in `mode/css/test.js`. I've added one in 74e7447cf7.

PR #5108, opened 2017-12-03
[continuelist addon] Fix bug where continuelist for unordered list is not working

craftzdog:
You will get `NaNundefinedunordered` when inputted a return key in the unordered list.

Bug reproduce demo: https://continuelist-demo.netlify.com/demo/continuelist.html

This PR would fix it.


marijnh:
Thanks. I've pushed c7853a98, which seems to fix it in a more direct way, instead. Let me know if that doesn't solve the problem for you.

craftzdog:
That's way better. Thanks!

PR #5116, opened 2017-12-06
Fix OCaml syntax

### LINKED ISSUE #5113: Problem highlighting OCaml code
We found some issues when highlighting OCaml code, especially with:

 * Some number literals (hex, binary, exponential floats)
 * Bracket string literals
 * Some reserved words (sig, module)
 
You can see a working example with current CodeMirror version at http://jsbin.com/yutewuwafa/edit?html,js,output

cprieto:
There were some issues with OCaml syntax:
 * First element in a list is not highlighted (when it is a number)
 * Binary number literal was not highlighted
 * Hexadecimal number literal was not highlighted
 * `module` and `sig` were not highlighted

This closes #5113 

I will suggest, if the OCaml syntax continue growing we should split it in a separate file to avoid issues with F# and other ML languages (for example, standard ML).

### DIFF
--- a/mode/mllike/mllike.js
+++ b/mode/mllike/mllike.js
@@ -90,6 +90,9 @@ CodeMirror.defineMode('mllike', function(_config, parserConfig) {
         if (stream.eat('.')) {
           stream.eatWhile(/[\d]/);
         }
+        if (stream.eat(/[eE]/)) {
+          stream.eatWhile(/[\d\-+]/);
+        }
       }
       return 'number';
     }
--- a/mode/mllike/mllike.js
+++ b/mode/mllike/mllike.js
@@ -142,7 +142,9 @@ CodeMirror.defineMIME('text/x-ocaml', {
     'print_endline': 'builtin',
     'true': 'atom',
     'false': 'atom',
-    'raise': 'keyword'
+    'raise': 'keyword',
+    'module': 'keyword',
+    'sig': 'keyword'
   }
 });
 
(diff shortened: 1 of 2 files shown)


mode/mllike/mllike.js
```diff
@@ -74,9 +81,18 @@ CodeMirror.defineMode('mllike', function(_config, parserConfig) {
       return 'comment';
     }
     if (/\d/.test(ch)) {
-      stream.eatWhile(/[\d]/);
-      if (stream.eat('.')) {
+      if (ch === '0' && stream.eat(/[bB]/)) {
+        stream.eatWhile(/[01]/);
+      } if (ch === '0' && stream.eat(/[xX]/)) {
+        stream.eatWhile(/[0-9a-dA-D]/)
```
adrianheine: What about e and f?

cprieto: I have no idea what you meant.

if it is about the case with `1e-10` that is already covered in line 53 with the patch. Now, with the `f` case I have no idea what is that, OCaml doesn't have a 1f notation (that I know).

Please, if I am misunderstanding what you are pointing out I will appreciate it.

Now that you point that out, I notice I forgot Octal Integer Literals (`0o1234`), I am adding them now.

adrianheine: The line I commented on is parsing hexadecimal literals, which use digits from 0 to f, right?

cprieto: Oh thanks mate!

I really missed that one! thanks man!

Thanks to you I got as well the long integer literals and octal 😄 

Thanks again 👍

marijnh:
Great. I've merged this as 32f63fc3, minus the external theme (most of the demo pages work offline, and might as well continue to do so)

PR #5119, opened 2017-12-07
Add jsonnet (see jsonnet.org)

sparkprime:
I'm not sure this is exactly right but it seems to work.  I don't know how to run the test.

### DIFF
--- a/mode/jsonnet/index.html
+++ b/mode/jsonnet/index.html
@@ -1,0 +1,128 @@
+<!doctype html>
+
+<title>CodeMirror: Python mode</title>
+<meta charset="utf-8"/>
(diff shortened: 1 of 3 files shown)


adrianheine:
Hi. We're no longer accepting new modes into the main distribution. The recommended way to distribute new modes is to maintain them in their own repository and publish them to NPM as separate packages—see for example the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir).

sparkprime:
Thanks for the tip.  It now lives here: https://github.com/google/codemirror-mode-jsonnet

adrianheine:
Cool! Feel free to add it to https://github.com/codemirror/CodeMirror/wiki/CodeMirror-addons.

PR #5123, opened 2017-12-10
Support expanding brackets selection in selectBetweenBrackets

liorgol1:
Pressing Ctrl+Shift+m (in sublime mode) more than once will expand the selection to the next brackets.

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -183,9 +183,15 @@
         var closing = cm.scanForBracket(pos, 1);
         if (!closing) return false;
         if (closing.ch == mirror.charAt(mirror.indexOf(opening.ch) + 1)) {
-          newRanges.push({anchor: Pos(opening.pos.line, opening.pos.ch + 1),
-                          head: closing.pos});
-          break;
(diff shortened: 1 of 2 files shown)


adrianheine:
That looks good! Could you try to add a test in `test/sublime_test.js`? For reference, I found [a comment](https://stackoverflow.com/questions/13306648/expand-selection-with-brackets/16849310#16849310) and Microsoft/vscode#14089 describing the indented behavior.

adrianheine:
Thanks! Now, a last nitpick: Can you [squash your commits into one](https://www.devroom.io/2011/07/05/git-squash-your-latests-commits-into-one/) and prefix the commit message with `[sublime keymap]`?

liorgol1:
Note that the current implemented behavior is simpler than sublime's behavior (don't select the brackets). If you prefer I can try to implement something more similar to sublime but it will require a larger change to the code.
Also, the fact that sublime selects the condition of the if after the body (item 3 in the comment you referred to) seems like a bug more than intended behavior (e.g., if there is a line break after the condition  and before the "{" it will not do this).

marijnh:
Thank you! For now, this looks fine. Merged!

Issue #5129, opened 2017-12-13
[closetag addon] Add Option to disable auto indentation on auto close tag based on slash


sobisht:
Currently, on auto closing tags based on slash, it also indents the line. This causes issue in some cases.
 For example if
        smartIndent: false
        autoCloseTags: true
        
```
<html>
    <head>
    </head>
    
    <body>
        <p><
    </body>
<html>
```
On pressing slash the output is 
```
<html>
    <head>
    </head>
    
    <body>
    <p></p>
    </body>
<html>
```

Expected
```
<html>
    <head>
    </head>
    
    <body>
        <p></p>
    </body>
<html
```

The line should not be outdented .

PR #5131, opened 2017-12-13
Adding an option to disable auto indenting when auto closing tag based on slash

sobisht:
This PR is related to the issue #5129 

An option dontIndentOnAutoClose has been added which can be used to disable the code which indent the line when auto closing tags based on slash.
By default dontIndentOnAutoClose is false. This can be added as a property of "autoClosetags".

### DIFF
--- a/addon/edit/closetag.js
+++ b/addon/edit/closetag.js
@@ -97,6 +97,8 @@
   function autoCloseCurrent(cm, typingSlash) {
     var ranges = cm.listSelections(), replacements = [];
     var head = typingSlash ? "/" : "</";
+    var opt = cm.getOption("autoCloseTags");
+    var dontIndentOnSlash = (typeof opt == "object" && opt.dontIndentOnSlash);
     for (var i = 0; i < ranges.length; i++) {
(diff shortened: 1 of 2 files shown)


marijnh:
The idea is good, but this should be implemented as an option to the _addon_, rather than the main editor. See, for example, the way `indentTags` is implemented in `closetag.js`.

marijnh:
Much better, though you have some linter issues (trailing whitespace, a misaligned closing brace).

I notice `autoCloseGT` also indents lines in some circumstances. I guess the option should also disable that behavior.

sobisht:
Thanks for suggestion. I have moved the option from src to addon. 
autoCloseGT is called when tags are auto closed based on opening tag (i.e. '>' in case of html tags). I have added this option specific to auto close based on slash.

marijnh:
Right, but wouldn't it make more sense to have an option to control indentation in general, rather than one that only applies to the slash-key closing?

sobisht:
Correct. Making the option general to auto close tags.

marijnh:
Great! Merged as d9f05c90

PR #5140, opened 2017-12-22
[stylus mode] parse CSS4 hex colors - #RGBA and #RRGGBBAA

tophf:
Instead of alternation groups that match in descending lengths I've used `\b` and progressive matching.


marijnh:
Thanks!

PR #5148, opened 2017-12-28
Adding mime based on CodeMirror documentation

insign:
Based on https://codemirror.net/mode/nginx/index.html


adrianheine:
I'm pretty sure that mimetype doesn't exist, so instead the documentation should be updated.

marijnh:
Agreed. Done in 865102a0

PR #5150, opened 2017-12-29
fix text/x-php in CodeMirror.defineMIME

paladox:


marijnh:
What is this trying to fix? The fact that `text/x-php` refers to the full PHP mode, including wrapping HTML, is intentional, and changing that will break things for a lot of users. So it is probably not a good idea.

paladox:
Ah ok. I was trying to make it consistent with how the other modes do it.

PR #5152, opened 2017-12-30
Add Arduino file extension (C/C++)

4oo4:
I noticed that gogs wasn't recognizing the syntax of my Arduino sketches because of the `.ino` file extension, since Arduino code is essentially C/C++, this fixes it.

### DIFF
--- a/mode/meta.js
+++ b/mode/meta.js
@@ -12,6 +12,7 @@
   "use strict";
 
   CodeMirror.modeInfo = [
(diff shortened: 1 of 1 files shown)


marijnh:
That file should have only a single entry per MIME type, so you should add the extension to the `"C"` entry instead.

4oo4:
@marijnh Got it, added the extension under C.

marijnh:
Thanks. Merged as b045bfa7

PR #5155, opened 2018-01-02
Add Sparlql auto-complete functions

JonanOribe:
Now Codemirror have the possibility of auto-complete Sparql code too


marijnh:
Thank you, but we're not accepting additional modules into the main distribution. I recommend that you distribute this separately, for example as codemirror-sparql-hint on NPM.

PR #5156, opened 2018-01-02
[Markdown] Adding more thorough clearing for styles across list items and paragraph breaks

sjml:
Most Markdown parsers cut off the inline styles with new list items, but the Markdown mode in CodeMirror keeps them open. This means the preview in CodeMirror usually doesn't match the rendered output. I've made an example [here](http://jsbin.com/dosisudusu/edit?html,output). 

I don't 100% understand exactly how the tokenizer is working here, so this might not be the exact right spot to make the change. It worked in my tests, though.

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -166,6 +166,10 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
         }
         if (state.list !== false) {
           state.indentationDiff = lineIndentation - state.listStack[state.listStack.length - 1]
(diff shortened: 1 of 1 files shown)


sjml:
(FYI, I've realized that the patch doesn't actually fix the illustrated case; I'm investigating further. Wish pull requests had a "paused" setting.)

sjml:
I added one more fix to keep links from staying on across paragraph breaks.

marijnh:
Thanks. Merged as cacaa545

PR #5162, opened 2018-01-06
Include LC_CTYPE and LC_COLLATE keywords in x-pgsql

sql-migrate:
The CREATE DATABASE command supports LC_CTYPE and LC_COLLATE keywords as
per https://www.postgresql.org/docs/10/static/sql-createdatabase.html.
This commit adds them to the postgres sql mode.


marijnh:
Thanks. Merged as d2798d22

PR #5163, opened 2018-01-07
Add firefox-dark theme

overdodactyl:
Hello,

I'm not sure if you are accepting any new themes, but I recently put together a new template to mimic the one utilized by Firefox's inspector and browser toolbox in dark mode and thought I would pass it along.

For reference, here is what the Firefox design looks like:

firefox

And here is what my code mirror theme (in stylus) looks like:

firefox-dark

I originally submitted the code to Stylus, but the devs suggested trying here.

I understand there are already a lot of themes available, but I think this would be a nice addition for users like myself who both like the look and would prefer a more consistent look across editors.

Thanks for your time!

P.S. I'm not really sure how freely the name "firefox" can be used - so the name of the theme might have to change if it were to be added :)


marijnh:
Mozilla has a trademark on "Firefox", so yeah, I think you'd have to come up with another name. Other than that, I'd be willing to merge this.

overdodactyl:
Thanks @marijnh! Just changed the name

marijnh:
Thanks. Merged as e9e5f23b

PR #5164, opened 2018-01-07
[Markdown] linkHref status should get copied along with the rest of the state.

sjml:
The `linkHref` data wasn't getting copied through when the state was copied, so no tokens were ever carrying it.


marijnh:
Thanks. The failed CI build seems to have been a Travis fluke, so I've merged this as ce2fb7c8

PR #5165, opened 2018-01-07
Add types and keywords for ML

cprieto:
* Add keywords `as` and `when` for ML like languages
 * Add OCaml builtin types


marijnh:
Merged as ed9f4e39

PR #5166, opened 2018-01-07
Add sml

yeguanghao:


marijnh:
It doesn't look like mllike.js defines that mime type. Did you forget to add a file to the commit?

cprieto:
I had been refactoring the ML mode (for both F# and OCaml), I can take care and add SML to the ML like languages (and define x-sml as a type).

What do you reckon @marijnh ?

marijnh:
Yes, adding a definition for plain ML sounds like a good idea.

marijnh:
This has been addressed by #5173

PR #5167, opened 2018-01-08
Fix SQL hint nullpointer

neon-dev:
If you try to autocomplete at line 0, column 0 after previoulsy having edited for example a WHERE clause, the sql-hint plugin calls `doc.getRange` with an invalid start position, since prevItem is null.
Maybe my fix isn't the best solution, and you could even avoid to enter findTableByAlias enitrely, I don't know.

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -222,18 +222,20 @@
       prevItem = separator[i];
     }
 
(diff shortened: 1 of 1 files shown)


marijnh:
This file is a mess, and I'm going to stop maintaining it at the next major version, but for now, yeah, your fix looks like it at least doesn't make things worse, so I've merged it as 350f71b0

PR #5168, opened 2018-01-08
Switch order of SQL hints

neon-dev:
I propose to show column hints (if a defaultTable is set) above table hints, since you are far more often in clauses where you need those.


marijnh:
Sounds reasonable. I've merged it as 2786ff0a—let's see if anyone complains.

PR #5172, opened 2018-01-09
Refactor ML mode

cprieto:
ML mode is shared between F# and OCaml, I had been adding some keywords for OCaml but then I realized there was not a clear separation between common keywords and builtins.

I reordered the common keywords between F# and OCaml, added common builtins and noted atoms and bultins for each of them. Hopefully this will be enough for both languages for the next near future.


marijnh:
Please rebase your branch to only contain the _new_ patches (and look into the test build failure).

cprieto:
I will be opening a new PR to avoid conflicts, I think I botched something in my fork.

PR #5173, opened 2018-01-09
Refactor ML mode

cprieto:
* Add common keywords for ML languages
 * Add builtins for OCaml and F#
 * Add Standard ML as ML language


cprieto:
This relates to #5166 and #5172

marijnh:
Thanks! Merged as 774575d6

PR #5184, opened 2018-01-16
[javascript-lint addon] Fix incorrect severity

neon-dev:
When enabling strict equality checks via `lint: {options: {eqeqeq: true}}` ([see](http://jshint.com/docs/options/#eqeqeq)), found problems showed up as errors instead of warnings.
To fix it, the `fixWith()` logic had to be changed since simply adding the phrase to the warnings array would not have worked. This is because both the warnings and errors array matched this exact error and therefore the severity could never be "warning" (errors were checked after warnings).

I didn't include "Missing property name" and "Unmatched " in the new error array since they already are errors with the new logic.
"Stopping, unable to continue" also got removed since it didn't appear anywhere in the current jshint.js.

For now I've implemented everything to not break previous behavior/hinting, except the strict equality hint severity. Although I want to suggest removing the following codes from the new error array (so they can stay warnings):
- W033 (Missing semicolon) - since erroneous missing semicolons have their own code: E058
- W084 (Expected a conditional expression and instead saw an assignment) - since something like `switch (var2 = var1 + 42)` is valid js code, though not recommendable
- maybe W023/24/30/95, since there are many more " and instead saw an" hints that are already errors with their own codes, so I think they should be pretty accurate. Unfortunately I couldn't force these warnings so I couldn't check.

### DIFF
--- a/addon/lint/javascript-lint.js
+++ b/addon/lint/javascript-lint.js
@@ -51,13 +51,13 @@
   }
 
   function fixWith(error, forcedErrorCodes, replacements) {
-    var errorCode, description, fix, find, replace, found;
+    var errorCode, description, i, fix, find, replace, found;
 
     errorCode = error.code;
     description = error.description;
 
     if (error.severity !== "error") {
-      for (var i = 0; i < forcedErrorCodes.length; i++) {
+      for (i = 0; i < forcedErrorCodes.length; i++) {
         if (errorCode === forcedErrorCodes[i]) {
           error.severity = "error";
           break;
--- a/addon/lint/javascript-lint.js
+++ b/addon/lint/javascript-lint.js
@@ -68,11 +68,12 @@
     for (i = 0; i < replacements.length; i++) {
       fix = replacements[i];
       find = fix[0];
-      replace = fix[1];
       found = description.indexOf(find) !== -1;
 
       if (found) {
+        replace = fix[1];
         error.description = replace;
+        break;
       }
     }
   }
--- a/addon/lint/javascript-lint.js
+++ b/addon/lint/javascript-lint.js
@@ -14,12 +14,20 @@
 
   var bogus = [ "Dangerous comment" ];
 
-  var warnings = [ [ "Expected '{'",
+  var replacements = [ [ "Expected '{'",
                      "Statement body should be inside '{ }' braces." ] ];
 
-  var errors = [ "Missing semicolon", "Extra comma", "Missing property name",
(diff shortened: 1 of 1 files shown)


neon-dev:
I also found that the ominous `isBogus()` function seems to be obsolete. "Dangerous comment" should never occur in error.description, at least it can not come not from jshint.js. Maybe it can be removed?

marijnh:
> Maybe it can be removed?

Sure, feel free to remove it.

marijnh:
Looks good. Merged as d8d68a8a

neon-dev:
Thank you :D I really appreciate your great work.

What do you think about my suggestion removing some forced errors? I'm pretty sure about W033 and W084, but don't know about the others mentioned. It almost looks like JSHint didn't differentiate between errors and warnings at the time you implemented the addon so, maybe now they're not needed anymore.

marijnh:
I didn't implement this, so I'm not sure about the rationale for those exceptions. If JSHint has a system for distinguishing warnings and errors, I agree it doesn't make much sense for this addon to second guess it.

neon-dev:
Ok I dug a little and found that when it got implemented, JSHint was pretty different. It indeed didn't differentiate between warnings and errors, yet.
Your doc pages still include such an old version: http://ajax.aspnetcdn.com/ajax/jshint/r07/jshint.js
There are `warningAt` and `errorAt` functions, but they both lead to the same method without any distinct parameters, so it was planned but not fully implemented. Error codes were introduced in https://github.com/jshint/jshint/releases/tag/1.0.0-rc1

So my commit not only broke your doc pages (I'll fix that in a moment) but also breaks backwards compatibility to JSHint versions older than 1.0.0-rc1 (31 Dec 2012).
_

Back to topic:
I get why you may would want to mark W070 as an error, although it's the developers decision how far back he wants to support browser versions, especially IE. W112 and W117 sound pretty reasonable also, but the JSHint devs should know why they designated them as warnings. Same goes for the rest of " and instead saw" warnings. Since we don't know in which cases exactly you trigger each of these problems, all's just guessing.
So I would at least remove W033 and W084, but if you think there's no need for special overrides, I could remove the error code override entirely. I'm also not sure about the replacements array, maybe the message was unclear back then and now it could be just fine.

marijnh:
> I could remove the error code override entirely. 

I think that's a good idea, and the replacement can also go—if we're using a library to find and describe problems, we may as well leave that part of the job to that library.

I don't have a problem with breaking compatibility with ancient JSHint versions.

neon-dev:
Great!

PR #5192, opened 2018-01-17
[javascript-lint addon] Fix error markers

neon-dev:
When using tabs for indentation, the markers would not always align with the actual problem.
The old workaround for JSHint never really worked (only when a line had one tab) and I found a simple option which fixes the issue.
This time I also checked backwards compatibility, which works even fine with the old JSHint version R07.
As a second marker fix, I added an error line check, since JSHint for some reason adds configuration issues in the errors array with line: 0, character: 0. Now, if you pass an invalid option you will get a console error instead of an invisible marker.

### DIFF
--- a/addon/lint/javascript-lint.js
+++ b/addon/lint/javascript-lint.js
@@ -19,6 +19,8 @@
       }
       return [];
     }
+    if (!options.indent) // JSHint error.character actually is a column index, this fixes underlining on lines using tabs for indentation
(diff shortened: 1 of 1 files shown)


marijnh:
I like how you keep making the file smaller. Merged!

neon-dev:
Minimalism is my second name. Chopping line by line to remove clutter, improve readability as well as performance :D

Thanks again and sorry for the pull request spam lately.

PR #5193, opened 2018-01-17
[mode/meta] Add handlebars extensions to htmlmixed

silverwind:
[Handlebars](http://handlebarsjs.com/) is a template language primarily used for HTML which renders pretty well in `htmlmixed` mode, so I added its extensions to the `htmlmixed` mode.


marijnh:
Thanks!

PR #5195, opened 2018-01-18
Remove types from list of atoms

cprieto:
* By suggestion of users of OCaml Jupyter, the types are removed from
 the list of atoms so it looks more like utop

### DIFF
--- a/mode/meta.js
+++ b/mode/meta.js
@@ -128,6 +128,7 @@
     {name: "Smalltalk", mime: "text/x-stsrc", mode: "smalltalk", ext: ["st"]},
     {name: "Smarty", mime: "text/x-smarty", mode: "smarty", ext: ["tpl"]},
     {name: "Solr", mime: "text/x-solr", mode: "solr"},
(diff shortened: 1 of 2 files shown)


marijnh:
What is utop? Are types in the same namespace as variables in OCaml? If so, you could consider assigning them the token style `type` instead.

Also could you please rebase your branch before submitting a pull request, so that it's easy for me to see which patches are actually relevant? You can force-push to this branch to update the PR.

cprieto:
thanks @marijnh 

utop is the REPL for OCaml (https://i.stack.imgur.com/xnQjZ.png)

Sure, I always forget to rebase, sorry, it was my fault.

I didn't think about using the `type` annotation, I basically took the tokens from what I could figure out from the source code of other parsers in CodeMirror...

I am doing the changes right now, I will probably close this branch and reopen it when it looks better, is that ok?

marijnh:
> utop is the REPL for OCaml 

That doesn't seem to have highlighting at all so, um, I'm not sure we'll want to use that as a reference point.

> I am doing the changes right now, I will probably close this branch and reopen it when it looks better, is that ok?

Sure!

PR #5196, opened 2018-01-18
Types are not atoms but types

cprieto:
* Sorry for the previous mess
 * I sent the wrong image for utop! it has highlighting but it looks like most of the screenshots out there are a little old... Sorry!


marijnh:
Thanks! Merged as 945c9673

Issue #3638, opened 2015-11-12
DOT language support


AlgyTaylor:
Hi CodeMirror folk!

What's the chances of getting some support for DOT in CodeMirror?
https://en.wikipedia.org/wiki/DOT_%28graph_description_language%29

http://www.graphviz.org/Documentation.php

marijnh:
Not good, unless you do it yourself. See [the mode wish list](https://github.com/codemirror/CodeMirror/wiki/Mode-wish-list)

PR #5202, opened 2018-01-21
Dot mode

inaimathi:
Basic `dot` highlighting mode and example page. This should address https://github.com/codemirror/CodeMirror/issues/3638


marijnh:
Hi Leo, we're not accepting new modes into the main distribution anymore. The recommended way to distribute new modes is to give them their own repository and NPM package. You can look at the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir) for an example.

PR #5203, opened 2018-01-22
Doc constructor can now accept another Doc object

NeekSandhu:
Before, extending base `Doc` class had serious limitation, that if sub-class(say for example `DocPlus`)
required itself to be linked to another `DocPlus`, which also meant to  have their respective bases(`Doc`)
in sync, that wouldn't have worked, as `DocPlus` constructor needs to call `super()` before anything else.

Thus not being able to have a synced/linked base `Doc`(which originally, only possible by using `linkedDoc`)

This patch will allow developers to have:

```javascript
import { Doc } from 'codemirror';

class DocPlus extends Doc {
	constructor(contents) {
		super(contents, (contents instanceof Doc) ? { sharedHist: true } : 'javascript');
	}

	writeToDisk(path) {
		// ...
	}
}

const doc = new DocPlus('cool');
const synced = new DocPlus(doc);

doc.replaceRange('really ', { line: 0, ch: 0 });
console.log(synced.getValue());
// > really cool

synced.undo();
console.log(doc.getValue());
// > cool

```

So, `Doc` constructor now has following signatures:

```javascript
const newDoc = new Doc(text, [mode, firstLine, lineSep, direction])
const linkedDoc = new Doc(doc, [linkingOptions])
```

### DIFF
--- a/src/model/Doc.js
+++ b/src/model/Doc.js
@@ -25,6 +25,28 @@ let Doc = function(text, mode, firstLine, lineSep, direction) {
   if (firstLine == null) firstLine = 0
 
   BranchChunk.call(this, [new LeafChunk([new Line("", null)])])
+
+  let history = new History(null)
+
+  if (text instanceof Doc) {
+    const options = mode || {}
+    const other = text
+
+    let from = other.first, to = other.first + other.size
+    if (options.from != null && options.from > from) from = options.from
+    if (options.to != null && options.to < to) to = options.to
+    ;(other.linked || (other.linked = [])).push({doc: this, sharedHist: options.sharedHist})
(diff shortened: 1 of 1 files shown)


marijnh:
Extending `Doc` isn't really something that is supported. I recommend you use composition, rather than inheritance, if you want to associate additional data with documents.

NeekSandhu:
> Extending `Doc` isn't really something that is supported

I'm doing it in my fork. Is there any side effect I should consider?

> I recommend you use composition

You mean `defineDocExtension`?

marijnh:
> You mean defineDocExtension?

That might work. But usually, it's better to keep a data structure that has a `Doc` instance as one of its properties, and pass that around.

PR #5204, opened 2018-01-23
Add support for snippets in addon hint

DavidBabel:
Add snippet support for the hint.

demo

Instead of pushing static `values` in the array "list" parameter, you can now push simple name/value object too 👍 :
```
{
  name: "my snippet name in the widget list",
  value: "the real code which will be injected"
}
```
The code already support \n, so it's very usefull for snippets

This is done with a little hack, but actually, it works great.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -80,12 +80,18 @@
       return this.cm.state.completionActive == this;
     },
 
-    pick: function(data, i) {
+    pick: function (data, i) {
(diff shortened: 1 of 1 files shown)


marijnh:
As far as I can see, this already exists—you can give `{text: "inserted text", displayText: "visible text"}` object. Am I missing something?

DavidBabel:
Actually yes, i did not look close enought. Thanks for the work ;)

PR #5207, opened 2018-01-25
[sublime keymap] In addCursorToNextLine, skip lines that are too short to keep the cursor's position

liorgol1:
The current behavior of addCursorToNextLine is to select the next line even if it has less characters than the cursor's column. This creates the following effect (# is the cursor's position):

```
 int main() {
    #int x = 0;
    #int y = 0;
#
#    return 0;
#}
```

Instead, this change skips short lines (in this case the empty line) and keeps the cursor's column:

```
 int main() {
    #int x = 0;
    #int y = 0;

    #return 0;
}
```

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -156,11 +156,23 @@
     var ranges = cm.listSelections(), newRanges = [];
     for (var i = 0; i < ranges.length; i++) {
       var range = ranges[i];
-      var newAnchor = cm.findPosV(range.anchor, dir, "line");
-      var newHead = cm.findPosV(range.head, dir, "line");
-      var newRange = {anchor: newAnchor, head: newHead};
+      var newAnchor;
+      var newHead;
+      var amount = 1;
+      do {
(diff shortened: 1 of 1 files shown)


marijnh:
That doesn't seem to match Sublime's own behavior (going by build 3114), which adds a cursor at the end of shorter lines.

liorgol1:
That's true. On the other hand:
1. I think this fix is an improvement of the current behavior.
2. In my opinion this can be more useful than sublime's behavior as many times you do want to skip empty lines (and if you do want to select all lines, you can still do it - just press home first).
3. Sublime's behavior on this issue is very buggy (build 3126): for example, pressing Alt+Shift+Down 3 times gives:
```
int main() {
    #int x = 0;  // <--- Start with a cursor here.
#
#   #return x;  // <--- Note the two cursors on this line.
}#
```

marijnh:
> I think this fix is an improvement of the current behavior.

Maybe, but these are the Sublime Text-style bindings, so Sublime Text is our reference point. For a customization like this, I recommend you override the definition of that command in your local code.

PR #5209, opened 2018-01-26
Update regex in text/x-kotlin for parsing numbers in code

AlexandrPrendota:
Now we have a problem with drawing dots in range in Kotlin code.
For instance:
now `1..5` => **1.**(green) **.**(black) **5**(green) 
expected `1..5` => **1**(green) **..**(both black) **5**(green).

What i did? 
1) replaced existing alternative  `\d* => [^\.\D]+`
We want to see NO dot after dot and NO any symbols except numbers `/D`
2) added 1 more alternative `[\d_]+ `
Just only for writting plain digitals in code like **"1"**, **"11"**, **"12345"**

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -617,7 +617,7 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
     intendSwitch: false,
     indentStatements: false,
     multiLineStrings: true,
-    number: /^(?:0x[a-f\d_]+|0b[01_]+|(?:[\d_]+\.?\d*|\.\d+)(?:e[-+]?[\d_]+)?)(u|ll?|l|f)?/i,
+    number: /^(?:0x[a-f\d_]+|0b[01_]+|(?:[\d_]+\.?[^\.\D]+|\.\d+|[\d_]+)(?:e[-+]?[\d_]+)?)(u|ll?|l|f)?/i,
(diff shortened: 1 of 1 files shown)


marijnh:
Wouldn't a negative lookahead like `(?!\.)` after the decimal dot be a more readable way to address this?

AlexandrPrendota:
And what about digitals after the decimal dot? 
BTW.
I use https://regex101.com/ to check regex.

marijnh:
How is `[^\.\D]` different from `\d`? Also, doesn't Kotlin allow number like `1.`?

AlexandrPrendota:
If you mean `var a = 1. + 1.1` - doesn't allow.
\D - It's like a`!\d`:) We do not allow any characters except numbers [0-9]

AlexandrPrendota:
I like regex discussions ❤️😄

marijnh:
> \D - It's like a!\d

I know what `\D` does. Which is why I'm wondering why you constructed `[^\.\D]`, which says "anything except a dot or a non-digit character", which is redundant, because a dot is already a non-digit character, and equivalent to the much shorted `\d`.

AlexandrPrendota:
Exactly! `[^\.\D]+`can simplify to `[\d]+`. My mistake!

marijnh:
All right, I've merged ef234137 instead

AlexandrPrendota:
@marijnh Thanks 👍

Issue #5211, opened 2018-01-26
Matchbrackets breaks on multiple instances of CodeMirror


blittle:
There currently is a bug when rendering multiple instances of CodeMirror with the matchbrackets addon. It is due to a shared variable at https://github.com/codemirror/CodeMirror/blob/master/addon/edit/matchbrackets.js#L105

Currently this is effecting `react-codemirror2` which internally uses multiple instances of CodeMirror. 

https://github.com/scniro/react-codemirror2/issues/50

PR #5212, opened 2018-01-26
Fix matchBrackets to work with multiple cm instances

blittle:
This fixes https://github.com/codemirror/CodeMirror/issues/5211 by storing the `currentlyHighlighted` variable on the instance of CodeMirror itself. I am unsure if `state` is the best spot to store the reference.

### DIFF
--- a/addon/edit/matchbrackets.js
+++ b/addon/edit/matchbrackets.js
@@ -102,18 +102,23 @@
     }
   }
 
-  var currentlyHighlighted = null;
   function doMatchBrackets(cm) {
     cm.operation(function() {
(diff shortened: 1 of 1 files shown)


blittle:
@adrianheine thank you for the response, I'll make the change.

blittle:
@adrianheine the PR is updated

blittle:
I thought that already happens here: https://github.com/blittle/CodeMirror/blob/2f4ba2ea190b6994be98e35b00e5b9dd7fbb7521/addon/edit/matchbrackets.js#L124

blittle:
Also note, we have tested this code and are running a fork in prod until CodeMirror is fixed.

adrianheine:
Oh wow, the code in this file is in an interesting state.

PR #5221, opened 2018-01-31
Add condition on addLineWidget function

AlexandrPrendota:
Hello! 
Unfortunately we've used an old version of CodeMirror(~2016), but recently updated on version 5.34.0.
And get one problem with `addLineWidget` function.
In this function we can see the `signalLater` function.
`signalLater(cm, "lineWidgetAdded", cm, widget, typeof handle == "number" ? handle : lineNo(handle))` 
FYI: `signalLater` was added ~02.2017 but `addLineWidge` ~2016. 
In my case I always get exception in this function in `getHandlers`, because  `var cm = doc.cm;` always undefined and then in  `getHandlers` => `cm._handlers` => exception.

Probably It will be nice to add one condition on undefined before calling `signalLater` function.

Thanks.

### DIFF
--- a/src/model/line_widget.js
+++ b/src/model/line_widget.js
@@ -73,6 +73,8 @@ export function addLineWidget(doc, handle, node, options) {
     }
     return true
   })
-  signalLater(cm, "lineWidgetAdded", cm, widget, typeof handle == "number" ? handle : lineNo(handle))
+  if (cm !== undefined) {
+    signalLater(cm, "lineWidgetAdded", cm, widget, typeof handle == "number" ? handle : lineNo(handle));
+  }
   return widget
 }


marijnh:
Can you provide an example of code that produces this exception? In principle, `singalLater` is only called for documents that have an editor associated with them.

AlexandrPrendota:
Yes, sure. I can but I guess it doesn't help you :disappointed: 

```
val cmDocument = CodeMirror.Doc(file.userText, type)
val helpWrapper = document.create.div("task-help-wrapper") 
helpWidget = cmDocument.addLineWidget(0, helpWrapper, json("above" to true, "noHScroll" to true))
```
And I get exception in the third string.

Actually, I think, it's possible to add the condition on undefine or throw a pretty exception like "only with editor", but I prefer the first solution: may be somebody else has the same problem.

Thanks

marijnh:
Ah, I see, yeah, that's clearly a bug and I guess no one really tested line widgets on editor-less documents yet. I've pushed a slightly different patch 291fd84f, since there's no need to constrain to exactly `undefined` (the link may also be `null`).

AlexandrPrendota:
@marijnh When you have a plan to publish a patch for 5.34.0?

marijnh:
In three weeks. CodeMirror releases are strictly time based unless a major regression or security issue is found

AlexandrPrendota:
@marijnh so sad  :(

PR #5225, opened 2018-02-01
[sublime keymap] Fix addCursorToSelection.

liorgol1:
In addCursorToSelection, keep the cursor's position after short lines. For example (# is the cursor's position):

```
 int main() {
    #int x = 0;   // <--- Start with a cursor here.
    #int y = 0;
#
    #return 0;
}
```

instead of the current behavior:
```
 int main() {
    #int x = 0;   // <--- Start with a cursor here.
    #int y = 0;
#
#    return 0;
}
```

This is similar to [this pull request](https://github.com/codemirror/CodeMirror/pull/5207), but the behavior is more similar to Sublime's behavior.

### DIFF
--- a/keymap/sublime.js
+++ b/keymap/sublime.js
@@ -156,8 +156,14 @@
     var ranges = cm.listSelections(), newRanges = [];
     for (var i = 0; i < ranges.length; i++) {
       var range = ranges[i];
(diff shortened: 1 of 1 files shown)


marijnh:
That's reasonable! Merged.

PR #5226, opened 2018-02-01
[stex mode] LaTeX-style inline math delimiters \( \)

nickkolok:
Add support for inline math in LaTeX-style:
```tex
This is an equation \( E=mc^2 \), named after Einstein.
```


marijnh:
Thank you. Merged!

PR #5233, opened 2018-02-02
[stex mode] Minor technical fastenation

nickkolok:
I used CodeMirror to parse files of size ~ 1MB and this improvement was noticeable.

### DIFF
--- a/mode/stex/stex.js
+++ b/mode/stex/stex.js
@@ -131,10 +131,10 @@
       }
 
       var ch = source.next();
-      if (ch == "%") {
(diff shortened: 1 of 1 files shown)


marijnh:
As long as no actual type conversion happens (which is the case here—all operands are strings), there's not actually a speed advantage to `===`. JSPerf is down again, so I don't have easy access to benchmarks, but I doubt you rigorously measured the difference.

Since this project prefers `==` when appropriate, I'm not going to merge this.

PR #5236, opened 2018-02-03
[haskell mode] Add "Applicative" and "*>" and "<*"

vaibhavsagar:
As of GHC 7.10 these are in the Prelude.


vaibhavsagar:
Thanks!

PR #5241, opened 2018-02-06
CodeMirror.hasMode

neon-dev:
I created a `hasMode(spec)` function, to check if a given spec resolves to an existing mode. I needed this for dynamic mode creations via overlay.js, so I don't end up creating the same mode over and over again.
Thought it could be useful to others.

### DIFF
--- a/addon/runmode/runmode-standalone.js
+++ b/addon/runmode/runmode-standalone.js
@@ -90,9 +90,15 @@ CodeMirror.resolveMode = function(spec) {
   if (typeof spec == "string") return {name: spec};
   else return spec || {name: "null"};
 };
+function getModeFactory(spec) {
+  var modeName = CodeMirror.resolveMode(spec).name;
+  return modes[modeName];
+}
+CodeMirror.hasMode = function (spec) {
+  return typeof getModeFactory(spec) === "function";
+};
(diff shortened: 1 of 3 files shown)


neon-dev:
I didn't update the docs in my commits, that'd be up to you :x

addon/runmode/runmode-standalone.js
```diff
@@ -90,9 +90,15 @@ CodeMirror.resolveMode = function(spec) {
...
+  var modeName = CodeMirror.resolveMode(spec).name;
+  return modes[modeName];
+}
+CodeMirror.hasMode = function (spec) {
+  return typeof getModeFactory(spec) === "function";
+};
 CodeMirror.getMode = function (options, spec) {
-  spec = CodeMirror.resolveMode(spec);
-  var mfactory = modes[spec.name];
+  var mfactory = getModeFactory(spec);
```
adrianheine: `spec` is not the result from `resolveMode` in the remainder of this function.

neon-dev:
Oops, you're right. Sorry, will fix that later.

marijnh:
This can already be done with the existing interface (`CodeMirror.modes.hasOwnProperty(CodeMirror.resolveMode(spec).name)`), and I don't think it warrants an additional exported method.

PR #5242, opened 2018-02-06
[Dockerfile mode] Some improvements

ficristo:
add more instructions
highlight instructions as 'keyword'
highlight 'from ... as'
add tests

### DIFF
--- a/mode/dockerfile/dockerfile.js
+++ b/mode/dockerfile/dockerfile.js
@@ -12,9 +12,9 @@
   "use strict";
 
   // Collect all Dockerfile directives
-  var instructions = ["from", "maintainer", "run", "cmd", "expose", "env",
+  var instructions = ["arg", "from", "maintainer", "run", "cmd", "label", "expose", "env",
                       "add", "copy", "entrypoint", "volume", "user",
-                      "workdir", "onbuild"],
+                      "workdir", "onbuild", "stopsignal", "healthcheck", "shell"],
       instructionRegex = "(" + instructions.join('|') + ")",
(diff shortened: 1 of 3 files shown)


mode/dockerfile/dockerfile.js
```diff
@@ -12,9 +12,10 @@
   "use strict";
 
   // Collect all Dockerfile directives
-  var instructions = ["from", "maintainer", "run", "cmd", "expose", "env",
+  var instructions = ["arg", "from", "maintainer", "run", "cmd", "label", "expose", "env",
                       "add", "copy", "entrypoint", "volume", "user",
-                      "workdir", "onbuild"],
+                      "workdir", "onbuild", "stopsignal", "healthcheck", "shell"],
+      instructionFromAs = /^\s*\b(?:(FROM))\b(.*?)\b(?:(AS))\b/i,
```
adrianheine: I think this would not correctly handle the following:

> FROM debian:stretch # I tend to use stable as that is more stable

ficristo: I fixed it and added a testcase.

mode/dockerfile/dockerfile.js
```diff
@@ -26,22 +26,62 @@
         regex: /#.*$/,
         token: "comment"
       },
+      {
+        regex: /^(\s*)\b(?:(from))\b/i,
```
adrianheine: You can just drop the non-capturing group.

ficristo: Done.

ficristo:
Thanks!

PR #5254, opened 2018-02-10
Highlight NULL instead of null in C/C++

vincentwoo:
CodeMirror C and C++ currently highlight the string "null", which isn't defined in C/C++. Those languages do macro "NULL", though.


marijnh:
Thanks. Merged!

PR #5255, opened 2018-02-10
Add missing NSIS commands

idleberg:
This PR adds `PEDllCharacteristics` and `PESubsysVer`, two commands introduced in NSIS 3.0 Beta 1


adrianheine:
Thanks! `PESubsysVer` apparently was added in 3.0b2, though.

PR #5259, opened 2018-02-12
Add support for dynamic import operator to JavaScript mode

tmcw:
Adds support for dynamic import syntax to the JavaScript mode. Currently this isn't handled, so, for instance, property access from dynamic imports - like

```js
import('foo').then(() => {});
```

Highlights `then` as a `variable`, not a `property`.

This PR allows for dynamic imports at the top level by detecting `(`, and handles imports in expressions by just continuing to the expression mode when they're detected.


adrianheine:
dynamic import is stage 3, but I think that's okay for CodeMirror. Could you add a test?

tmcw:
👍 Test cases added.

marijnh:
Thanks!

PR #5263, opened 2018-02-13
[Dockerfile mode] Highlight strings and ports

ficristo:
- highlight strings
- highlight ports as numbers in expose instruction
- highlight comments as such only when they appear at the start of a line
- consider also ` as a line continuation

### DIFF
--- a/mode/dockerfile/dockerfile.js
+++ b/mode/dockerfile/dockerfile.js
@@ -11,35 +11,64 @@
 })(function(CodeMirror) {
   "use strict";
 
+  var from = "from";
+  var fromRegex = new RegExp("^(\\s*)\\b(" + from + ")\\b", "i");
+
+  var shells = ["run", "cmd", "entrypoint", "shell"];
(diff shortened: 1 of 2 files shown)


mode/dockerfile/dockerfile.js
```diff
@@ -11,35 +11,64 @@
 })(function(CodeMirror) {
   "use strict";
 
+  var from = "from";
+  var fromRegex = new RegExp("^(\\s*)\\b(" + from + ")\\b", "i");
```
marijnh: Does changing this from a simple inline regexp to this complicated definition add anything?

ficristo: It seemed more consitent with the other instructions but I can change it back. Let me know.

marijnh: It's harmless, I guess, but it seemed a bit of a pointless change.

marijnh:
Thanks! Merged.

ficristo:
Thank you!

Issue #5265, opened 2018-02-14
Match highlighter breaks if the match contains regex special characters


tsznuk:
If the match highlighter addon is used with the following parameters:
   - `showToken` set to an expression that can accept a character with a special meaning in regular expressions (e.g. `[?\w]`)
  - `annotateScrollbar` set to `true`

the following error will be printed on the console when a match is being highlighted:

`Uncaught SyntaxError: Invalid regular expression: /\b?o\b/: Nothing to repeat`

The problem seems to be caused by this line:

https://github.com/codemirror/CodeMirror/blob/ab8bf8e5efd210568893824ee7c8f8b08000906c/addon/search/match-highlighter.js#L93

Which does not escape the match before turning it into a regex (and also uses the default word boundaries instead of relying on `showToken`).

PR #5266, opened 2018-02-14
Support regex special characters in match highlighting addon (#5265)

tsznuk:
showMatchesOnScrollbar now accepts an additional option named 'filter'.
If set it must be a function that will receive a CodeMirror handle
and a match position and return a boolean result. Only matches for
which the filter returned 'true' will be highlighted.

This new mechanism has been used to check for boundaries when
annotating the scrollbar in the match highlighter. This should resolve 
issue #5265.

### DIFF
--- a/addon/search/match-highlighter.js
+++ b/addon/search/match-highlighter.js
@@ -87,12 +87,20 @@
   }
 
   function addOverlay(cm, query, hasBoundary, style) {
+    function filter(cm, match) {
+      return boundariesAroundString(cm.getLine(match.from.line),
+                                    match.from.ch,
+                                    match.to.ch,
+                                    hasBoundary);
+    }
     var state = cm.state.matchHighlighter;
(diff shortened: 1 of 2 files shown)


marijnh:
In the issue you point at the unescaped regexp string (which indeed is wrong), but this fix doesn't seem to have anything to do with that. Did you link the wrong issue number? If not, why don't we solve this by escaping the string given to the regexp?

tsznuk:
The fix replaces the whole regex construction by matching the raw pattern without any boundaries and then filtering matches by checking for boundaries.

Escaping the string would solve the issue too, but it would be hard to match boundaries correctly (I guess that is technically a separate issue) - We'd need to negate a regex somehow.

marijnh:
Patch aab9d716 adds the escaping. I think it should be possible to also do the boundary checking in the regexp (something like `(^|[^...])` in front and `(?![^...])` after), and we don't need an extra feature in matchesonscrollbar for that, but I haven't added that part yet.

PR #5273, opened 2018-02-19
Add ECMAScript 2015 keywords to JavaScript hint addon

tmcw:
This explicitly adds keywords that are already syntax-highlighted
and correctly parsed by CodeMirror already in the JavaScript mode.


marijnh:
Thanks. Merged as 749409d4

PR #5281, opened 2018-02-26
Fix operator regex

demoore:
Since django operators aren't words we should use the negated version of `\b`: `\B`.


marijnh:
That would only match if there are additional non-word characters on the sides of the token, so you couldn't put an operator directly next to a word. Maybe the proper fix is to drop these boundary checks entirely?

Issue #5287, opened 2018-02-28
Velocity variables with dash are not well highlighted


cmglez10:
In velocity language, when a variable contains a dash (eg. ${MASK-LAN}), only the left part is highlighted as a variable.

adrianheine:
That should be the regular expression in [line 85](https://github.com/codemirror/CodeMirror/blob/master/mode/velocity/velocity.js#L85). Would you like to try to fix this?

cmglez10:
Sorry, Regexp and me are not best friends... :(

adrianheine:
Ok, it's actually pretty easy. Currently, variable names after dollar signs are matched with `/[\w\d\$_\.{}]/`, which means: any letter, number, the dollar sign, an underscore, a dot or square brackets are acceptable. You want to add a dash, so the correct regular expression would be `/[\w\d\$_\.{}-]/`. Do you want to try submitting a pull request for that change?

cmglez10:
Done! Sorry about the delay.
Thanks!

marijnh:
Solved via #5292

PR #5292, opened 2018-03-06
[BUGFIX] Velocity variables with dash are not well highlighted #5287

cmglez10:
It is introduced dash in regexp for variables
#5287


marijnh:
The test failures seem to be unrelated to your patch—I'm not sure what's up with them. I've merged this, will look into the test issue separately. Thanks!

PR #5296, opened 2018-03-06
Implement binding types for var/let/const variables.

### LINKED ISSUE #5138: JavaScript mode scope highlighting doesn't adjust for let/const block scopes
Here's an example:

2017-12-20 at 1 15 pm

```js
function findSequence(goal) {
  {
    let aBlockScopedVar = 10;
  }
  return aBlockScopedVar;
}
```

Kind of a nitpicky issue, but `aBlockScopedVar` is highlighted as `variable-2` in this case, when it should be highlighted as `variable`, because the block containing the `let` binding contains the binding.

As far as I can tell, fixing this would require context objects to have a 'type' - there'd be 'function contexts' and 'block contexts': `var` bindings would go into `register` and get registered to the nearest function context if there is one, and `let` or `const` bindings would attach themselves to the nearest context of either type.

tmcw:
Work in progress! This doesn't pass the full test suite yet, and I'm not sure why yet. This aims to fix #5138 - it makes the register() method in the JavaScript highlighting mode aware of the type of binding it is in, and also adds block context that contains block-scoped definitions. Review + hints for how to get to the finish line much appreciated!

### DIFF
--- a/mode/javascript/javascript.js
+++ b/mode/javascript/javascript.js
@@ -302,8 +302,13 @@ CodeMirror.defineMode("javascript", function(config, parserConfig) {
   function popcontext() {
     if (cx.state.context.block) {
       var newVars = {name: "this", next: {name: "arguments"}};
+      // drop block-scoped variables
       for (var v = cx.state.localVars; v; v = v.next) {
         if (!v.block) newVars = { name: v.name, scope: v.scope, next: newVars };
+      }
+      // inherit from previous scope
+      for (v = cx.state.context.vars; v; v = v.next) {
+        newVars = { name: v.name, scope: v.scope, next: newVars };
       }
       cx.state.localVars = newVars;
     } else {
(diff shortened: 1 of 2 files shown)


marijnh:
Nice. I'd keep a single chain of scopes, rather than separate ones for function and block scopes, though. You can add a property that describes the type of the scope.

tmcw:
Okay, refactored to use a single chain of prototypes. Still pretty baffled by what's going wrong with the indentation failures, though - I can see that it's now setting `align=true` when re-entering the mode, and that's aligning certain blocks incorrectly, but how that connects to the very minimal code change is a mystery.

tmcw:
It works! May also want to implement `for` scopes.

marijnh:
Merged as 40027e21423 and adjusted in 2e99e1414ac to avoid allocating too many objects

PR #5301, opened 2018-03-08
Fix bug with indentation in multiplex.js

thomasjm:
Some modes may return `CodeMirror.Pass` from their `indent` methods, which in the current code can get fed into the start state of another mode and mess things up. Fixed to check for this case.


marijnh:
Thanks. Modes and addons are still written in ES5 and not compiled. Could you change your `let`s to `var`s?

thomasjm:
Done

marijnh:
Thanks! Merged.

PR #5309, opened 2018-03-13
Fix for some mouse events being sent to the wrong document

Me1000:
If the CodeMirror script is being run in a different document than the CodeMirror instance itself, some mouse events will be sent to the wrong document. In particular mouse move is sent to `document` rather than the editor's document, which causes selection to break.

This pull request fixes that issue allowing instances of CodeMirror to be run in different documents than the script itself. 

To get code mirror to run in different documents you need to do something like:
```javascript
var editor = CodeMirror.fromTextArea(window.frames["iframe"].document.getElementById("code"), {
    lineNumbers: true,
    lineWrapping: true
  });
```

We're specifically using this technique for some upcoming changes to our [RunKit Embeds](https://runkit.com/docs/embed). We load our main app in a background frame, which then can render to any number of additional iframes without paying the price of additional loads.

### DIFF
--- a/src/edit/mouse_events.js
+++ b/src/edit/mouse_events.js
@@ -149,8 +149,8 @@ function leftButtonStartDrag(cm, event, pos, behavior) {
   let dragEnd = operation(cm, e => {
     if (webkit) display.scroller.draggable = false
     cm.state.draggingText = false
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks!

Issue #5169, opened 2018-01-08
getTokenAt() includes braces


neon-dev:
I've encountered an inconsistent behavior while implementing custom hints in sql mode using `CodeMirror.hint.fromList`. These custom hints are available inside double curly braces (will be parsed/substituted before the actual SQL statement execution).
The problem is, `fromList` uses `getTokenAt` internally which returns strange tokens in this mode.
This issue is restricted to modes where curly braces and regular words have no special token type.

Example:
The cursor position is anywhere inside, or at the end of `test`, which is enclosed in double curly braces (`{{test}}`).
Now I call `CodeMirror.hint.fromList(cm, {words: ['testWord']})`, but testWord will not be found as a possible result.
This is because `cm.getTokenAt` returns a token containing `{test` instead of `test`. Currently I need to modify the input words, as well as the returned list to make it work and display properly, which is an ugly workaround.

I'd expect `cm.getTokenAt(cursor)` to return either `test` or `{{test}}`. The first one feels more intuitive and surely is more useful in my case, but I get that curly braces could also be interpreted as plain text in some languages. In fact, in sql mode curly braces have no special meaning at all.

marijnh:
The mode determines how the code is split into tokens. `getTokenAt` just queries those. There is no guarantee that token boundaries happen to fall at the point where it is convenient for your completion code. Unless you can change the mode, of course, in which case you can make sure it produces different token types for the braces and their content.

neon-dev:
I know there's no guarantee for that, that's not what I was after. Just thought that this could be wrong, since these braces have the same token type and aren't special syntax chars.

Regarding my problem, I think I'll keep the workaround I have. Changing the mode on the fly seems to be the worse solution if it introduces more overhead than iterating and replacing a few array values.
If `hint.fromList` would consider word boundaries instead of tokens, that'd be prettier. It takes "words" as an argument, so word boundaries seem reasonable, but I don't know how you intended that function to work.

marijnh:
> It takes "words" as an argument

It is indeed called "words", but it can hold any strings, which can occasionally be useful. I think fixing this in the mode is the way to go here.

neon-dev:
I guess enabling [overlay.js](https://codemirror.net/demo/mustache.html) in my sql editor like in that example should be an adequate solution. Plus we get nice highlighting.
If I understand it correctly, it will override tokenizing and thus the the token type, if returned.
[...]

PR #5312, opened 2018-03-16
[show-hint addon] fixed fromList didn't show all completions if the cursor was inside the token

neon-dev:
The hint logic of the `fromList` command did not respect the cursor position inside the token.
Now it will list all matching completions for the substring up to the cursor, instead of only the ones matching the complete token. The behavior is now the same as for the regular autocompletion.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -400,7 +400,8 @@
     var cur = cm.getCursor(), token = cm.getTokenAt(cur);
     var to = CodeMirror.Pos(cur.line, token.end);
     if (token.string && /\w/.test(token.string[token.string.length - 1])) {
-      var term = token.string, from = CodeMirror.Pos(cur.line, token.start);
+      var length = token.string.length - (token.end - cur.ch);
+      var term = token.string.substr(0, length), from = CodeMirror.Pos(cur.line, token.start);
     } else {
       var term = "", from = to;
     }


addon/hint/show-hint.js
```diff
@@ -400,7 +400,8 @@
     var cur = cm.getCursor(), token = cm.getTokenAt(cur);
     var to = CodeMirror.Pos(cur.line, token.end);
     if (token.string && /\w/.test(token.string[token.string.length - 1])) {
```
marijnh: Maybe this test should also check for the character before the cursor now, rather than the one at the end of the token?

neon-dev: Do you have an example? I'm not quite getting what you mean :D

marijnh: I meant the test right before your change, where it only uses the token text when it ends in a word character.

neon-dev: I saw the line, but didn't understand why you'd want to check for the character in front of the cursor.
So effectively, that would skip completions for tokens containing non-word characters, if the cursor is right after one?

Actually, I'd even remove the current regex test and let the word check loop handle everything. As you said once [here](https://github.com/codemirror/CodeMirror/issues/5169#issuecomment-356540344), being able to put anything into the word list may be useful. For example if you work with AngularJS, var and function names starting with `$` could be completed right at the first character.

marijnh: That's a good point. But then we'd want to handle the case where no matches are found _and_ the cursor is directly after a non-word character specially, by also trying to complete with the empty string, so that you can still complete (usual, word completions) with the cursor after a non-word token.

neon-dev: Ah I see, you mean something like completions after a dot for example. Since the dot would be considered as a valid token, the completion list would be empty.

In that case maybe this could solve the problem `if (token.string && (token.type == 'variable' || token.type == 'property'))`. A third OR condition would be needed if the current mode doesn't support these token types. Also this would make the addon complete in less situations, which may be good or bad, depending on your situation.
Now that I think about it, we maybe should make this configurable (array of valid token types), then the power lies in the users hand and the addon can be fine tuned for each use case which would make it way more practical.
But that's just an idea and out of scope for this pull request. If you still want the word check you suggested initially, I can implement that for now, even though I don't think it would add anything because we are inside one token which ends with a word character.

marijnh:
I've merged this as e732d0e9fd11 and followed up with d49ef290157380, which was the simplest way to make things consistent again.

neon-dev:
Thanks!

PR #5313, opened 2018-03-17
[merge addon] Use CSS :after to style scroll-lock icon

gorhill:
Primary motivation was to remove the use of `innerHTML` because the
Firefox extension validator raises a warning about this.

The obvious solution is to use `textContent` instead of `innerHTML`
but I thought it could be useful to have the ability to pick a
custom visual for the scroll-lock icon.

As per MDN documentation, all the supported browsers as per
CodeMirror documentation support `:after`.

### DIFF
--- a/addon/merge/merge.css
+++ b/addon/merge/merge.css
@@ -48,6 +48,12 @@
   color: #555;
   line-height: 1;
 }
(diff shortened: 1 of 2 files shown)


marijnh:
That seems sensible. Merged!

PR #5328, opened 2018-03-26
Fix argument of scanForBracket method call issue#5334

hidaiy:
I created this pull requect for issue #5334


marijnh:
Thank you! Merged.

PR #5338, opened 2018-03-31
[python mode] Highlight expressions inside python f-strings

BoPeng:
This patch enables syntax highlighting of [python f-string](https://www.python.org/dev/peps/pep-0498/). As shown by [added examples](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/index.html#L130)

image

CodeMirror now goes inside of f-strings and highlights expressions inside braces as regular python expressions.

Format strings are handled by a [formatStringFactory](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L190) which is extended (copied) from [tokenStringFactory ](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L260) with an added [state for f-string](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L198). Inside the f-string expression, a regular [tokenBaseInner](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L214) is used to parse the string until `}` or end of string is encountered.

The patch seems to work but has **not** been thoroughly tested because I would like to know if this is useful/acceptable to the community and if my implementation is technically sound.

### DIFF
--- a/mode/python/index.html
+++ b/mode/python/index.html
@@ -126,6 +126,15 @@ class ExampleClass(ParentClass):
     def __init__(self, mixin = 'Hello'):
         self.mixin = mixin
 
+# Python 3.6 f-strings (https://www.python.org/dev/peps/pep-0498/)
+f'My name is {name}, my age next year is {age+1}, my anniversary is {anniversary:%A, %B %d, %Y}.'
+f'He said his name is {name!r}.'
+f"""He said his name is {name!r}."""
+f'{"quoted string"}'
+f'{{ {4*10} }}'
+f'This is an error }'
+f'This is ok }}'
+fr'x={4*10}\n'
 </textarea></div>
 
 
--- a/mode/python/index.html
+++ b/mode/python/index.html
@@ -132,6 +132,7 @@ f'He said his name is {name!r}.'
 f"""He said his name is {name!r}."""
 f'{"quoted string"}'
 f'{{ {4*10} }}'
+f'{ f"{nested f-string}" is treated as "normal string" }'
 f'This is an error }'
 f'This is ok }}'
 fr'x={4*10}\n'
--- a/mode/python/index.html
+++ b/mode/python/index.html
@@ -132,7 +132,7 @@ f'He said his name is {name!r}.'
 f"""He said his name is {name!r}."""
 f'{"quoted string"}'
 f'{{ {4*10} }}'
-f'{ f"{nested f-string}" is treated as "normal string" }'
+f'{ "Nested f-string " + f"{2**100}" + " is treated as normal string" }'
 f'This is an error }'
 f'This is ok }}'
 fr'x={4*10}\n'
--- a/mode/python/python.js
+++ b/mode/python/python.js
@@ -142,7 +142,7 @@
 
       // Handle Strings
       if (stream.match(stringPrefixes)) {
-        var isFmtString = stream.current().toLowerCase().includes('f');
+        var isFmtString = stream.current().toLowerCase().indexOf('f') !== -1;
         if (!isFmtString || state.fstr_state !== null) {
           // if this is a nested format string (e.g. f' {   f"{10*10}" + "a" }' )
           // we do not format the nested expression and treat the nested format
--- a/mode/python/python.js
+++ b/mode/python/python.js
@@ -143,7 +143,7 @@
       // Handle Strings
       if (stream.match(stringPrefixes)) {
         var isFmtString = stream.current().toLowerCase().indexOf('f') !== -1;
-        state.tokenize = isFmtString ? formatStringFactory(stream.current()) : tokenStringFactory(stream.current());
+        state.tokenize = tokenStringFactory(stream.current(), isFmtString);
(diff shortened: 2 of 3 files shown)


BoPeng:
[This test](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/test.js#L33) fails, not sure how to fix.

```
python_fValidStringPrefix
f'this is a {formatted} string'
expected:
f'this·is·a·{formatted}·string'
string
observed: [display states]
f'this·is·a·	{	formatted	}	·string'
string	null	variable	null	string

1 failure
```

marijnh:
Thanks! Would it be possible to, instead of having a new `fstr_state` property on the state, to encode this information in the closure stored in `state.tokenize`? Avoiding that extra separate field would be nice.

BoPeng:
`fstr_state` is primarily used by [line 211](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L211)

```
tokenBaseInner(stream, state.fstr_state)
```
to parse expressions inside f-string. It is also used outside of `formatStringFactory` to [detect nested fstr](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L146) (therefore has to be global in `state`) to treat nested f-string `f"nsted f-str with {100}"` in
```
f' { f"nested f-str with { 100 }"  + " hello"}'
```

as a regular string. Moving `fstr_state` to the closure will cause an error in this case, unless we set a flag in `tokernize` (e.g. set `tokenString.isFormatString=True` [here](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L253)) and use a stack to push/pop `fstr_state` to handle nested f-string properly. Do you think it is worth the trouble to do this?

BoPeng:
OK, I have merged `formatStringFactory` to `tokenStringFactory` with a [`isFmtString` flag](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L178) and used an [enclosure state](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L184) to maintain states of expressions.

The trick to get around of nested f-string problem is to explicitly handle nested f-strings as regular strings in [this line](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L204) so that [tokenBaseInner](https://github.com/BoPeng/CodeMirror/blob/f-string/mode/python/python.js#L207) would not see them. It is possible to dig into nested f-strings but it is very troublesome and there is no real need to do it.

marijnh:
Modifying a closed-over variable like that isn't safe—CodeMirror will copy and resume states, so this will lead to copies (which hold the same tokenize closure) seeing changes made by previous runs. A better approach would be to update `state.tokenize` to a new closure that closes over the updated `fstr_state` value.

BoPeng:
I am sorry but I am afraid that I do not know CodeMirror/JS well enough to do that. I was surprised that there is no `copyState` function in the Python mode as others do but I still put `fstr_state` in global python state as recommended by the documentation.  Putting `fstr_state` in a closure as you suggested was already a bit of stretch for me and I do not really know what you meant by a new closure. I have invited @marijnh to my fork, perhaps you can make the change by yourself?

BoPeng:
Allow me to write down in detail what is going on here after more reading.

If I understand correctly, you are referring to [this line of code](https://github.com/BoPeng/CodeMirror/commit/0ff649f993253ff26d70bb6e10b61a84ba34ad47#diff-04f7f1f1bbbab888742c7e849187a79cR207)

```
            return tokenBaseInner(stream, fstr_state);
```
which, in case of f-string, resort to a basic `tokenBaseInner` tokernizer to parse the expression inside f-strings. It would work very well without `copyState` because `tokenBaseInner` would parse the entire expression and return the correct tokens with state saved in `fstr_state`. The key point here is that the global python state and tokenize are **not changed** (still ` tokenStringFactory`) so CM would return to this line for the next token until the expression is terminated by `tokenStringFactory()`, not by `tokenBaseInner`.

I do not know when `copyState` will be called but during the parsing of nested expressions, if the state is copied, the `tokenStringFactory()` would be copied, with `fstr_state` copied by reference, which is considered unsafe because restoring `fstr_state` is unpredictable. Putting `fstr_state` in the global state as the original patch did would prevent us from this problem because `fstr_state` would be copied and restored correctly, but let us focus on the problem of closure state.

What you suggested was to 
>  update state.tokenize to a new closure that closes over the updated fstr_state value

I first thought you meant to do
```
state.tokenize = tokenBaseInner
```
but that would totally mess up the parsing logic. I suppose what you really meant was to create a new `fstr_state` for each new copy of `state.tokenize` (currently `tokenStringFactory`) so I have submitted a patch just for that. Note that the `clone` function is basically `CodeMirror.copyState(??, state)` but I do not know what to put as `??` here.

marijnh:
Nope, that's not what I meant either. I've merged your patches up to 2c2fefd, and followed up with d0b1500b which, I think, also adds proper support for nesting format strings.

PR #5354, opened 2018-04-10
Fix space key clearing composition with Korean IME in contenteditable mode

muan:
There is a problem with Korean IME in the contenteditable mode, where `space` key will clear the current composition. As demonstrated here: (Key strokes: 'g', 'k', 's', 'r', 'm', 'f', SPACE in Qwerty layout)

This is caused by the `keypress` event. `keypress` should not happen during IME composition ([as spec](https://www.w3.org/TR/uievents/#keys-IME) + [ref](https://lists.w3.org/Archives/Public/www-dom/2011JulSep/0132.html)), but it does on Windows and Ubuntu for Korean IMEs, not for Chinese and Japanese IME, or anything on Mac. 

This PR fixes it by returning the keypress event if composing. Fixed behavior:

### DIFF
--- a/src/input/ContentEditableInput.js
+++ b/src/input/ContentEditableInput.js
@@ -355,7 +355,7 @@ export default class ContentEditableInput {
   }
 
   onKeyPress(e) {
-    if (e.charCode == 0) return
+    if (e.charCode == 0 || this.composing) return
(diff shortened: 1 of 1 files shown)


marijnh:
Thank you! Merged.

PR #5358, opened 2018-04-13
Adds support for options.finishInit to CodeMirror.fromTextArea

tonytlwu:
This allows `options.finishInit` to work in a more consistent manner, even when running `CodeMirror.fromTextArea()`

Although this is an undocumented feature, it's quite useful and it would be good to know that this works consistently.

### DIFF
--- a/src/edit/fromTextArea.js
+++ b/src/edit/fromTextArea.js
@@ -38,6 +38,11 @@ export function fromTextArea(textarea, options) {
     }
   }
 
+  let _finishInit
+  if (typeof options.finishInit === 'function') {
+    _finishInit = options.finishInit
+  }
+
   options.finishInit = cm => {
     cm.save = save
     cm.getTextArea = () => textarea
@@ -51,6 +56,10 @@ export function fromTextArea(textarea, options) {
         if (typeof textarea.form.submit == "function")
           textarea.form.submit = realSubmit
       }
+    }
+
+    if (typeof options.finishInit === 'function') {
+      _finishInit(cm)
     }
   }
 


marijnh:
`finishInit` is an internal hack, not a documented feature. What are you doing with it?

tonytlwu:
I'm using it to ensure I have a reliable callback for when the CodeMirror UI has finished initializing. I have several instances on a page and I'm using Promises to ensure they're all initialized before running further operations.

marijnh:
Why not just define a wrapper function that initializes an editor and then calls your callback?

tonytlwu:
Like this?

```js
function initializeCodeMirror(callback) {
  var instance = CodeMirror.fromTextArea(textarea, options);
  if (typeof callback === 'function') {
    callback(instance);
  }
}

initializeCodeMirror(function(){
  console.log('CodeMirror UI initialized!');
});
```

marijnh:
Well, if you're explicitly providing the callback to the function, you might as well just call it directly when initializing the editor. I.e. where you used to have `finishInit: ...` in your code, you could just do whatever was in there after creating the editor, no?

tonytlwu:
Are you suggesting **A** or **B** from below? **A**, I assume?

#### A

```js
function initializeCodeMirror(callback) {
  var instance = CodeMirror.fromTextArea(textarea, options);
  console.log('CodeMirror UI initialized!');
}

initializeCodeMirror();
```

#### B

```js
function initializeCodeMirror(callback) {
  var instance = CodeMirror.fromTextArea(textarea, {
    finishInit: function () {
      console.log('CodeMirror UI initialized!');
    }
  });
}

initializeCodeMirror();
```

marijnh:
I really don't care how you do it, I'm just not seeing the value in using this option.

tonytlwu:
I wanted to understand what you had in mind and compare my understanding of it.

**Case A** calls the function too early because the UI has not been fully initialized yet. The instance object is created, but the UI initialization hasn't completed.

`options.finishInit` helps in calling back at the correct time, after the UI initialization is complete. However,  in **Case B** above, the callback doesn't work because `CodeMirror.fromTextArea` [overrides the configured callback](https://github.com/codemirror/CodeMirror/blob/master/src/edit/fromTextArea.js#L41-L55).

marijnh:
> Case A calls the function too early because the UI has not been fully initialized yet

That's not true—CodeMirror initializes synchronously.

Issue #5368, opened 2018-04-20
Multiplexing mode doesn't work if `parseDelimiters` is true and `open` and `close` parameters are equal


AlbertHilb:
This happens because when the parser finds an open delimiter, it switches to inner mode without consuming it.
Then, when in inner mode, the (unconsumed) open delimiter is parsed as a close delimiter.
So, the same substring starts and ends the inner mode.

See that [example](http://jsbin.com/zihavuhafi/edit?html,js,output).

Cheers

PR #5369, opened 2018-04-20
Add `startingInner` property to multiplexing mode state.

AlbertHilb:
This PR intends to fix issue #5368

It adds `startingInner` property to multiplexing mode state.  That property is used, when `parseDelimiters` is true and `open` and `close` parameters are equal, to avoid the same substring starts and ends an inner mode.

Thanks

### DIFF
--- a/addon/mode/multiplex.js
+++ b/addon/mode/multiplex.js
@@ -29,7 +29,8 @@ CodeMirror.multiplexingMode = function(outer /*, others */) {
       return {
         outer: CodeMirror.startState(outer),
         innerActive: null,
-        inner: null
+        inner: null,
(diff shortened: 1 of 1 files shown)


AlbertHilb:
Is there any chance this will be merged or #5368 will be fixed in some other way?
I'll be glad if I can help you to make this happening.

djbarker:
+1 on getting this fixed.

marijnh:
Not sure what happened here, but I've pushed a slightly simpler patch  f015405 that should take care of this. Let me know if it doesn't work for you.

djbarker:
Thanks.

Issue #4307, opened 2016-10-12
Deleting text throws many exceptions in Android Chrome


TWiStErRob:
Google Chrome 53 (latest) on Galaxy S5 running Android 5.0.1; reproducible with both Google Keyboard and Samsung Keyboard.

**Repro**:
1. open http://codemirror.net
2. long-tap inside the demo to select a word (this works around #4306)
3. Try to delete contents of the textarea by pressing backspace on the soft keyboard  

**Actual**: deleting stops after a while, and content cannot really be deleted
**Expected**: deletion continues until no content remains before the cursor

**Note**: keeping it pressed or tapping it repeatedly both fail badly. These errors essentially make CodeMirror unusable on Android. Not only the demo, but other sites that use it are affected. And some of these exceptions are thrown when entering text too.

**Diagnostic steps**:
1. Start Google Chrome on Android device
2. Visit http://codemirror.net on the device
3. Plug in the device to a computer
4. Start Google Chrome on computer
5. Visit chrome://inspect on the computer
6. Select codemirror.net tab to inspect  
   _See https://developers.google.com/web/tools/chrome-devtools/remote-debugging/ for more details if it doesn't work._
7. Do the repro above to see what's going on

The following errors show repeatedly:

> Uncaught TypeError: Cannot set property 'data' of undefined
> (anonymous function) @ codemirror.js:8173

---

> Uncaught TypeError: Cannot read property 'nodeType' of null
> walk @ codemirror.js:8480
> domTextBetween @ codemirror.js:8510
> pollContent @ codemirror.js:8388
> (anonymous function) @ codemirror.js:8194

---

> Uncaught NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
> updateLineGutter @ codemirror.js:2170
> buildLineElement @ codemirror.js:2223
> patchDisplay @ codemirror.js:4147
> updateDisplayIfNeeded @ codemirror.js:4056
> endOperation_W1 @ codemirror.js:3642
> endOperations @ codemirror.js:3619
> (anonymous function) @ codemirror.js:3608
> finishOperation @ codemirror.js:2050
> endOperation @ codemirror.js:3605
> (anonymous function) @ codemirror.js:3750

justbittin:
This is confirmed.  Our mobile site has broken on android.

Codemirror becomes unstable.

justbittin:
Has anyone found a work around for this?  This is a critical bug for android at the moment.

[...]

PR #5379, opened 2018-04-25
Fix nbsp (for space splitting) getting added as real spaces (contentEdtiable mode)

muan:
## Problem

When using IME in `lineWrapping` + `contentEdtiable` mode, typing CJK characters puts the cursor at the wrong position. Steps to reproduce:

1. With existing text like:
  ```
  - 123
     - 456
```
2. Type Chinese characters at the end of Line 1 `123` with an IME (example: "台灣" with Zhuyin that is `w` `9` `6` `j` `0` `(space)`, `(enter)`)
3. See the cursor is put on the next line (after a `&nbsp;`). 

## Cause

This problem only exist when the next line has `&nbsp;` created by `splitSpaces`

https://github.com/codemirror/CodeMirror/blob/9ed217d5b46f98581b7b751758686c5a70ad05d1/src/line/line_data.js#L192-L203

After composition ends & `pollContent`, the `&nbsp;` created to split spaces becomes real content, which I believe causes miscalculation of the cursor position.
## Solution

This PR replaces nbsp with regular spaces in `pollContent`.

Before the fix

```javascript
$ editor.getValue()
> "- 123台灣
  - 321
"
$ editor.getValue().match(/\u00A0/)
> [" ", index: 9, input: "- 123台灣↵  - 321↵", groups: undefined]
```

After the fix

```javascript
$ editor.getValue()
> "- 123台灣
  - 321
"
$ editor.getValue().match(/\u00A0/)
> null
```

---

@marijnh Honestly I don't have enough context to be 100% sure if this is the right fix. It also doesn't seem like there's a reliable way to test for this problem.

It could be worth noting that this problem also goes away with the removal of https://github.com/codemirror/CodeMirror/blob/9ed217d5b46f98581b7b751758686c5a70ad05d1/src/input/ContentEditableInput.js#L261-L262 But I think we do need them. Ref: #4307.

Since you have more context here perhaps you'd know if there's a better fix. 🙏🏻

### DIFF
--- a/src/input/ContentEditableInput.js
+++ b/src/input/ContentEditableInput.js
@@ -431,7 +431,7 @@ function domTextBetween(cm, from, to, fromLine, toLine) {
         walk(node.childNodes[i])
       if (isBlock) closing = true
     } else if (node.nodeType == 3) {
-      addText(node.nodeValue)
+      addText(node.nodeValue.replace(/\u00a0/g, " "))
     }
   }
   for (;;) {


marijnh:
I think the fix is a good idea (and should also fix issues where non-breaking spaces will incorrectly appear in the editor content)—the editor read the next line as containing a non-breaking space when the composition finished, and then made a change that was bigger than the actual composition, which moved the cursor to the end of that change.

PR #5383, opened 2018-04-26
Fix line break disappearing with pasted text in IE11 under contentEditable mode

muan:
## Problem

Line breaks disappear when pasting plain text from nodepad/textarea/command prompt in IE11 in `contentEditable` mode.

Specifically, this is because in `domTextBetween` a `fromNode` like`<span cm-text="">1<br>2<br>3<br>4<br>5</span>` is handled with `addText(node.textContent); return`, ignoring the `<br>`s inside.

This PR does a few things:

- only `addText` & `return` if `cm-text` has value
- otherwise walk the child nodes
- treat `br` along with `table`, `li` as blocks
- ignore empty elements (except for `br` which should be empty) to avoid unnecessary line breaks
- add extra line breaks for `p` and `pre`

FWIW this issue reportedly is also on archlinux/debian in Firefox 52/58, but I have yet been able to reproduce them.

---

As far as I can see changes to this method effects IME input in modern browsers and IE11 plus pasting in IE11. I've manually tested for the IME issues I sent PR for previously (#5379, #5354) and they're still OK.

## Before (IE11)

Text flashes, line breaks disappear.

## After (IE11)

Line breaks preserved.

### DIFF
--- a/src/input/ContentEditableInput.js
+++ b/src/input/ContentEditableInput.js
@@ -396,12 +396,13 @@ function isInGutter(node) {
 function badPos(pos, bad) { if (bad) pos.bad = true; return pos }
 
 function domTextBetween(cm, from, to, fromLine, toLine) {
-  let text = "", closing = false, lineSep = cm.doc.lineSeparator()
+  let text = "", closing = false, lineSep = cm.doc.lineSeparator(), extraLinebreak = false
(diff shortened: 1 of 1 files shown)


marijnh:
That looks reasonable. Thanks!

Issue #3698, opened 2015-12-10
ShowHint `mousemove` is triggered on cycling through the suggestions


basarat:
Started with https://github.com/codemirror/CodeMirror/pull/3636. Only happens if you go across the hints *first -> last* (up arrow) or *last -> first* (down arrow) and if the cursor is in a position over the show hint list. `mousemove` is triggered even if the cursor isn't moved. So instead of going to `first` or `last` it jumps to the cursor location.

Tested on `chrome` on both windows and mac.

Here is a recording showing the issue: 
![show hint error](https://cloud.githubusercontent.com/assets/874898/11730997/85daa856-9fec-11e5-8cd3-105c5d0022a2.gif)

If one puts a `console.log` here : https://github.com/codemirror/CodeMirror/blob/4bdf6440429bc455f4ce39486d424fba0be7fa14/addon/hint/show-hint.js#L300 it will get triggered. Would love your input on how to approach a fix :rose:

marijnh:
Could it be that your mouse is so sensitive that the desktop vibrations caused by hitting the arrow keys cause it to move? There shouldn't actually be any mousemove event firing unless the mouse actually moves, but I guess that could happen.

I think attached patch solves this.

basarat:
> Could it be that your mouse is so sensitive that the desktop vibrations caused by hitting the arrow keys cause it to move?

No, because it only happens on first to last or last to first transition. Also the patch looks empty 🌹 

Thanks for this project 🌹

basarat:
@marijnh I applied the patch locally and its made it worse in that it doesn't work if I use a custom render function:

![broken](https://cloud.githubusercontent.com/assets/874898/11833513/78bc8606-a417-11e5-91e4-8c9918c59228.gif)

And for the simple one it is still broken. I noticed it also jumps if you cause any form of `scroll` to happen on the autocomplete (e.g. moving an item before the first item shown using `up` arrow):

![alsojumpsonscroll](https://cloud.githubusercontent.com/assets/874898/11833540/a948a930-a417-11e5-8c4e-6e70136ad292.gif)

Thanks :rose:

marijnh:
Ugh. Well, this week is not one where I have time to debug other people's code. I've reverted the whole behavior for now. @continuata Do you want to take a look at this?

continuata:
Will do

continuata:
[...]

PR #5385, opened 2018-04-27
[show-hint addon] Add option to change active hint on mouseover

cone56:
Adds an bool option, `changeActiveOnMouseover` (disabled by default), which updates the selected hint index for mouseover events. Using the selected hint variable to maintain the index of the active hint allows the `ACTIVE_HINT_ELEMENT_CLASS` to be consistently applied when the user uses a combination of mouse selections and keyboard while deciding the hint to choose.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -279,6 +279,15 @@
       hints.style.left = (left + startScroll.left - curScroll.left) + "px";
     });
 
+    if (completion.options.changeActiveOnMouseover) {
+      CodeMirror.on(hints, "mouseover", function(e) {
+        var t = getHintElement(hints, e.target || e.srcElement);
+        if (t && t.hintId != null) {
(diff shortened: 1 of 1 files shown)


marijnh:
Sorry for the slow response. We had this feature at one point, but it was removed (06e83fd66eb) because it caused problems (#3698), so I'm not really interested in adding it back again.

cone56:
Looks like your reverted commit was enabling the feature by default. This PR has it disabled by default which should make it safe to include. Additionally, I've not been able to reproduce #3698 with the latest codebase.

yz4:
Has there been any update about this? The mouseover behavior seems reasonable in many use cases and not having the active hint on highlight is actually highly unintuitive if the feature work flow relies more heavily on mouse movements than keyboard cursor. I think having a boolean toggle for the mouseover behavior is a reasonable solution if it was causing problems for some people but not for others.

PR #5389, opened 2018-04-30
Add tests for html auto completion

physikerwelt:


marijnh:
Thank you. Merged as d232e5075

PR #5402, opened 2018-05-11
'dialog-opened' class when dialog is triggered

goldsmcb:
Thanks for considering my other PR about adding insertBefore to the dialog. I think this approach is less invasive to the existing dom structure and allows for styling like this to avoid the dialog covering up the first few lines of text in the editor.
```
.dialog-opened .CodeMirror-scroll {
	margin-top: 40px;
}
```

An advantage I see with this is it gives the application an indication of when a dialog is opened, allowing for many styling opportunities.

### DIFF
--- a/addon/dialog/dialog.js
+++ b/addon/dialog/dialog.js
@@ -25,7 +25,7 @@
     } else { // Assuming it's a detached DOM element.
       dialog.appendChild(template);
     }
-    wrap.classList.add('dialog-opened');
+    CodeMirror.addClass(wrap, 'dialog-opened');
     return dialog;
   }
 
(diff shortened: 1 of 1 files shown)


marijnh:
This is a good idea. However, CodeMirror is targeting pre-`classList` browsers so you'd have to formulate it using the `CodeMirror.addClass` and `CodeMirror.rmClass` utilities instead.

goldsmcb:
Nice I didn't know about the addClass and rmClass utilities, hows this new commit looking?

marijnh:
Much better, thanks!

goldsmcb:
Awesome! Any chance you could tell me when the next release would be?

marijnh:
Next week. Release monthly around the 20th.

jdbruxelles:
@goldsmcb [here](https://codemirror.net/doc/releases.html) is all release notes and version history.

PR #5407, opened 2018-05-16
Replace stream.match(/ +$/, false) with a more efficient regex

paddya:
With some inputs, especially when they contain long lines with many whitespace characters, `stream.match(/ +$/, false)` in the Markdown mode gets terribly slow. In our case, it could lock up the editor for 10s of seconds.

~~As a quick fix, I added a method `StringStream.matchTrailing(character)` which replicates the behavior of `StringStream.match()` for the special case of checking for specific trailing characters.~~

I changed the regex to `^ +$` which improves the performance a lot.

* All tests pass
* No complaints from `bin/lint`

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -619,7 +619,9 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
     }
 
     if (ch === ' ') {
-      if (stream.match(/ +$/, false)) {
+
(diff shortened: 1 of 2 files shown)


marijnh:
Does changing the expression to `stream.match(/^ +$/, false)` also yield the performance improvements that you're looking for?

paddya:
@marijnh Yes! And it is so much simpler. I will change it.

marijnh:
Thanks!

paddya:
Thank you!

PR #5411, opened 2018-05-18
Support BigInt literals in JavaScript

tmcw:
This adds support for BigInt literal syntax: `10n` is highlighted as `[number 10n]` instead of `[number 10][variable n]`. Not sure what the policy on 'when is soon enough' to implement language features - this one's quite fresh: it's a stage 3 proposal, shipped in Chrome Canary. The BigInt interface is implemented in JSC and SpiderMonkey, but afaict, the literal syntax isn't yet.

Todo:

- [x] Support binary, octal, and hexadecimal syntax

References:

- Google Chrome announcement post: https://developers.google.com/web/updates/2018/05/bigint
- tc39 proposal: https://github.com/tc39/proposal-bigint


marijnh:
I've implemented this in 08a764161c instead (which also cleans up some weird code in the number-matching), and added your test in 875913f52d0d

PR #5423, opened 2018-05-23
Add Papyrus mode

azufluup:
Adds a highlighting mode for Papyrus, a scripting language by Bethesda
for video games mods (Skyrim and the Fallout series).

Documentation on the syntax is bit all over the place, but the
CreationKit wiki has lots of examples scattered around:

https://www.creationkit.com/index.php?title=Arrays_(Papyrus)
https://www.creationkit.com/index.php?title=Literals_Reference

Credit to the authors of the Lua mode, which acted as a good base to
work up from.


marijnh:
Hi. Thanks, but we aren't adding any new modes to the main distribution. The recommended way to distribute new modes is as separate npm packages (see for example the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir)).

PR #5425, opened 2018-05-23
Change cursor color in darcula theme to be visible

bjornhansen:
The "darcula" theme added in version 5.38.0 is missing a CSS rule for coloring the cursor, so it falls back to the default black. This makes it difficult to see with its dark gray background. This commit adds a rule that makes the cursor a light gray for better contrast.


marijnh:
Thanks. Merged as dc573b5a

PR #5451, opened 2018-06-08
[markdown] improve for HTMLs

laobubu:
## Commits

[markdown] support line-breaks in tags
[markdown] support multi-line blocks (eg. CDATA and comments)
[markdown] clear unused htmlState

## Example

```
<!-- Comments may contain

blank lines -->

<?php echo "good"; ?>

<![CDATA[ Support CDATA ]]>

<div id="xxx"
class="yyy">
    ↑ line-breaks are accepted between attributes
</div>
```

See:

- https://spec.commonmark.org/0.28/#example-146
- https://spec.commonmark.org/0.28/#example-147
- https://spec.commonmark.org/0.28/#example-149
- https://spec.commonmark.org/0.28/#example-614

## BTW, introducing ... [HyperMD](https://laobubu.net/HyperMD/)

**WYSIWYG Markdown editor** based on CodeMirror.
(Yes, it's *"What You See Is What You Get"*)

 [Demo](https://laobubu.net/HyperMD/)

### DIFF
--- a/mode/markdown/markdown.js
+++ b/mode/markdown/markdown.js
@@ -125,9 +125,10 @@ CodeMirror.defineMode("markdown", function(cmCfg, modeCfg) {
     state.quote = 0;
     // Reset state.indentedCode
     state.indentedCode = false;
-    if (state.f == htmlBlock) {
+    if (state.f == htmlBlock && (!state.htmlState.tokenize || !state.htmlState.tokenize.isInBlock)) {
       state.f = inlineNormal;
       state.block = blockNormal;
+      state.htmlState = null;
     }
     // Reset state.trailingSpace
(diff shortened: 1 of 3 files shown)


marijnh:
Could you add tests for this? I think we can avoid the `isInBlock` kludge by checking for `state.tagName || state.context` but without tests it's hard to be sure that I actually didn't break what you implemented.

laobubu:
Tested. Both `tagName` and `context` are `null`.

laobubu:
- [x] Added test cases
- [x] Fix: compatibility with `htmlmixed` mode *(previously only tested with `xml` mode)*

marijnh:
Merged as 11dfc2f26 and followed up with 9ea789b9

PR #5452, opened 2018-06-09
 Add highlight active line gutter

jdbruxelles:


marijnh:
Changing the default style to make all gutter elements transparent is not acceptable—it'll mess up people's existing setups.

PR #5466, opened 2018-06-20
Fix hint not closing correctly when at beginning of line

40thieves:
Fixes a bug where opening the hint widget at the beginning of a line would not close correctly after deleting. It's somewhat hard to explain, so I've created a small repro case [here](http://jsbin.com/becuselate/edit?html,js,output).

If the cursor started at the beginning of a line and show hint was triggered, deleting a character would not close the hint correctly. I think the issue here is that `pos.ch` is checked for existence but falsy 0 wasn't considered.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -98,7 +98,7 @@
       var pos = this.cm.getCursor(), line = this.cm.getLine(pos.line);
       if (pos.line != this.startPos.line || line.length - pos.ch != this.startLen - this.startPos.ch ||
           pos.ch < this.startPos.ch || this.cm.somethingSelected() ||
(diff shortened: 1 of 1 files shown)


marijnh:
It doesn't look like that was the intention of the check—positions never have an undefined `ch`, and further on in the line it uses `ch - 1`, which will produce a bogus position if `ch` was zero.

I guess we should always close when hitting the start of the line, so making that condition `(!pos.ch || this.options.closeCharacters.test(line.charAt(pos.ch - 1))` is probably a good idea. Could you adjust the patch?

40thieves:
Right ok. Your suggestion makes sense I'll adjust the patch

marijnh:
Great. Merged as d1063d2c

40thieves:
Great. Thanks!

PR #5472, opened 2018-06-25
Support themes in Hints

AlexanderPrendota:
Hello!

This PR for using different themes in hints.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -200,7 +200,8 @@
     var widget = this, cm = completion.cm;
 
     var hints = this.hints = document.createElement("ul");
-    hints.className = "CodeMirror-hints";
+    var theme = completion.cm.options.theme;
+    hints.className = "CodeMirror-hints " + theme;
     this.selectedHint = data.selectedHint || 0;
 
     var completions = data.list;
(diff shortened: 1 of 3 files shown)


AlexanderPrendota:
So, if CodeMirror can not support "`" I can replace to ```var theme = "CodeMirror-Hints " + theme;```

addon/hint/show-hint.js
```diff
@@ -200,7 +200,8 @@
     var widget = this, cm = completion.cm;
 
     var hints = this.hints = document.createElement("ul");
-    hints.className = "CodeMirror-hints";
+    var theme = completion.cm.options.theme;
+    hints.className = `CodeMirror-hints ${theme}`;
```
marijnh: The addons should be ES5, so you can't use template literals here.

AlexanderPrendota: @marijnh okay. I'll change to 
`var theme = "CodeMirror-Hints " + theme;`

AlexanderPrendota:
@marijnh Element.classList is not supported some browsers that's why I use plain string concatenation.

AlexanderPrendota:
@marijnh Yes. Exactly. Right now all is fine. I tested it. =) Thanks!

marijnh:
Great. Merged as 70408ced

Issue #5371, opened 2018-04-23
[kotlin mode] Indentation of nested "get/set" operators


AlexanderPrendota:
Hello, @marijnh 
The indentation of nested "get/set" operators is sometimes incorrect.
Correct indentation:
```
public val <T> List<T>.lastIndex: Int 
    inline get() = this.size - 1
```
CodeMirror (incorrect) indentation:
```
public val <T> List<T>.lastIndex: Int 
inline get() = this.size - 1
```

Also few examples:
Correct: 
```
var counter = 0 
    set(value) {
        if (value >= 0) field = value
    }    
val isEmpty: Boolean
    get() = this.size == 0
```
CodeMirror (incorrect) indentation:
```
var counter = 0 
set(value) {
    if (value >= 0) field = value
}
val isEmpty: Boolean
get() = this.size == 0
```

PR #5487, opened 2018-07-04
[clike] update kotlin indent

AlexanderPrendota:
Hello!

I try to decide several problems with Kotlin indents.

So, the problems are:

1) Indent after operator 
```kotlin
fun main(args: Array<String>) {
    val str =  
    "cm"               
}
```
Correct 
 ```kotlin
fun main(args: Array<String>) {
    val str =  
        "cm"               
}
```

2)  Chained method calls
```kotlin
val allowedEntrance = people
.filter { it.age >= 21 }
.map { it.name }
.take(5)
```
Correct 
 ```kotlin
val allowedEntrance = people
        .filter { it.age >= 21 }
        .map { it.name }
        .take(5)
```

3) Lambdas indents

```kotlin
args.forEach { arg ->
              println(arg)
             }
```
Correct 
 ```kotlin
args.forEach { arg ->
              println(arg)
 }
```

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -216,15 +216,15 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
     indent: function(state, textAfter) {
       if (state.tokenize != tokenBase && state.tokenize != null || state.typeAtEndOfLine) return CodeMirror.Pass;
       var ctx = state.context, firstChar = textAfter && textAfter.charAt(0);
+      var closing = firstChar == ctx.type;
       if (ctx.type == "statement" && firstChar == "}") ctx = ctx.prev;
       if (parserConfig.dontIndentStatements)
         while (ctx.type == "statement" && parserConfig.dontIndentStatements.test(ctx.info))
           ctx = ctx.prev
       if (hooks.indent) {
-        var hook = hooks.indent(state, ctx, textAfter);
+        var hook = hooks.indent(state, ctx, textAfter, indentUnit, closing);
         if (typeof hook == "number") return hook
       }
-      var closing = firstChar == ctx.type;
       var switchBlock = ctx.prev && ctx.prev.info == "switch";
       if (parserConfig.allmanIndentation && /[{(]/.test(firstChar)) {
         while (ctx.type != "top" && ctx.type != "}") ctx = ctx.prev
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by marijnh** [comment]
The idea is that the main clike mode doesn't have language-specific logic. You should implement this as an indent hook instead, if possible. If not, see how you can extend the clike mode to allow extension like that.

**Comment by AlexanderPrendota** [comment]
Hello! I took a look to other languages and I guess this changes don't effect any languages in clike mode.

**Comment by AlexanderPrendota** [comment]
Hm. I think in this changes also need to add double indent for streams in Java and etc. It's dot after `)`

**Comment by marijnh** [comment]
It seems like it should be really easy to put this in an `indent` hook instead. Most clike languages indent multi-line statements by looking at semicolons, so they don't need this logic, and I'd prefer if it didn't run for them at all.

**Comment by AlexanderPrendota** [comment]
@marijnh You mean hooks in `def("text/x-kotlin")` right?

**Comment by AlexanderPrendota** [comment]
Okay! Thanks I will take a look =)

**Comment by AlexanderPrendota** [comment]
@marijnh Tell me please how I can get `indentUnit` variable in an indent hook?

**Comment by marijnh** [comment]
Looks like you can't, currently, but it would make sense to pass that as fourth parameter to `hooks.indent`, when it's called.

**Comment by AlexanderPrendota** [comment]
Also add conditions `state.prevToken == "operator" && textAfter != "}" ` to fix cases like that:
```kotlin
inline fun <reified T> TreeNode.findParentOfType(): T? {
    var p = parent
    while (p != null && p !is T) {
        p = p.parent
    }
    return p as T?
}
```

`if ((state.prevToken == "}" || state.prevToken == ")") && textAfter == "")` - keep writing stream after enter

marijnh:
Thanks! Merged as e338ab23b69dd

AlexanderPrendota:
@marijnh Thank you for your help.
Could you give some advice to fix [5371](https://github.com/codemirror/CodeMirror/issues/5371).

Just use some regex in indent hook?

marijnh:
I have no idea what that syntax does (never wrote Kotlin), but yeah, looking ahead with a regexp in the indent hook seems like it might work.

AlexanderPrendota:
@marijnh it's just getter and setter of property. Like in C#;

Also. I just noticed. We have some problems with indents with condition and loops:
```kotlin
if (a > b)
    return a
```
render: 
```kotlin
if (a > b)
return a
```
In JS mode this case works fine?

PR #5490, opened 2018-07-08
Add 'force' option to autorefresh addon

yuki-takei:
Purpose
---------

* Add 'force' option to autorefresh addon

### Usage

```javascript
var editor =CodeMirror.fromTextArea(document.getElementById("taCodemirror"), {
  autoRefresh: { delay: 500, force: true },
});
```

Motivation
-----------

Please see https://github.com/scniro/react-codemirror2/issues/83#issuecomment-398825212

### DIFF
--- a/addon/display/autorefresh.js
+++ b/addon/display/autorefresh.js
@@ -16,7 +16,7 @@
       stopListening(cm, cm.state.autoRefresh)
       cm.state.autoRefresh = null
     }
-    if (val && cm.display.wrapper.offsetHeight == 0)
+    if (val && (val.force || cm.display.wrapper.offsetHeight == 0))
       startListening(cm, cm.state.autoRefresh = {delay: val.delay || 250})
   })
 
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -2954,7 +2954,13 @@ editor.setOption("extraKeys", {
       500}</code> as the option value to configure this). Note that
       this addon will only refresh the editor <em>once</em> when it
       first becomes visible, and won't take care of further restyling
-      and resizing.</dd>
+      and resizing.<br>The option's value may be a boolean or an object specifying the following options:
+      <dl>
+        <dt><code><strong>delay</strong>: number</code></dt>
+        <dd>Delay before starting polling (msec)</dd>
+        <dt><code><strong>force</strong>: boolean</code></dt>
+        <dd>Force to start polling to refresh even if after the CodeMirror element has already initialized.</dd>
+      </dl></dd>
 
       <dt id="addon_simplescrollbars"><a href="../addon/scroll/simplescrollbars.js"><code>scroll/simplescrollbars.js</code></a></dt>
       <dd>Defines two additional scrollbar


marijnh:
It seems like this will just asynchronously call `refresh`—if the editor is visible, it'll probably still be visible the first time the timeout fires. What use case does the new option address?

yuki-takei:
@marijnh 
Thank you for your reply.
I do not know details about what conditions trigger such a situation. The following are guesses.

## Case A

1. `CodeMirror.defineOption("autoRefresh", ...` is invoked first
1. `CodeMirror.fromTextArea` is invoked

It's simple and works fine.

## Case B

But we ( scniro/react-codemirror2 users including me ) are using webpack and babel, confronted with unfamiliar situation.

Maybe..

1. A react-codemirror2 instance is initialized
    1. CodeMirror DOM is visible
    1. `cm.display.wrapper.offsetHeight` set
1. `CodeMirror.defineOption("autoRefresh", ...` is invoked
    1. `startListening` is **NOT** invoked

I found this situation by debugging with browser.

One thing I can say for sure is that this option will help us and doesn't be evil for others.

marijnh:
I think you can just call `.refresh()` on your editor directly then, and avoid the whole autorefresh situation. Or, alternatively, make the autorefresh addon a dependency of the module that initializes the editor, to make sure it's loaded first. This option seems like a confusing kludge, and I don't want to include it in the distribution.

yuki-takei:
# Why do you have the confirmation?

@marijnh 

> you can just call .refresh() on your editor directly

We want to do it if it is easy.

> make the autorefresh addon a dependency of the module that initializes the editor, to make sure it's loaded first

We really want to do it if it is easy. Please see scniro/react-codemirror2#83 .

> This option seems like a confusing kludge

Exactly. The current uncertain and defective code produces this PR.

Conversely, how about it?

## Before (Current)
```
if (val && cm.display.wrapper.offsetHeight == 0)
  startListening(cm, cm.state.autoRefresh = {delay: val.delay || 250})
```

## After (New proposal)
```
if (val && cm.display.wrapper.offsetHeight == 0 && cm.display.lastWrapHeight != cm.display.wrapper.clientHeight)
  startListening(cm, cm.state.autoRefresh = {delay: val.delay || 250})
```

## The defect of current code

- I agree that the instance might be already refreshed if `cm.display.lastWrapHeight` equals `cm.display.wrapper.clientHeight`
    - It is checked in [`check() function`](https://github.com/yuki-takei/CodeMirror/blob/4689cbe3deb9388784a6a89e8ac6963f9361755b/addon/display/autorefresh.js#L27)
- **Why do you have the confirmation that `cm.display.lastWrapHeight` ALWAYS EQUALS `cm.display.wrapper.clientHeight` when `cm.display.wrapper.offsetHeight == 0`?**
    - It is possible that the relationship does not hold.
    - It is possible that `cm.display.wrapper.offsetHeight` changes multiple times in initializing process in complex environments
        - e.g. using with react, angular, splitting chanks with webpack, packing DLLs with webpack, lazy loading some modules ...


## Pros of the new proposal
- Certainly refreshed even if under the situation that `cm.display.wrapper.offsetHeight` is not zero and the instance is not refreshed yet
- You need not to add 'force' option
## Cons of the new proposal
- **Nothing**
    - How much processing costs are increased?

PR #5501, opened 2018-07-13
Fix #5486

### LINKED ISSUE #5486: [bug] cursor locating wrongly, if LineWidget(below a collapsed line) changed
Reproduction: https://codepen.io/laobubu/pen/BVEemW

1. assuming there are 3 lines (0,1,2)
2. `addLineWidget` below line 1
3. `markText` to **collapse line 0 and line 1**
4. change the lineWidget's size, increasing its height
5. call `lineWidget.changed()`
6. **chars and cursor can't be correctly located since line 2**

However, if line 1 is modified(eg. adding some char), things goes back to normal. Maybe a measurement cache problem?

laobubu:


marijnh:
Thanks for looking into that! Merged.

PR #5505, opened 2018-07-21
Use setImmediate(fn) Instead Of setTimeout(fn, 0)

anonyco:
IE10+ and Edge provide native non-blocking setImmediate and, for non-supporting browsers, the messageChannel API can provide a non-blocking temporary polyfill. Non-blocking means that DOM updates, HTTP requests, and everything else is able to happen in between the time setImmediate is called, and its call back is executed. Thus, a non-blocking setImmediate would provide the exact same functionality as setTimeout(fn, 0) INCLUDING the below improvements.

* setImmediate could potentially make CodeMirror more responsive in some cases because although 4ms is less than most monitors framerates, things built on top of CodeMirror or that extend CodeMirror would be given 4ms more to process before there is a "lag-frame" which fails to update because the CPU is swamped.
* setImmediate put into CodeMirror at this critical juncture will make CodeMirror several times more power efficient because instead of the browser having to swamp the CPU with operations to get 4ms resolution timing, the function will be executed right away, leaving the CPU able to "idle-out" and conserve lots of power.
* Lastly, I would like to explain my reasoning for only including the setTimeout backup in this change without also including postMessage. Indeed, Microsoft has taken windows Vista off extended support, clearing the crowds of people quite a lot. Yes, it is true that you could elaborate on this polyfill and extend support to IE9 by adding a postMessage fallback, however you would increase the size of CodeMirror by an appreciable degree, run the risk of conflicting with other javascript that uses postMessage, and break IE8 compatibility because IE8 does synchronous postMessage. For all these reasons, I would say that the postMessage implementation presented in this change is good as is, and should not be modified to add in extra compatibility.

Therefore, because this setImmediate change will make CodeMirror more responsive and more power efficient, all without harming the IE-crowd, I submit that this major change should be accepted right away and not discarded.

### DIFF
--- a/src/util/operation_group.js
+++ b/src/util/operation_group.js
@@ -40,6 +40,33 @@ export function finishOperation(op, endCb) {
   }
 }
 
+// The below implementation of setImmediate is intended to be only one-way. For the
+// purposes of where this function would be used in CodeMirror, adding in support
+// for clearImmediate would be non-sensible because it would never be used, only 
+// only bloat the code size, make setImmediate run a little slower than otherwise.
+const setImmediateOrFallBack = (function(){
+    if (typeof setImmediate !== "undefined") return setImmediate // for IE10+
+    if (typeof MessageChannel === "undefined") return setTimeout // for IE9 and below
+    const msgChannel = new MessageChannelport1 = msgChannel.port1, port2 = msgChannel.port2
+    const postMessage = port2.postMessage.bind(port2)
+    var waitingImmediates = null, immediatesLen=0
+    port1.onmessage = function(){
+        var curImmediates = waitingImmediates, i = 0
+        waitingImmediates = null
+        for (; i !== immediatesLen; i++) curImmediates[i]()
+    }
+    return function(f){
+        if (waitingImmediates === null) {
+            waitingImmediates = [f]
+            postMessage(void 0)
+            immediatesLen = 1
+        } else {
+            waitingImmediates.push(f)
+            return immediatesLen++
+        }
+    }
+})()
+
 let orphanDelayedCallbacks = null
 
 // Often, we want to signal events at a point where we are in the
@@ -59,7 +86,7 @@ export function signalLater(emitter, type /*, values...*/) {
     list = orphanDelayedCallbacks
   } else {
     list = orphanDelayedCallbacks = []
-    setTimeout(fireOrphanDelayed, 0)
+    setImmediateOrFallBack(fireOrphanDelayed)
   }
   for (let i = 0; i < arr.length; ++i)
     list.push(() => arr[i].apply(null, args))
--- a/src/util/operation_group.js
+++ b/src/util/operation_group.js
@@ -44,27 +44,32 @@ export function finishOperation(op, endCb) {
 // purposes of where this function would be used in CodeMirror, adding in support
 // for clearImmediate would be non-sensible because it would never be used, only 
 // only bloat the code size, make setImmediate run a little slower than otherwise.
-const setImmediateOrFallBack = (function(){
-    if (typeof setImmediate !== "undefined") return setImmediate // for IE10+
-    if (typeof MessageChannel === "undefined") return setTimeout // for IE9 and below
-    const msgChannel = new MessageChannelport1 = msgChannel.port1, port2 = msgChannel.port2
-    const postMessage = port2.postMessage.bind(port2)
-    var waitingImmediates = null, immediatesLen=0
-    port1.onmessage = function(){
-        var curImmediates = waitingImmediates, i = 0
-        waitingImmediates = null
-        for (; i !== immediatesLen; i++) curImmediates[i]()
+const setImmediateOrFallBack = (() => {
+  if (typeof setImmediate !== "undefined") return setImmediate // for IE10+
+  if (typeof MessageChannel === "undefined") return setTimeout // for IE9 and below
+  const msgChannel = new MessageChannel
+  const port1 = msgChannel.port1
+  const port2 = msgChannel.port2
+  const postMessage = port2.postMessage.bind(port2)
+  var waitingImmediates = null
+  var immediatesLen = 0
+  port1.onmessage = () => {
+    var curImmediates = waitingImmediates
+    waitingImmediates = null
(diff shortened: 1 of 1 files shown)


marijnh:
Could you remove the big fallback and just use `setTimeout` on any platform that doesn't have `setImmediate`? It seems more code than is justified by having a slightly better polyfill on a few old platforms. Also, are you sure that not passing a second argument to `setTimeout` is supported and behaves as passing 0 on all platforms?

The-Peppester:
(P.S. I am the same guy as AnonyCo, but on a different P.C.)

Marinjh, let me first answer your question about the fall back. As [seen here](https://caniuse.com/#feat=setimmediate), direct native setImmediate has unfortunately been removed from the standard. and very few browsers support it. Thus, a polyfill/alternative is required. However, I can assure you that this alternative setImmediate I have put in is just as good as the real thing. To prove this, open up a new tab in your browser at [example.com](https://example.com). Then, run the below code in the console.

```Javascript
(() => {
  "use strict";
    const setImmediateOrFallBack = (() => {
      if (typeof setImmediate !== "undefined") return setImmediate // for IE10+
      if (typeof MessageChannel === "undefined") return setTimeout // for IE9 and below
      const msgChannel = new MessageChannel
      const port1 = msgChannel.port1
      const port2 = msgChannel.port2
      const postMessage = port2.postMessage.bind(port2)
      var waitingImmediates = null
      var immediatesLen = 0
      port1.onmessage = () => {
        var curImmediates = waitingImmediates
        waitingImmediates = null
        for (var i = 0; i !== immediatesLen; i++) {
          curImmediates[i]( /*no arguments*/ )
        }
      }
      return f => {
        if (waitingImmediates === null) {
          waitingImmediates = [ f ]
          postMessage( undefined )
          immediatesLen = 1
        } else {
          waitingImmediates.push( f )
          immediatesLen += 1
        }
      }
    })()
    var i = 1000000
    console.log("starting!");
    setImmediateOrFallBack(function doNext(){
    if (i--) {
      setImmediateOrFallBack(doNext)
        } else {
      console.log("All Done!")
        }
    })
})()
```

After running the above code in the console, try interacting with the page (selecting text and what-not). Then, open up a task manager and observe that the page is using up 100% of the CPU power for its thread (100/#virtual_threads% of the total CPU power). Now, try closing the page and observe that even page close events are not blocked. As you can see, this polyfill for setImmediate is non-blocking which is incredibly good for the UI while at the same tie saving power because there is only a minimal delay. As for browsers that natively support a setImmediate like this, which are IE10+ and Edge, I think that it would be a shame if we did not take advantage of native support in those browsers. 

As for your second question about the second argument to setTimeout, I am pretty certain that the second argument to setTimeout is coerced to an integer. Thus, when the second argument is undefined, that undefined is coerced to an integer, resulting in 0 like so.

```Javascript
undefined | 0 === 0 // is true
```

To prove this for absolute certain, I did tests in the two browsers that will use the setTimeout fallback: IE8 and IE9. I used IETester and the result (shown below in images) conclusively prove that when there is no second argument to setTimeout, setTimeout behaves as if the second argument is 0. Please open the images in a new tab for a better view of them.

Internet Explorer 8:
ie8 settimeout test results

Internet Explorer 9:
ie9 settimeout test results

marijnh:
I really do not want that big polyfill in the code for the obsure, rare case where this `setImmediate` implementation is actually being used. Is there a concrete issue this patch is solving, or did you grep for 0-ms timeouts?

anonyco:
marijnh, for the issue of the polyfill size, I have run my code through the closure compiler and, as seen below, there will only be a 367 byte increase (before gzip is applied).

```Javascript
var constsetImmediateOrFallBack=function(){if("undefined"!==typeof setImmediate)return setImmediate;if("undefined"===typeof MessageChannel)return setTimeout;var e=new MessageChannel,f=e.port2,g=f.postMessage.bind(f),a=null,c=0;e.port1.onmessage=function(){var b=a;a=null;for(var d=0;d!==c;d++)b[d]()};return function(b){null===a?(a=[b],g(void 0),c=1):(a.push(b),c+=1)}}();
```
If you want to sacrifice a little performance for a smaller size, then we can even make that 240 bytes before gzipped.

Source Code (564 bytes):
```Javascript
const setImmediateOrFallBack = self.setImmediate /*IE10+*/ || (MessageChannel => { // for all others
  if (!MessageChannel) return setTimeout; // for IE9 and below
  const msgChannel = new MessageChannel
  const waitingImmediates = []
  var immediatesLen = 0;
  msgChannel.port1.onmessage = () => {
    while (immediatesLen) {
      waitingImmediates.shift()( /*no arguments*/ )
      immediatesLen--
    }
  }
  return f => {
    if (!immediatesLen) msgChannel.port2.postMessage( 0 )
    immediatesLen = waitingImmediates.push( f )
  }
})(self.MessageChannel)
```
Compressed (240 bytes without gzip):
```Javascript
window.setImmediateOrFallBack=self.setImmediate||function(a){if(!a)return setTimeout;var c=new a,d=[],b=0;c.port1.onmessage=function(){for(;b;)d.shift()(),b--};return function(a){b||c.port2.postMessage(0);b=d.push(a)}}(self.MessageChannel);
```

Next, for the issue of how this could be practical, please see my code. This setImmediate function will be used in every single signalLater call. signalLater, as I have seen frequently in the CodeMirror source code, is used quite frequently. With this change that I am proposing, every single place in the entire CodeMirror code that uses signalLater will receive a major power-saving performance benefit that will make CodeMirror a lot more appealing to many more people. For me personally, I would consider 240 un-gzipped bytes of code to be just a small, but very worthy, sacrifice for tremendous power saving that I would gladly implement.

Lastly, there is no issue. This proposal does not fix any code since the code was already working before. Rather, this proposal simply makes the existing code better.

marijnh:
> This setImmediate function will be used in every single signalLater call.

That's not the case, though—almost all `signalLater` calls are used under an operation, in which case the `setTimeout` doesn't run at all.

As such, I'm going to close this as adding more complexity than warranted.

anonyco:
Marijn, thank you for clarifying that. I now understand. I am very glad that you have reached the best decision for the future of CodeMirror. Thank you for your time.

Issue #5518, opened 2018-08-07
panel addon: when replacing the only existing panel, the new panel also is removed.


nkappler:
When calling `addPanel` with the option `replace` and the panel that will be replaced is the only panel, the whole panel container is removed, including the just newly created panel.

this bug can also be observed on the demo page (https://codemirror.net/demo/panel.html), steps to reproduce:
1. delete panel no. 2. There should be now only panel no. 1
2. click on "replace panel no. 1" button
now panel no. 1 disappears (as expected), but panel 3 does not show up.

I have narrowed it down to the following behaviour:
when calling `addPanel(panel3, {replace: panel1})`, panel 3 gets created and appended, but `info.panel` counter is not yet increased by one. (this is done at the end of `addPanel`.)
now `panel1.clear()` gets called, the counter which is still at 1 even though it should be 2, gets decreased to 0 and thus `removePanels()` is called inside the `Panel.prototype.clear()` funtion

marijnh:
Fixed via #5519

PR #5519, opened 2018-08-07
Fix for issue 5518

nkappler:
See description of issue here:
https://github.com/codemirror/CodeMirror/issues/5518


marijnh:
Thanks!

PR #5523, opened 2018-08-10
Outdated comment block

nozen:
I think the ( ) commenting is from a previous version, the new method should be { }, please confirm.

I have tested with latest version, works for () and {}.


marijnh:
Done in #6306

Issue #5210, opened 2018-01-26
addon/search/search.js contains hardcoded English text and cannot be translated


chrisdeeming:
Forgive me if this has been brought up before, but a search didn't come up with anything.

The `addon/search/search.js` file contains hard coded text such as this:

```js
var queryDialog =
    '<span class="CodeMirror-search-label">Search:</span> <input type="text" style="width: 10em" class="CodeMirror-search-field"/> <span style="color: #888" class="CodeMirror-search-hint">(Use /re/ syntax for regexp search)</span>';
```

We're using CodeMirror within an application that allows all text to be translated from English and currently areas such as this cannot be covered by that system.

Are there any plans to implement a way to translate such text? Is there a reason why that hasn't been done, yet? If we wanted to fork and add code to support this ourselves, would a pull request be approved for such a feature?

Many thanks

marijnh:
> Are there any plans to implement a way to translate such text? Is there a reason why that hasn't been done, yet?

There are no plans by us to continue expanding the search addon. It's more or less a proof of concept that got out of hand.

> If we wanted to fork and add code to support this ourselves, would a pull request be approved for such a feature?

Yes, a pull request that makes the pieces of text in the dialogs configurable would be welcome.

PR #5531, opened 2018-08-15
UI language independence

Kier:
Per https://github.com/codemirror/CodeMirror/issues/5210 this modification allows CM to check for the existence of a variable called CodeMirrorPhrases, and use text stored within it in place of the standard text in dialogs etc.

Hard-coded text has been wrapped with a `CodeMirror.phrase(phraseText)` method, for example:
`txt = '<span>Replace with:</span>'`
becomes
`txt = '<span>' + CodeMirror.phrase('Replace with:') + '</span>'`

Inside src/edit/CodeMirror.js, `CodeMirror.phrase()` checks for the existence of `window.CodeMirrorPhrases['Replace with:']` and uses its contents if available, otherwise it will simply return `Replace with:`.

Initially, I had an array of phrases being passed as part of the options object, but all the addons evaluate their text prior to options being inspected, so the phrases were not available prior to the addon text being evaluated, hence the final approach using the CodeMirrorPhrases global variable.


marijnh:
This is a good idea. I'd prefer the phrases to live in `CodeMirror.phrases` (initialized to null) rather than using a global. Also, could you please not change the quote style of the strings (the rest of the code uses double quotes).

Kier:
Implemented.

As noted in my most recent commit message, it would be great to be able to defer evaluation of text in addon/*.js until initialisation, so that we could pass `phrases` as a property of `options` when calling `CodeMirror.fromTextArea(..., options)`, such that it would be possible to override the phrases even after the addon scripts have been included.

Kier:
A complete change of approach, following my previous comments.

By changing (almost) all instances of the .phrase() call to exist within functions, we *can* defer evaluation of the text until init.

Therefore, phrases can now be passed as a property of the `options` object passed to functions like `CodeMirror.fromTextArea(..., options)`, and an option handler does the work of adding them to the CodeMirror.phrases collection.

For example:

```javascript
this.editor = CodeMirror.fromTextArea(myTextarea, {
	phrases: {
		"Jump to line:": "Hop to line number:",
		"No": "Nope",
		"Push to left": "⬅︎",
	}
});
```

There is one issue - in `addons/merge.js->buildGap()` we have `lock.title = "Toggle locked scrolling";`, which at this point can not make use of the phrase system, because it is called prior to an editor being instantiated, so `CodeMirror.phrase()` (or in this case, `dv.mv.editor().phrase()` is not available for use. I'm stuck on this one, and so have simply added a comment in the code to reflect the problem and left the text hard-coded.

marijnh:
Thanks. Merged as c9892c85b02. Patch be4793137cf8066 simplifies the functionalty (no more separate object or `setPhrases` method—the option entirely determines the set of phrases) and fixes the issue in merge.js.

vasanthv16:
This is a nice feature. Thanks for implementing it.

I do have a suggestion. The phrase(phraseText) function looks for a phrase in the phrases variable.

`let phrases = this.options.phrases`
`return phrases && Object.prototype.hasOwnProperty.call(phrases, phraseText) ? phrases[phraseText] : phraseText`

In addition to that, can you also provide a more dynamic way to look up a phrase? In other words, can phrase(phraseText) look to see if this.options.phrases is a function and call it if it is a function, instead of assuming that it will be an array?

Kier:
That’s a nice idea, I’m away until Sept 4th but if nobody has implemented
it before then I will do so.

PR #5536, opened 2018-08-22
the shorthand if else is causing ReferenceError b is not defined when…

gcziprusz:
the shorthand if else is causing ReferenceError b is not defined when using babel-loader


marijnh:
You're just adding braces. I doubt babel-loader can't handle code without braces, and if it can't, that would be a bug in babel-loader, and not something that we should be changing CodeMirror's code for.

PR #5547, opened 2018-08-26
Fixes bug where vim new line creates a permanent thick cursor

goofiw:


adrianheine:
Thanks for your contribution! I can reproduce the issue and your changes indeed fix it. I wonder what the root cause for this issue is and if there's a better way to fix it, though.

marijnh:
I agree that this looks like it's fixing the issue on the wrong level. Could you say a bit more about how you diagnosed this and why making the mode change asynchronous helps?

goofiw:
@marijnh honestly I don't remember what it was.  I think it had to do with the way operations (src/display/operations.js) were getting batched, events were happening out of order and the <span> that is the thick curser wasn't getting removed properly.

It was 6 weeks ago, but I think another issue of using the operations module was the operation functionality wasn't exported.

Setting the timeout is just a hack for putting off the new line command to the next pass of the run loop so it happens after the current reconciliation of the render steps (and, yea, does another pass through the render steps).

The

PR #5550, opened 2018-08-29
[clojure mode] Provide `hintWords`

joe-loco:
This provides default `hintWords` when auto-completing in Clojure code.


marijnh:
Thanks!

PR #5556, opened 2018-08-29
[javascript mode] allow default values in array destructuring

tmcw:
This permits default assignments in array destructuring patterns, so:

```js
let [a = 10, b] = [];
function z([a = 10, b]) { return [a, b]; }
let y([a = 10, b]) => { return [a, b]; }
```

Currently the parser bails out after an assignment like `a = 10` in this example, which means that any other parameters (like `b`) are not registered.


marijnh:
Thanks for catching that. You can also have default values on non-variable patterns, though. I've applied patch eaf3a7ec

PR #5558, opened 2018-09-01
[clojure mode] Add tests.

finalfantasia:
Besides adding tests, this PR also fixes a few minor issues revealed by the tests:
- Keywords or numbers following a single quote (`'`) were being styled as `variable` instead of `atom` or `number`, respectively.
- Single quote (`'`) was missing on the list of valid symbol characters. Such symbols include `+'`, `*'` and, `-'`.
- `new` and `when-some` were missing on the keyword list. There are a few other symbols (e.g., `as->`, `cond->`, `condp->`, `some->` etc.) that are also missing on the list which I plan to add soon.
- `tests.keyword_char` and `tests.symbol` had redundant characters in them.
- `+` and `+'` were missing on the core symbols list as a regression introduced by commit 74fca7c .

Please note that there are some `FIXME`es in the tests. I'm planning to fix them soon once this PR is accepted.


marijnh:
Thanks! More test coverage is always good to have.

PR #5560, opened 2018-09-02
[clojure mode] Further clean-up.

finalfantasia:
This PR:
- Highlights special forms and core symbols as `keyword`, following the same convention used by the Common Lisp mode and Scheme mode;
- Removes indentation rules for non-core symbols to lay the ground for covering more core symbols for indentation.

### DIFF
--- a/mode/clojure/clojure.js
+++ b/mode/clojure/clojure.js
@@ -117,11 +117,7 @@ CodeMirror.defineMode("clojure", function (options) {
         // Binding forms
         "let", "letfn", "binding", "loop", "for", "doseq", "dotimes", "when-let", "if-let",
         // Data structures
-        "defstruct", "struct-map", "assoc",
-        // clojure.test
-        "testing", "deftest",
-        // contrib
-        "handler-case", "handle", "dotrace", "deftrace"];
+        "defstruct", "struct-map", "assoc"];
 
     CodeMirror.registerHelper("hintWords", "clojure",
         commonAtoms.concat(commonSpecialForms, commonCoreSymbols));
--- a/mode/clojure/clojure.js
+++ b/mode/clojure/clojure.js
@@ -17,8 +17,10 @@
 "use strict";
 
 CodeMirror.defineMode("clojure", function (options) {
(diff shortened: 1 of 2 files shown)


marijnh:
>Removes indentation rules for non-core symbols to stay focused on covering more core symbols for indentation.

I don't quite get this—how does removing those symbols help cover more core symbols?

finalfantasia:
Oh, my bad. I should have articulated my thoughts more clearly.

My focus for this round of refactoring is on special forms and core symbols and the goal is to make syntax highlighting more consistent (by using the same style) and indentation as complete as possible (by covering the remaining core symbols). I expect this to take a little while given the number of remaining symbols in the core namespace and therefore I would like to be "laser-focued" and not distracted by other concerns (i.e., one namespace at a time).

The next step will be thinking about covering symbols from non-core namespaces. I don't know if we should yet since both the Common Lisp mode and the Scheme mode don't seem to do so. If we somehow decide to cover these namespaces, I would like to apply the same approach and do it one namespace at a time while maintaining the same level of consistency and completeness.

So that's the plan and that's why I removed those symbols for now in order to add of them back later (maybe) with better quality of coverage.

marijnh:
>  both the Common Lisp mode and the Scheme mode don't seem to do so

The CL mode has a heuristic for (see `assumeBody`) for when to use body-style indentation versus argument-style indentation. Maybe something like that would be appropriate here too? Though I don't know what the Clojure community's conventions look like.

I guess the number of removed symbols in this PR is small, and any regression it might cause obscure, so I'll go ahead and merge.

finalfantasia:
> The CL mode has a heuristic for (see assumeBody) for when to use body-style indentation versus argument-style indentation. Maybe something like that would be appropriate here too?
Yes and that's exactly what I've been thinking about doing (instead of completely relying on a static list of symbols). The Common Lisp mode is a very good example and starting for this. I've been experimenting with tweaking it to syntax highlight Clojure code and it feels very promising. I might end up re-writing the Clojure mode based on the code of the Common Lisp mode since it is relatively easier to reason about than the current Clojure mode code.

PR #5566, opened 2018-09-06
[clojure mode] The Grand Refactoring of the Clojure Mode

finalfantasia:
I refactored the entire Clojure mode based on the Common Lisp mode so that it's now much easier to reason about thanks to the separation of concerns (styling and indentation). As part of this PR, I also polished the example code so that its code style is up to date.


finalfantasia:
Before:
before

After:
after

marijnh:
Nice, thanks!

finalfantasia:
Thank _you_ for creating and actively maintaining CodeMirror and making it a pleasure to contribute code to it! 🙂

PR #5568, opened 2018-09-07
Do not autoclose single quotes in scala mode

jasongrout:
A single quote in scala indicates the start of a symbol literal (like 'x), not a string, so should not be paired with a closing single quote.

See https://www.scala-lang.org/files/archive/spec/2.12/01-lexical-syntax.html#string-literals and https://www.scala-lang.org/files/archive/spec/2.12/01-lexical-syntax.html#symbol-literals for more information.


marijnh:
Thanks!

Issue #1896, opened 2013-10-20
Add support for alternative comment syntaxes in Clojure and Clojurescript


chrisfjones:
In Clojure and Clojurescript:
Codemirror currently supports line comments:
```clojure
; line comment
```

... but not the comment macro
```clojure
(comment
  (println "this is a comment macro example"))
```

... nor the form comment reader macro
```clojure
#_ (do
     (println "this is a comment reader macro example"))
```

marijnh:
I'm going to close this. Patches welcome, but it doesn't seem all that critical.

Issue #2314, opened 2014-02-28
clojure mode should recognize special comment forms


Gozala:
In clojure,  `form` typed after #_ is ignored by a reader. Which is kind of identical of
`(comment form)` (although comment can take multiple forms).

Here is an example:

``` clj
(defn foo
  [& etc]
  ;(map + 0 etc)
  #_(reduce + 0 etc)
  (comment (reduce + 0 etc))
  (reduce + 0 etc))
```

I think second and third expressions should show up as slightly faded out, to indicate the fact that they are ignored.
I think codemirror should reco

PR #5570, opened 2018-09-08
[clojure mode] Highlight forms wrapped by `(comment ...)` as `comment`.

finalfantasia:
This PR partially resolves issues #1896 and #2314.


marijnh:
Not too sure about abstracting `===` into a function, but sure, the mode seems to be improving so I won't nitpick.

finalfantasia:
Yeah, that was the only part I felt I had gone too far with and was not 100% certain either. I suppose I can limit the usage of `is` to the scope of the tokenizers only. But please feel free to point out anything that doesn’t feel right. I’m open to any idea that helps improve code quality, readability, and maintainability.

PR #5572, opened 2018-09-08
[clojure mode] Use heuristics for identifying symbols with well-known patterns for indentation.

finalfantasia:


mode/clojure/clojure.js
```diff
@@ -137,23 +137,18 @@ CodeMirror.defineMode("clojure", function (options) {
...
   CodeMirror.registerHelper("hintWords", "clojure",
       commonAtoms.concat(commonSpecialForms, commonCoreSymbols));
 
   var atom = createLookupMap(commonAtoms);
   var specialForm = createLookupMap(commonSpecialForms);
   var coreSymbol = createLookupMap(commonCoreSymbols);
-  var indentSymbol = createLookupMap(commonIndentSymbols);
-  var assumeBody = /^(?:def|with)[^\/]+$|\/(?:def|with)/;
+  var indentCoreSymbols = createLookupMap(commonIndentCoreSymbols);
+  var assumeBody = /^cond|^def|^do|^extend|^if|^let|^when|^while|^with|\/(?:def|with)/;
```
marijnh: I think it'd make sense to use a `\b` marker after/around these words, so that things like `letter-at` or `default-for` (just making things up) don't match

finalfantasia: Good point! I’ll take care of that later today. Thanks!

finalfantasia: Welp, it turned out this is not as easy as I originally thought. I need to spend more time to think about how to do this right. I'll close this PR for now. Thanks!

PR #5580, opened 2018-09-13
Make text more clear

guraga:


doc/manual.html
```diff
@@ -946,6 +946,9 @@ <h2>Commands</h2>
       <dt class=command id=command_delWrappedLineLeft><code><strong>delWrappedLineLeft</strong></code><span class=keybinding>Cmd-Backspace (Mac)</span></dt>
       <dd>Delete the part of the line from the left side of the visual line the cursor is on to the cursor.</dd>
 
+      <dt class=command id=command_delLineRight><code><strong>delLineRight</strong></code></dt>
```
marijnh: This command doesn't exist, so documenting it seems like a bad idea.

doc/manual.html
```diff
@@ -2066,8 +2069,8 @@ <h3 id="api_misc">Miscellaneous methods</h3>
       instance, you can call this method with a function argument. It
       will call the function, buffering up all changes, and only doing
       the expensive update after the function returns. This can be a
-      lot faster. The return value from this method will be the return
-      value of your function.</dd>
+      lot faster. This method return value, returned by your
+      function.</dd>
```
marijnh: "This method return value, returned by your function." isn't a grammatically correct sentence.

marijnh:
Thanks. I added some comments, could you fix those?

marijnh:
I've merged the non-problematic parts as e0fae825

PR #5582, opened 2018-09-18
[clojure mode] Clojure mode can now indicate invalid tokens.

finalfantasia:
Types of invalid tokens that can be indicated include:
- invalid numbers (e.g., `42a`, `0x0042m`, etc.)
- invalid character literals (e.g., `\ab`, `\a1`, `\newlines`, `\NEWLINE`)
- invalid symbols (e.g., `42foo`, etc.)

### DIFF
--- a/mode/clojure/clojure.js
+++ b/mode/clojure/clojure.js
@@ -156,23 +156,34 @@ CodeMirror.defineMode("clojure", function (options) {
   var specialForm = createLookupMap(specialForms);
   var coreSymbol = createLookupMap(coreSymbols);
   var hasBodyParameter = createLookupMap(haveBodyParameter);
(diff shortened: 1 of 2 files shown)


finalfantasia:
whoopsie... Node.js doesn't seem to support the `u` flag yet.
```
mode/clojure/clojure.js: Invalid regular expression flag (161:26)
SyntaxError: Invalid flag supplied to RegExp constructor.
```
I'll push another commit with a fix tonight.

marijnh:
We're still targeting ECMAScript 3 with this library, so sorry, can't use any halfway modern language features

finalfantasia:
It’s alright. 🙂

It was my fault that I had run the tests only in Firefox before I pushed the offending code. I removed it in the last commit. Is there anything else in this PR that I should downgrade to ES 3?

kernelp4nic:
Hi, after this patch was applied, commas in clojure are now marked as syntax errors.

Commas are optional in EDN notation and Clojure, from docs:
>Elements are generally separated by whitespace. Whitespace, other than within strings, is not otherwise significant, nor need redundant whitespace be preserved during transmissions. Commas , are also considered whitespace, other than within strings.

I can reproduce this on the official Codemirror Clojure mode demo:

image

Issue #5574, opened 2018-09-09
Scheme mode quoted items get colorized improperly


dsblank:
Scheme mode incorrectly shows keywords in quoted list as highlighted. Notice that the `and` is purple in the demo below:

![screenshot from 2018-09-09 15-56-10](https://user-images.githubusercontent.com/168568/45268352-ed7f7b80-b448-11e8-85e2-50be4e0b1dc0.png)

marijnh:
You could try submitting a pull request to fix this (which I suppose would involve tracking quote depth in the mode state). I don't consider this a big enough problem to work on it myself, so I'm going to close the issue.

dsblank:
Ok, I fixed this (and a couple of other issues) in https://github.com/codemirror/CodeMirror/pull/5585

Before:

![screenshot from 2018-09-20 11-30-21](https://user-images.githubusercontent.com/168568/45829603-5082d500-bcc9-11e8-8905-25db6ee4963b.png)

After:

![screenshot from 2018-09-20 11-29-46](https://user-images.githubusercontent.com/168568/45829619-5d072d80-bcc9-11e8-9f7b-a8ab00addb71.png)

PR #5585, opened 2018-09-20
Scheme mode fixes: properly marks 'atoms and '(lists of s-exp)

dsblank:
Also fixes `#;` comments

Fixes https://github.com/codemirror/CodeMirror/issues/5574

### DIFF
--- a/mode/scheme/scheme.js
+++ b/mode/scheme/scheme.js
@@ -122,7 +122,7 @@ CodeMirror.defineMode("scheme", function () {
                         state.sExprComment = 0;
                     }else{
                         // if not we just comment the entire of the next token
-                        stream.eatWhile(/[^\s]/); // eat non spaces
+                        stream.eatWhile(/[^\s\(\)\[\]]/); // eat symbol atom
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Could you make the indentation match the rest of the file (spaces, not tabs)? And maybe also add `()` to the set of characters that end a commented atom.

dsblank:
Will do!

dsblank:
By "commented atom", do you mean that you would like to properly handle:

```scheme
#; '(this is an sexp marked as an atom) but this is not
```

marijnh:
No, I was thinking more in the direction of `(foo bar #;baz)`, but I've never used `#;` in the wild, so maybe it doesn't work the way I'm assuming it does (which is to comment out the expression after it)?

dsblank:
Ah, I see. I think you are right. I'll see if I can fix this. (I have some other fixes, but I'll save those for another PR).

dsblank:
Ok, it should address those issues.

marijnh:
Thanks! Merged

PR #5586, opened 2018-09-20
R: highlight TRUE/FALSE similar to NA, etc.

haraldschilly:
this is a very minor change, and it just looks better to have `function(x=TRUE)` to highlight this, similar to as it would be `x=NULL`. Also note that you can't assign anything to `TRUE` in R.


marijnh:
Looks good!

PR #5588, opened 2018-09-22
Update colour contrast of commented code

ninjanails:
Update colour contrast of commented code
The current contrast is 3.49:1, which is below the minimum 4.5 ratio needed to meet the WCAG AA guidelines. The proposed change increases the contrast to 4.52:1

Example of how new contract looks
image


marijnh:
I'm sympathetic to increasing the contrast, but I'd like to keep the hue close to what it is now. Could you try to just darken the current color, instead of making it so gray?

ninjanails:
Turns out there was CSS overwriting the brown to make it grey. The original brown passes colour contrast checkers so I will close this pr :)

PR #5600, opened 2018-10-02
package.json/dependency updates

brettz9:
- ~Enhancement: Add ES6 Module distributed file to build process; add minified files with terser and add source map files~
- ~Enhancement: For ESM bundlers, point to new distributed ESM entrance file as `module` in `package.json`~
- Refactoring: ~Switch from Buble to Babel 7 with preset-env (Buble was breaking on array destructuring) and~ update to latest Rollup API
- npm: Update node-static, Rollup deps; add npm-recommended `package-lock.json` ~and `package.json` linting recommended~ meta-data field[s]

Let me know if you need further clarification on any of these. Btw, I put the new ES distribution inside of `dist`. I could put it in `lib` also, but `dist` seemed more clear to me (though if it weren't backward-breaking, I'd think all of `lib` should be moved to `dist`). Let me know if you'd like it elsewhere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 /node_modules
 /npm-debug.log
+/package-lock.json
 /test*.html
 .tern-*
 *~
--- a/package.json
+++ b/package.json
@@ -3,6 +3,11 @@
   "version": "5.40.3",
   "main": "lib/codemirror.js",
   "style": "lib/codemirror.css",
+  "author": {
+    "name": "Marijn Haverbeke",
+    "email": "marijnh@gmail.com",
+    "url": "http://marijnhaverbeke.nl"
+  },
   "description": "Full-featured in-browser code editor",
   "license": "MIT",
   "directories": {
@@ -17,11 +22,11 @@
   },
   "devDependencies": {
     "blint": "^1",
-    "node-static": "0.6.0",
+    "node-static": "0.7.11",
     "phantomjs-prebuilt": "^2.1.12",
-    "rollup": "^0.41.0",
-    "rollup-plugin-buble": "^0.15.0",
-    "rollup-watch": "^3.2.0"
+    "rollup": "^0.66.2",
+    "rollup-plugin-buble": "^0.19.2",
+    "rollup-watch": "^4.3.1"
   },
   "bugs": "http://github.com/codemirror/CodeMirror/issues",
   "keywords": [
@@ -30,13 +35,6 @@
     "Editor"
   ],
   "homepage": "https://codemirror.net",
-  "maintainers": [
-    {
-      "name": "Marijn Haverbeke",
-      "email": "marijnh@gmail.com",
-      "web": "http://marijnhaverbeke.nl"
-    }
-  ],
   "repository": {
     "type": "git",
     "url": "https://github.com/codemirror/CodeMirror.git"
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,7 +1,9 @@
 import buble from 'rollup-plugin-buble';
 
 export default {
-  banner: `// CodeMirror, copyright (c) by Marijn Haverbeke and others
+  input: "src/codemirror.js",
+  output: {
+    banner: `// CodeMirror, copyright (c) by Marijn Haverbeke and others
 // Distributed under an MIT license: https://codemirror.net/LICENSE
 
 // This is CodeMirror (https://codemirror.net), a code editor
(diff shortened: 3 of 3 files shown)


marijnh:
I'm not interested in switching to Babel. Please report the issue you ran into on the Bublé bug tracker and link it here—I'll take a look.

(Also, there's intentionally no package lock in the repository, so please don't add one.)

brettz9:
I've rebased and reverted to Buble; I reported the issue with array destructuring at https://github.com/Rich-Harris/buble/issues/152

I also added `package-lock.json` to `.gitignore` to make clear your desire it not be added (though I personally find it helpful to know everyone is working against the same versions unless they wish to explicitly override).

Btw, are you open to a PR for ESLint and config that matches your current (intentional) conventions?

marijnh:
Okay. But thinking about this a bit more, given that none of the modes or addons are loadable as ES modules, I actually don't think this is a very useful thing to provide.

No, i'm not interested in adding ESLint. I'm also not interested in getting random changes made to package.json (contributors are listed in the AUTHORS file, I'm really not sure why you'd go ahead and just add yourself to package.json when there's no other contributors listed there)..

brettz9:
Ok, re: ESLint and re: `package.json`. I added `contributors` merely to get rid of the linting warnings I have for my `package.json` linter, and if I spent the time searching for who the others were, I would have wasted time because, as confirmed by your answer, I suspected you are pretty particular about how you want things, or rather, how you don't want things. Sometimes one makes a tentative PR and then without wasting a lot of time, get feedback to adjust accordingly; I would obviously add them if you had wanted the field.

As far as the modes/add-ons, would you be interested in my refactoring them as ES6 modules?

Until such time as dynamic import is available (so that one would merely need to pass the mode/add-on names to some CodeMirror method), one could perform static imports which may be imported and then passed the CodeMirror instance, e.g.,

```js
import CodeMirror from '...';
import cssMode from './modes/css/css.js';
cssMode(CodeMirror);
```

marijnh:
> As far as the modes/add-ons, would you be interested in my refactoring them as ES6 modules?

No. Are you aware of the work around [CodeMirror 6](https://codemirror.net/6/)? That's going to be TypeScript and modules-from-the-ground up. I don't think it's worthwhile to make big, potentially disruptive changes to the way the existing version is distributed or maintained at this point. Tooling that integrates CommonJS modules into ES-module oriented bundlers works pretty well, in general.

brettz9:
Okedoke, makes sense... I've amended the commit to minimize the scope... The one `package.json` meta-data field I thought should at least be kept was changing `maintainers` to `author` as it looks like that is how it is supposed to be expressed, with the [docs](https://docs.npmjs.com/files/package.json) merely mentioning that "npm also sets a top-level "maintainers" field with your npm user info."

marijnh:
Right, now it just adds minified files... which is also something I didn't ask for (people tend to have their own build pipelines nowadays, and can do minification on that level). Sorry to be so resistant here, but opening a pull request that makes a number of unprompted changes is usually not a great idea—we have a way in which we've organized our project, and though that may not be the way you'd do it, we have our reasons.

So that just leaves the package.json tweaks and dependency upgrades, I guess. I'd happily merge those if you further reduce this.

brettz9:
No worries... I had a cold today limiting what I could do anyways and just thought I'd do something light with the intent of using it down the road; but glad to hear full module support is apparently all part of the plan even if my help wasn't needed in this form.

I've amended to remove the minified files... Let me know if you would like me to remove the sourcemaps change as well for the non-minified `lib` file. Even for non-minified files, they can be useful for debugging (esp. development work that needs the option of looking back at the specific source files)...

marijnh:
Nope, not a fan of source maps in distributed packages either (blows up the package size, users are only rarely interested in source locations)

brettz9:
Ok, amended... That look all right?

marijnh:
Yes. I'll wait for rollup to release my patch, and then merge this.

adrianheine:
I've published bublé 0.19.4 with your patch.

marijnh:
Great. Merged as 80eeba03634d . The rollup upgrade does seem to give us a bunch of cyclic dependency warnings, but I'm not immediately sure how to address those

adrianheine:
I think that's my fault given that I implemented that warning in rollup and introduced the circular dependencies in CodeMirror :D

marijnh:
Do I understand correctly that the Rollup warning is an approximation, and will fire even if the actual imports involved are not actually immediately accessed at load time (which I entirely understand—tracking what gets evaluated when is a whole different problem)?

adrianheine:
Yeah, it just looks at the import graph, not at usage.

PR #5606, opened 2018-10-06
Add a non-recursive map command to the Vim API

msuozzo:
This will allow for more powerful mapping capabilities like arbitrary command
remapping, rebinding the home row, etc.

This doesn't quite match Vim's non-recursive behavior because there's no support
for non-recursive keyToKey mappings. Without this support, we'd just be creating
normal, recursive keyToKey mappings which I think would be even more frustrating
and/or surprising (e.g. two inverse non-recursive maps would cause infinite recursion).
Once this support is added for this, though, it will be trivial to create these mappings
at the end of the added function and achieve the most accurate behavior.

One potential paper cut with this implementation is that multiple calls to `unmap` may
be required to fully remove the `noremap`ed values. I think this isn't that severe though
because this looks to be the current behavior with all other `map` operations.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -3,7 +3,7 @@
 
 /**
  * Supported keybindings:
- *   Too many to list. Refer to defaultKeyMap below.
+ *   Too many to list. Refer to defaultKeymap below.
  *
  * Supported Ex commands:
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks!

PR #5609, opened 2018-10-08
WIP: Support string templates in Kotlin mode

AlexanderPrendota:
Hello @marijnh !

Kotlin is supported [string templates](https://kotlinlang.org/docs/reference/basic-types.html#string-templates). A template expression starts with a dollar sign ($) and consists of either a simple name:
screen shot 2018-10-08 at 3 18 40 pm

Now in kotlin mode we have plain string. I notice the gradle mode supports string templates.

I tried to use the same logic to kotlin mode.
Unfortunately, I do not know the codemirror API.

I would be happy if you give some advice or help with string templates in kotlin mode.
Probably it's actual in C# mode too.

Thanks!


marijnh:
I'm not willing to walk you through implementing this—but if you read the docs and look at some other modes that have the same issue, I'm sure you can figure it out!

PR #5614, opened 2018-10-09
Ignore lastCopied if not paste.

### LINKED ISSUE #5613: Line-wise cut/copy of the last line can corrupt input.
Steps to reproduce:

1. From https://codemirror.net/, Command-A to select all, and replace the contents with:

```
a
b
```

2. Place the caret on the second line (after the `b`) and Command-C to copy the line.

3. Move the caret up the first line (after the `a`) and type `b`.

Expected result:

```
ab
b
```

Actual result:

```
ba
b
```

It seems the problem is the internal lastCopied state:

https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/src/input/input.js#L53

Specifically, lastCopied.lineWise is true, and the lastCopied.text is equal to the new input; but that’s only because the last copy was both line-wise and the last line, so there’s no trailing newline.

mbostock:
Fixes #5613.

An alternative fix here would be to ensure that lastCopied’s text always includes a trailing newline when it’s line-wise.

### DIFF
--- a/src/input/input.js
+++ b/src/input/input.js
@@ -50,7 +50,7 @@ export function applyTextInput(cm, inserted, deleted, sel, origin) {
         from = Pos(from.line, from.ch - deleted)
       else if (cm.state.overwrite && !paste) // Handle overwrite
         to = Pos(to.line, Math.min(getLine(doc, to.line).text.length, to.ch + lst(textLines).length))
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for the fix! Merged.

PR #5618, opened 2018-10-11
Add FreeMarker HTML Overlay

robotdan:
Initial support for FreeMarker overlay.


adrianheine:
Hi, we're not accepting new modes into the main distribution anymore. The recommended way to publish your mode is to put it in its own repository and publish it on npm---see for example the [elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir).

PR #5619, opened 2018-10-15
Add isRemote to replaceRange to apply remote changes

hackerwins:
First of all, thank you for your wonderful work.

I recently used `cm.replaceRange` to apply remote changes to the local editor while building the group editor with CRDT. `cm.replaceRange` works well in most cases, but when a remote change inserts text at the same location of local cursor, I notice that it moves the cursor like the local action.

http://recordit.co/I63LH3efd9

I'm not sure this is right way, but I added isRemote parameter to replaceRange and processed them differently.

https://github.com/codemirror/CodeMirror/compare/master...hackerwins:replace-range-on-remote-change?expand=1#diff-54bc1a87df59cc68a1005071201142a3R312

Please let me know if there is a better way.
Thank you.

### DIFF
--- a/src/model/Doc.js
+++ b/src/model/Doc.js
@@ -78,10 +78,10 @@ Doc.prototype = createObj(BranchChunk.prototype, {
     if (this.cm) scrollToCoords(this.cm, 0, 0)
     setSelection(this, simpleSelection(top), sel_dontScroll)
   }),
-  replaceRange: function(code, from, to, origin) {
+  replaceRange: function(code, from, to, origin, isRemote) {
     from = clipPos(this, from)
     to = to ? clipPos(this, to) : from
-    replaceRange(this, code, from, to, origin)
+    replaceRange(this, code, from, to, origin, isRemote)
   },
   getRange: function(from, to, lineSep) {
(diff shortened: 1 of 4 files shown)


marijnh:
That's a reasonable use case, and the new API in version 6 will handle this properly by default (it treats replacing the selection as a special case, and does what you need here by default). But I'm not sure this needs to be added to the core library. Have you considered using a kludge like this, in your code, instead?

```javascript
function addChange(editor, from, to, text) {
  let adjust = editor.listSelections().findIndex(({anchor, head}) => {
    return CodeMirror.cmpPos(anchor, head) == 0 && CodeMirror.cmpPos(anchor, from) == 0
  })
  editor.operation(() => {
    editor.replaceRange(text, from, to)
    if (adjust > -1) {
      let range = editor.listSelections()[adjust]
      if (range && CodeMirror.cmpPos(range.head, CodeMirror.changeEnd({from, to, text})) == 0) {
        let ranges = editor.listSelections().slice()
        ranges[adjust] = {anchor: from, head: from}
        editor.setSelections(ranges)
      }
    }
  })
}
```

hackerwins:
@marijnh Thank you for your kind feedback. I will try it as you suggest. 👍

PR #5628, opened 2018-10-18
Change Google Inc. to Google LLC

dmaclach:
(and added myself)

The main change here is to make sure that Google Inc. gets changed over to Google LLC (our proper name now).


adrianheine:
Make sure to fix `bin/authors.sh`, too.

dmaclach:
Done.

PR #5629, opened 2018-10-19
[clike mode] Update default cTypes

dmaclach:
Updates the cTypes list to cover newer types (c11) and by default picks up anything ending in '_t' as a type as defined by the POSIX standard (and generally adopted by C compiler types such as int16_t int32_t wchar_t etc).


dmaclach:
Added fix for breakage.  Can someone trigger a rebuild to verify?

marijnh:
Travis automatically re-tests when you change the branch. The problem appears to be that you're using `includes`, which isn't available on all the targets that we support.

dmaclach:
Ah.. Travis turned the code around so fast that I thought it hadn't run. Thanks!

dmaclach:
For some reason this PR got messed up (yay git!...) I'll open a new PR with updated changes.

PR #5634, opened 2018-10-20
[mode clike] Turn on fold indicators in demo/test.

dmaclach:
Since indent handling is part of mode, it's nice to have the indent handling turned on in the demo file so that you can easily test it.

### DIFF
--- a/mode/clike/index.html
+++ b/mode/clike/index.html
@@ -5,8 +5,13 @@
 <link rel=stylesheet href="../../doc/docs.css">
 
 <link rel="stylesheet" href="../../lib/codemirror.css">
+<link rel="stylesheet" href="../../addon/fold/foldgutter.css">
 <script src="../../lib/codemirror.js"></script>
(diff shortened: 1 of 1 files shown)


marijnh:
You seem to mean 'fold' when you say 'indent'. But fold handling isn't part of this mode's code at all, so I'm not really sure what purpose this would serve.

dmaclach:
Arrggh.. you are correct I meant fold instead of indent. Thank you. I was referring to this:

https://github.com/codemirror/CodeMirror/blob/41e7ff5d04825c94703b2153aa72466546f2d5ab/mode/clike/clike.js#L725

and was using it when testing my folding changes for disabling single line folds and seeing how it looked in the various languages.

marijnh:
Right, it does declare a fold strategy, but I don't want to complicate the demo just to make testing that one thing easier, so I'm going to reject this pr.

dmaclach:
No worries. Totally understand. Thanks!

PR #5637, opened 2018-10-22
[clike mode] Update default cTypes

dmaclach:
Updates the cTypes list to cover newer types (c11) and by
default picks up anything ending in '_t' as a type as defined
by the POSIX standard (and generally adopted by C compiler types
such as int16_t int32_t wchar_t etc).


dmaclach:
Was #5634. I screwed up git somehow in a way that github didn't like. I closed #5634 and have restarted here. Apologies.

marijnh:
Much better. Unfortunately, we also target IE8, which doesn't have `Array.indexOf`. So maybe use `words` to turn you sets into objects?

dmaclach:
Done.

marijnh:
Thanks!

PR #5643, opened 2018-10-25
[kotlin mode] update indents

AlexanderPrendota:
Fix indent after `operator` on top level.
**Example :**
After fix:
```kotlin
import kotlinx.coroutines.*
import kotlinx.coroutines.*
```
Should have the same indents.

before: 
```kotlin
import kotlinx.coroutines.*
    import kotlinx.coroutines.*
````

### CODE REVIEW

**Comment by adrianheine** [comment]
I think the issue here is that the `*` at the end is treated as an operator. For multi-line expressions like 
```kotlin
(someValue + x) *
  aBigCoefficient
```
it makes sense to indent the second line.

marijnh:
Right, that's what the patch tries to address, but using a factor that's not really directly related to the problem (whether we are in a braced block). Is that solid for Kotlin, or will this break some versions of indent-after-operator in other context?

AlexanderPrendota:
The case 

```kotlin
val a = (2 + 2) * 
     6
```
works fine.

But the operator `*` in the top level should have the 0 indent.

But also we have the case like
```kotlin
val x: String? = y as String?
```
the line after operator `?` should have the 0 indent in block.
Now It works wrong -> 
```kotlin
val x: String? = y as String?
    val a= "some"
```
But this PR only fix the indent on top level.

marijnh:
One approach might be to tokenize binary operators different from other operators or other uses of `*`, so that the indentation rely on a binary operator being a binary operator. But that means the Kotlin mode would have to also add some operator-parsing logic in a hook.

AlexanderPrendota:
I added some logic in a hook. Please take a look please. Thanks!

mode/clike/clike.js
```diff
@@ -667,6 +667,9 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
         stream.eatWhile(/[\w\$_]/);
         return "meta";
       },
+      '*': function(stream) {
+        return stream.string.match(/^import.*/) ? 'variable' : 'operator';
```
marijnh: Maybe check whether `prevToken` is `"."` instead? This regexp would go wrong on a line like `import foo.*; return x *` (which I'm not even sure is valid Kotlin, but you get the point.)

AlexanderPrendota:
@marijnh updated

PR #5661, opened 2018-11-12
Permit quoted names for object destructuring in JavaScript syntax

### LINKED ISSUE #5660: Renaming, destructuring, quoted properties in JavaScript aren't properly syntax highlighted
It's pretty picky but this feature _is_ supported by Chrome, Node, etc, so - for example:

```js
function a({ ['hello world']: helloWorld }) {
   return helloWorld;
}
```

Currently CodeMirror doesn't expect bracketed & quoted expressions in destructuring syntax, so `helloWorld` is never registered and thus isn't properly highlighted in the function body.

PR to come shortly after...

tmcw:
Fixes #5660: adds support for this case:

```js
function a({ ['hello world']: helloWorld }) {
   return helloWorld;
}
```


adrianheine:
Thanks!

PR #5663, opened 2018-11-13
Support cross-window hint rendering

markdalgleish:
We're working on a new feature for [Playroom](https://github.com/seek-oss/playroom) where you can [detach the code editor from the main window](https://github.com/seek-oss/playroom/pull/7) and open it in a new window. This has highlighted an issue in the `show-hint` addon.

Currently, `show-hint` references the globals `window` and `document` to make layout calculations and add elements to the page. In our case, the code is executing in the parent window, so these global references are actually pointing at the wrong window. When writing code in the popup window, the suggestions are instead rendered to the parent window ([video link](https://cl.ly/92433d9175fd)).

To fix this, we ensure that references to `document` and `window` are scoped to the window that contains the textarea. We achieve this by calling `cm.getInputField().ownerDocument`, which is [supported across all browsers](https://caniuse.com/#search=ownerdocument). We then get a reference to the parent window via `ownerDocument.defaultView || ownerDocument.parentWindow`, which ensures backwards compatibility with older versions of IE since it doesn't support [defaultView](https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView), instead providing its own proprietary [parentWindow](https://msdn.microsoft.com/en-us/windows/ms534331(v=vs.71)) property.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -198,20 +198,22 @@
     this.data = data;
     this.picked = false;
     var widget = this, cm = completion.cm;
+    var ownerDocument = cm.getInputField().ownerDocument;
+    var parentWindow = ownerDocument.defaultView || ownerDocument.parentWindow;
 
-    var hints = this.hints = document.createElement("ul");
+    var hints = this.hints = ownerDocument.createElement("ul");
     var theme = completion.cm.options.theme;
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for fixing that! Merged.

PR #5676, opened 2018-11-20
Changed styling for bracket matching in Base 16 Light

detly:
The current bracket matching highlight for Base 16 Light is pure white, which (a) isn't part of the theme and (b) is extremely difficult to see against the light background. I changed it to have a themed colour, a blue backround (like a block highlight)  and remove the underline (redundant with the other change).

Before:

bracket2

After:

bracket3

### DIFF
--- a/theme/base16-light.css
+++ b/theme/base16-light.css
@@ -35,4 +35,4 @@
 .cm-s-base16-light span.cm-error { background: #ac4142; color: #505050; }
 
 .cm-s-base16-light .CodeMirror-activeline-background { background: #DDDCDC; }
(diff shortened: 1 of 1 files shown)


marijnh:
Are the `!mportant` annotations necessary here?

detly:
> Are the `!mportant` annotations necessary here?

Not sure, the original CSS had it so I left it in. My guess (and it's only a guess) would be that it means it avoids a dependency on the order in which the final CSS is constructed, but maybe that's taken care of by using more specific selectors?

marijnh:
Oh, hm, yes, the rule for the default matching brackets is oddly specific for some reason, so to override it you have to either be more specific or add `!important`. So sure, let's just leave those there, since I'm not sure what would break if I fixed that.

PR #5683, opened 2018-11-29
Linting for htmlmixed

LaKing:
Updated the linter demo page, the html code had some syntax errors.

Updated links to unpkg links. Added prefix for https protocols, so that people that clone the repo can access it instantly. Problem is that if no protocol is specified file protocol is assumed in a local clone.

I created a separate js file for the htmlmixed mode linter, so we have an independent file not modifying anything else. Mostly combined the existing code into 3 functions that take a look at the input text. Regexp based separators separate it into code blocks and apply the linters. The base syntax is html, so for scripts and styles we assume the odd blocks are scripts and style blocks and run the linters in there, therefore lines in the message are not calculated with offsets. Offsets are calculated via line breaks.
Had to clone local options objects as the external linters modified the options object, and the linters complained at each other's runtime-added options. Added some extra properties as default so that there is a way for language-specific options: options.css, options.js and options.html are taken first if exists, if not, we take the clones.

Let me know if there is anything that should be done differently.

### DIFF
--- a/addon/lint/htmlmixed-lint.js
+++ b/addon/lint/htmlmixed-lint.js
@@ -1,0 +1,167 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others,
+// Initial htmlmixed-lint.js from István Király, https://github.com/LaKing
+// Distributed under an MIT license: https://codemirror.net/LICENSE
+
+// Depends on htmlhint jshint and csshint
+
+(function(mod) {
+    if (typeof exports == "object" && typeof module == "object") // CommonJS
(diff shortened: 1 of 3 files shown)


marijnh:
> Updated links to unpkg links.

Thanks. I've merged that as 5740f791

> Updated the linter demo page, the html code had some syntax errors.

HTML5 allows missing quotes around single-word attributes and such, there's no need to 'fix' those.

> I created a separate js file for the htmlmixed mode linter,

I'm not accepting new addons into the main distribution—the recommended way to distribute new functionality is to create separate repositories and npm packages for them.

PR #5688, opened 2018-11-30
Allow vim-mode to select text enclosed in angled brackets.

yoongu:
CodeMirror supports vim text objects [1] enclosed in parentheses, curly braces, square brackets, but **not** angled brackets. This is a known bug, and a previously documented TODO item in vim.js.

[1] http://vimdoc.sourceforge.net/htmldoc/motion.html#object-select


marijnh:
Thanks!

PR #5699, opened 2018-12-10
soy mode - classify param def type as type token

notoriousb1t:
- changed classification of param-def-type to "type" instead of using "variable-3"
- updated tests to match this expectation

This will provide more accurate syntax highlighting and make the parameter types standout better for theming.


marijnh:
Ah, yes, I missed this mode when introducing the `type` token type. Thanks for fixing that!

PR #5701, opened 2018-12-11
SCSS file could spit out CSS file

mcculloughjchris:
This might be a weird pull request, but a project I'm working on is giving me warnings because I'm running import on css files from within a scss file. Implement this and I could be importing the scss file instead, getting rid of some of the warnings. I haven't put it in yet, but colors, margins, etc. could be variables, I'm sure that would make some people's lives easier. I figured this would be a simple enough thing to add in? If not, no worries.

Edit:
Only issue left is all the theme and addon css files, but I don't want to trash up the repo by doubling the number of style-related files. Any suggestions?

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -1,346 +1,380 @@
 /* BASICS */
-
 .CodeMirror {
   /* Set height, width, borders, and global font properties here */
   font-family: monospace;
   height: 300px;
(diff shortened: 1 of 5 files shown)


marijnh:
Sorry, no, we're not changing our tooling to please other people's build process.

PR #5709, opened 2018-12-14
Updating MIME type of Jinja2

mwadman:
Hi CodeMirror team,

In this pull request I'm setting the mime type for the Jinja2 language to "plain/text".

The setting of this to "null" is causing issues with projects that reference/use CodeMirror.
For example, see a [pull request I've submitted on the Github/linguist](https://github.com/github/linguist/pull/4349) project.

Thanks for look at this PR. Let me know what you think.

### DIFF
--- a/mode/jinja2/jinja2.js
+++ b/mode/jinja2/jinja2.js
@@ -141,4 +141,6 @@
       blockCommentEnd: "#}"
     };
   });
+
+  CodeMirror.defineMIME("text/plain", "jinja2");
 });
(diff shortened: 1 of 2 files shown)


marijnh:
MIME types are used to find modes, and `text/plain` should not be mapped to the jinja2 mode, so this is not a good idea. You could find or make up another mime type that actually identifies the mode, but you can't use text/plain.

mwadman:
Thanks for taking a look at the PR, much appreciated.

Would you be okay with a submission of "text/jinja2" as the mime?

marijnh:
Definitely!

mwadman:
Hi, just coming back to this after the holiday break. (Happy new year!)
Would you prefer that I submit a new PR, or re-open this one?

marijnh:
Either is fine, doesn't really matter.

PR #5711, opened 2018-12-15
Fix/keymap/vim ctrl w idle

fraxx001:
Add new keymap for ctrl-w in normal mode.
type idle is not catched by Command dispatcher and therefore idles.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -152,6 +152,8 @@
     { keys: '~', type: 'operatorMotion', operator: 'changeCase', motion: 'moveByCharacters', motionArgs: { forward: true }, operatorArgs: { shouldMoveCursor: true }, context: 'normal'},
     { keys: '~', type: 'operator', operator: 'changeCase', context: 'visual'},
     { keys: '<C-w>', type: 'operatorMotion', operator: 'delete', motion: 'moveByWords', motionArgs: { forward: false, wordEnd: false }, context: 'insert' },
+    //ignore C-w in normal mode
+    { keys: '<C-w>', type: 'idle', context: 'normal' },
     // Actions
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks!

pmbauer:
### Questions
Is there context on why `<C-w>` was explicitly disabled in normal mode?  I'd work on a pull request to re-enable it (or submit a revert) but I want to understand the use case for the pull @marijnh @fraxx001.

1. Would it make sense to revert this change or would that break another use case?
2. If it would break another use case, would you accept a patch to over-ride entries in the default keymap for contexts it makes sense?

### Context
Because `<C-w>` is explicitly disallowed in the default keymap, [it is not possible to use window manipulation routines](https://github.com/esm7/obsidian-vimrc-support/issues/67) like split, vsplit, and pane navigation in a context that makes sense.

For context, I'm using vim mode in Obsidian with the `obsidian-vimrc-support` plugin, and would like to make the following work:

```
exmap focusRight obcommand editor:focus-right
nmap <C-w>l :focusRight

exmap focusLeft obcommand editor:focus-left
nmap <C-w>h :focusLeft

exmap focusTop obcommand editor:focus-top
nmap <C-w>k :focusTop

exmap focusBottom obcommand editor:focus-bottom
nmap <C-w>j :focusBottom

exmap splitVertical obcommand workspace:split-vertical
nmap <C-w>v :splitVertical

exmap splitHorizontal obcommand workspace:split-horizontal
nmap <C-w>s :splitHorizontal
```

I have an [hilarious hack to patch the `vim.js` in Obsidian](https://github.com/esm7/obsidian-vimrc-support/issues/67#issuecomment-1063734459), that does let me use the `<C-w>` leader to manipulate windows splits and navigate windows.

PR #5712, opened 2018-12-17
Add SUBSCRIPTION

sql-migrate:
See https://www.postgresql.org/docs/10/sql-createsubscription.html


marijnh:
Thanks, but could you maybe find a place where all postges keywords are enumerated and add them all at the same time, rather than one by one in separate pull requests?

marijnh:
Guess not. Merged anyway!

PR #5715, opened 2018-12-19
[soy mode] boolean and number tokenization in expressions

notoriousb1t:
Highlights { true } or { false } as atoms.  Also highlights all number literals.

- added tokenization for decimals
- added tokenization for hexadecimals
- added tokenization for true/false
- added test cases for those new atom tokens

### DIFF
--- a/mode/soy/soy.js
+++ b/mode/soy/soy.js
@@ -266,6 +266,17 @@
               state.quoteKind = match;
               return "string";
             }
+            if (match = stream.match(/(true|false)(?!\w)/)) {
+              return "atom";
+            }
+            if (match = stream.match(/0x([0-9a-fA-F]{2,})/)) {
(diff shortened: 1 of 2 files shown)


mode/soy/soy.js
```diff
@@ -266,6 +266,17 @@
...
+              return "atom";
+            }
+            if (match = stream.match(/0x([0-9a-fA-F]{2,})/)) {
+              // Match hexadecimal number.
+              return "atom";
+            }
+            if (match = stream.match(/-?([0-9]*[.])?[0-9]+/)) {
+              // Match all decimals and whole numbers.
+              return "atom";
+            }
```
marijnh: You can put all of these in a single `if` with `||`, and you don't need to assign to `match` here since you aren't going to read from it. Could you make those changes?

PR #5717, opened 2018-12-20
[soy mode] tokenize operators in expressions

notoriousb1t:
- tokenize arithmetic and equality operators as "operator"
- added tests to validate this

### DIFF
--- a/mode/soy/soy.js
+++ b/mode/soy/soy.js
@@ -271,6 +271,10 @@
               stream.match(/-?([0-9]*[.])?[0-9]+/)) {
               return "atom";
             }
+            if (stream.match(/(\||[+\-*\/%]|[=!][=]|[<>][=]?)/)) {
+              // Tokenize filter, binary, and equality operators.
+              return "operator";
+            }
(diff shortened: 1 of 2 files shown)


mode/soy/soy.js
```diff
@@ -271,6 +271,10 @@
               stream.match(/-?([0-9]*[.])?[0-9]+/)) {
               return "atom";
             }
+            if (stream.match(/(\||[+\-*\/%]|[=!][=]|[<>][=]?)/)) {
+              // Tokenize filter, binary, and equality operators.
+              return "operator";
+            }
```
marijnh: It's not a big deal here, but I also noticed it in a previous patch, so for next time—you don't have to put single characters in square brackets in regular expressions. I.e. `[=]` is equivalent to `=`.

notoriousb1t: You are right.  I am not sure why I used [=] and to be honest, I figured out some of these expressions 2 months ago so I am not sure what I was thinking at the time.

PR #5721, opened 2018-12-25
show-hint: bind "Ctrl-P" to "Up" and "Ctrl-N" to "Down" by default

gfx:
I think these are de-facto standard keybindings in completion panels.


marijnh:
Only on macOS, right? Could you make it so that these are only added there?

gfx:
Right. How about e9d3f1e3 (as long as I tried, browser.js seems not to be available for addons)?

PR #5725, opened 2018-12-28
[css mode] more consistent css function tokens

notoriousb1t:
- created a css function regexp with a positive lookahead for (. There
are still a few cases with IE functions that have . and : in it that
this does not catch.
- folded url/domain/regexp into that css function regexp including the
special string tokenization that was in place already
- updated css, less, and scss tests to reflect this new tokenization


marijnh:
Thanks!

PR #5728, opened 2018-12-28
Support ruby indented heredoc

### LINKED ISSUE #5727: [Ruby] Indented ending identifier in heredoc not supported
Highlighting is broken with indented ending identifier. It [should be supported](https://ruby-doc.org/core-2.2.0/doc/syntax/literals_rdoc.html#label-Here+Documents) if a '-' is placed after '<<'.

ruby_indented_ending_identifier

Hugo-Hache:
This PR fixes #5727

### DIFF
--- a/mode/ruby/ruby.js
+++ b/mode/ruby/ruby.js
@@ -63,8 +63,8 @@ CodeMirror.defineMode("ruby", function(config) {
     } else if (ch == "#") {
       stream.skipToEnd();
       return "comment";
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for the PR! I ended up fixing it in a simpler way in be9d254e

Hugo-Hache:
@marijnh Indeed much simpler than the regex 👏 However based on the specification we should only allow eating space when opening tag is prefixed with '-'. Would it be ok for you if I remove the regexp part from my PR and keeps the '-' detection ?

Hugo-Hache:
Perfect! Thanks for you reactivity @marijnh

PR #5739, opened 2019-01-08
Add options for autocorrect, autocapitalize

joewalsh:
Would like the ability to enable autocorrect and autocapitalize in addition to spellcheck. Verified that the same tests pass that pass on master.


marijnh:
Could you make it so that these can be modified at run time (which is what the callback passed to `option` for `spellcheck` is for), and add them to the manual?

joewalsh:
@marijnh thanks for the quick response! I've updated the PR with the changes you requested

marijnh:
Thanks!

Issue #3403, opened 2015-07-21
Capslock on iOS


D-32:
When focusing onto an empty line on an iOS device inside WKWebview or Safari the keyboard switches to all caps mode.
This can be reproduced easily in the example on http://codemirror.net. Just tap somewhere next to the last line, the cursor will focus as seen on the attached screenshot and the keyboard goes into all caps mode.

![img_0048](https://cloud.githubusercontent.com/assets/3063279/8808181/dedf8a60-2f95-11e5-8616-2c3e8e1ff8d0.PNG)

marijnh:
It goes into 'capitalize the next letter' mode, which is what mobile keyboards tend to do when at the start of what looks like a sentence. I don't think there is a way for CodeMirror to suppress this, but if you do know of one, tell me.

D-32:
That would make sense, but it stays in caps lock. So if I type A the next character is still upper case. Which is annoying, as the user manually has to get out of that mode.
This surely isn't an intended behavior.

marijnh:
That is indeed odd. I'll see if I can figure out what is happening the next time I have my iPad in my hand.

D-32:
Thanks :)

boucher:
Seems related to this bug filed against WebKit:
https://bugs.webkit.org/show_bug.cgi?id=148503

boucher:
Switching to inputStyle: "textarea" resolves this issue, but produces the issue where the native cursor is displayed in addition to the codemirror cursor. I might be willing to live with that by disabling the codemirror cursor, but perhaps there are other problems I haven't discovered yet since textarea is not the default on mobile?

denis-aes:
I've met same issue. The problem is that if e.preventDefault is called - then iOS will skip internal callbacks and will NOT switch case to lower.

For CM4 I've made a workaround for CM keypress event like this
    onKeyPress: function(e) {
      if (capitalizationHack.isPreventDefaultNeeded(e)){ // if uppercase -> not needed
        e.preventDefault();
      }
      operation(....)
    },

And after this - we have two events of adding letter. The extra letter should be removed

  if (cm.capitalizationHack.isHackEnabled()) {
    CodeMirror.on(cm.doc, 'change', function (instance, change) {
      var capHack = this.editor.capitalizationHack;
        if (capHack.isDuplicateChange() && (change.text.length == 1)) {
          console.log("Line BF: " + cm.getLine(change.from.line));

          var realFrom = {
            line: change.from.line,
[...]

PR #5750, opened 2019-01-17
Add option to disable preventDefault on key press for contentEditable input views

joewalsh:
Adds a setting to work around the iOS issue detailed in #3403.

This doesn't seem to break anything in our use case, but let me know if there are other unintended consequences to this setting & change.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -566,6 +566,9 @@
       
       <dt id="option_autocapitalize"><code><strong>autocapitalize</strong>: boolean</code></dt>
       <dd>Specifies whether or not autocapitalization will be enabled on the input.</dd>
+      
+      <dt id="option_preventDefaultOnKeyPress"><code><strong>preventDefaultOnKeyPress</strong>: boolean</code></dt>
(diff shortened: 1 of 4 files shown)


marijnh:
Thanks for looking into this, but it seems very scary—just turning off the applying of the actual change event in `applyTextInput` definitely cause things to get confused. Would it be possible just let it apply its input, and then making sure that the DOM change caused by the key press is somehow handled so that it doesn't cause the same change twice?

joewalsh:
@marijnh thanks for the quick feedback! The initial change I made cut out `operation(this.cm, applyTextInput)` entirely. I thought maybe that was too invasive, but there weren't any observable ill effects in our use case (syntax highlighting and undo continue to work as normal). I've reverted to that version of the change. 

I'm not sure what you mean by `making sure that the DOM change caused by the key press is somehow handled` - would calling `readFromDOMSoon()` on keypress when preventDefault is disabled suffice or is there something else I'm missing?

marijnh:
It may just work already, but we have to make sure that the DOM update happens before the editor tries to read from the DOM and update the document twice for the same change.

PR #5751, opened 2019-01-18
Fix middle-click paste on Firefox.

nchevobbe:
The current solution consists in focusing the
input when the scroll container capture a paste
event, which does not work on Firefox.

In order to fix this, creating a new paste event,
copying the clipboardData and dispatching it
on the textarea seems to do the trick.

This works on Firefox 64, with the
config set to true.

### DIFF
--- a/src/input/TextareaInput.js
+++ b/src/input/TextareaInput.js
@@ -76,8 +76,16 @@ export default class TextareaInput {
 
     on(display.scroller, "paste", e => {
       if (eventInWidget(display, e) || signalDOMEvent(cm, e)) return
-      cm.state.pasteIncoming = true
-      input.focus()
+      if (!te.dispatchEvent) {
+        cm.state.pasteIncoming = true
+        input.focus()
+        return
+      }
+
+      // Pass the `paste` event to the textarea so it's handled by its event listener.
+      const event = new Event("paste")
+      event.clipboardData = e.clipboardData
+      te.dispatchEvent(event)
     })
 
     // Prevent normal selection in the editor (we handle our own)


nchevobbe:
I'm only able to test this on OSX with Firefox (using the MiddleClick software to have a true middle-click). I did not find a way to enable middle-click paste on Chrone on OSX.

I would be interested to know if everything is okay in other browser and platforms.

marijnh:
Did you test whether this breaks middle-click paste on Chrome? There `dispatchEvent` will also be present, so if it works as-is on Chrome, it looks like the fallback code isn't going to be used anywhere.

marijnh:
Unfortunately, this works neither in Chrome nor on Firefox for me (using Linux)

nchevobbe:
> Did you test whether this breaks middle-click paste on Chrome? There dispatchEvent will also be present, so if it works as-is on Chrome, it looks like the fallback code isn't going to be used anywhere.

Yeah, I was just being cautious for old-browser support.

nchevobbe:
> Unfortunately, this works neither in Chrome nor on Firefox for me (using Linux)

:/ , I'm positive it should work on Firefox (maybe only Nightly for Linux, but I doubt it). Okay, time to  boot-up my VM.

nchevobbe:
Hello Martin,

I put up a page to test this PR : 

It's working for me on: 
- Firefox 64+ on OSX
- Firefox 67 on Ubuntu 17.10 (64-bit)
- Chromium 67.0.3396.99 on Ubuntu 17.10 (64-bit)

I'm currently upgrading my Ubuntu VM in order to test Firefox 64 on it.

Could you tell me if: 
- it still does not work for you with my demo page
- what steps you do to try to paste (I tested the patch by middle-clicking on the current cursor line, on the cursor itself, and also on part of the screen which is only covered by the scroll container, but maybe I'm missing something).

Thanks!

nchevobbe:
Verified it also works on:
- Chromium 71.0.3578.98, Ubuntu 18.04 (64-bit)
- Firefox 64, Ubuntu 18.04 (64-bit)

marijnh:
I can't see the page you put up (there's no link), but retrying the patch, I can see Firefox work now. Chrome (71.0.3578.98) still doesn't do anything when I middle click on an editor.

nchevobbe:
Sorry, here it is https://codemirror-firefox-middle-click.glitch.me/

marijnh:
Same problem. Though I did notice that middle-click paste on Chrome works when the editor is _unfocused_, but not when it has focus.

nchevobbe:
> Chrome (71.0.3578.98) still doesn't do anything when I middle click on an editor.

worst case, I could use the `gecko` floag from https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/src/util/browser.js#L6, but i'd like not to.

> Though I did notice that middle-click paste on Chrome works when the editor is unfocused, but not when it has focus.

That's a lead, I'll check this.

nchevobbe:
One silly question: does it works for you on Chrome without this patch?

marijnh:
Not a silly question at all, it seems! It doesn't, and in fact old versions also no longer work, so I guess something in Chrome changed that broke the existing kludge. Interesting that no one seems to have complained about it yet.

Given that, I guess it'd be a good idea to drop the focus kludge code entirely, since it won't be firing or working anywhere.

nchevobbe:
Could this be somehow OS specific? What distrib+version are you testing on? (As stated before, it does work for me on Chromium, on Ubuntu, both with and without my patch)

marijnh:
I guess it's more likely to be a strange local issue, at this point. I'm spinning up another Debian machine to see if it's better there.

marijnh:
Hm, nope, same problem in Chrome 69 there.

marijnh:
(And by 'same issue' I mean that it doesn't work at all before your patch, and only works when unfocused with the patch, so the patch still is an improvement.)

nchevobbe:
Okay, so on Debian it's working for Firefox, and for Chromium, but only if unfocused (same as you basically). I can see the textarea `focus` event listener being fired, but I don't know if it may cause the issue.
Should we proceed with this patch and create an issue for Debian+Chrome ?

marijnh:
Yes, I think that's reasonable. Thanks for figuring out this workaround for Firefox!

nchevobbe:
Thanks for helping me figuring this out @marijnh :) 
Also, thanks for maintaining this repo (and for the new version, I'm excited to try this one out :) )

dmitry-at-publons:
I've just tried https://codemirror-firefox-middle-click.glitch.me/:

**Firefox 64.0** at Ubuntu 18.04: pasting works, but **copying** text from a form (selecting with a mouse or double-clicking) does not.

**Chromium 71** at Ubuntu 18.04: everything works

Issue #5757, opened 2019-01-23
newlineAndIndentContinueMarkdownList isn't compatible with multiplex mode


thomasjm:
I'm trying to use `mode/multiplex.js` with Markdown as the outer mode and various other languages as the inner modes. I found that the `newlineAndIndentContinueMarkdownList` function is not multiplex mode aware; see here for how it tries to pull the current state for a given line:

https://github.com/codemirror/CodeMirror/blob/2dab4d527c7a55aabc0a52b28b47ab35b1412ca3/addon/edit/continuelist.js#L23

It assumes that the state is a Markdown mode state on the next line.

I think it would be better if it could detect the presence of multiplex mode and pull out the active mode, by adding something like this right after line 23:

```
// If we're in multiplexing mode, pull out the currently active state.
if (eolState.outer) {
    eolState = eolState.inner || eolState.outer;
}
```

Happy to make a PR if the maintainers agree but wanted to see what you think first.

(Is it possible to make this less fragile by doing a check on the state to see which mode it came from?)

marijnh:
Yes, it would make sense to fix that. You'll want to use `CodeMirror.innerMode` to get at the current mode.

thomasjm:
Fixed by #5761

PR #5761, opened 2019-01-25
Make newlineAndIndentContinueMarkdownList understand inner modes

thomasjm:
Fix for issue #5757. It works for my purposes but I haven't tested it super thoroughly, is there a test suite where it should go?

### DIFF
--- a/addon/edit/continuelist.js
+++ b/addon/edit/continuelist.js
@@ -21,20 +21,14 @@
     for (var i = 0; i < ranges.length; i++) {
       var pos = ranges[i].head;
 
-      // Walk the list of inner modes/states until we find a Markdown one
-      // If we don't find one, fall back to normal newlineAndIndent
-      var mode = cm.getMode(), eolState = cm.getStateAfter(pos.line);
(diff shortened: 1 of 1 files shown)


addon/edit/continuelist.js
```diff
@@ -20,7 +20,23 @@
     var ranges = cm.listSelections(), replacements = [];
     for (var i = 0; i < ranges.length; i++) {
       var pos = ranges[i].head;
-      var eolState = cm.getStateAfter(pos.line);
+
+      // Walk the list of inner modes/states until we find a Markdown one
+      // If we don't find one, fall back to normal newlineAndIndent
+      var mode = cm.getMode(), eolState = cm.getStateAfter(pos.line);
+      while (mode && mode.name !== "markdown") {
```
marijnh: You don't need the loop here—`CodeMirror.innerMode` is a helper that'll go down to the innermost mode for you. Then you can check whether it's markdown.

marijnh:
Thanks, that's that I meant. Merged!

PR #5762, opened 2019-01-31
[julia mode] Fix some indentation issues

### LINKED ISSUE #5702: Julia indentation broken after "else" and "elseif"
Indentation gets offset every time an "else" or "elseif" command is used in the Julia language. And it stays offset for the rest of the file. Here is the indentation I get in CodeMirror:
```julia
if x == 1
    println("one")
elseif x == 2
        println("two")
    else
            println("few")
        end
        println("all messed up now")
```
I would expect it to look like this:
```julia
if x == 1
    println("one")
elseif x == 2
    println("two")
else
    println("few")
end
println("all messed up now")
```

pabloferz:
Fixes #5702


marijnh:
(You have some linter errors due to using undefined variables.)

pabloferz:
> (You have some linter errors due to using undefined variables.)

Fixed

marijnh:
Thanks!

Issue #5763, opened 2019-02-01
API provides no way to insert text around inclusiveLeft/inclusiveRight markers


thomasjm:
This sandbox should show most clearly what I'm trying to do: https://codepen.io/anon/pen/bzWpGR

This is partly a question and partly a feature request, if it's not possible. I'm wondering if there's some way to programmatically insert text to the left of an `inclusiveLeft` marker (or right of an `inclusiveRight`) marker without the text being eaten by the marker, and preferably without clearing the marker.

(The motivation for this is to make a collaborative text editor. Suppose user 1 has a collapsed widget on a certain line, and user 2 in another browser inserts a newline on the previous line. When we apply this change to user 1's document we want to do so without the newline being eaten by the marker, but there doesn't appear to be an easy way to do this.)

thomasjm:
Possible solution: reach into the marker and temporarily change its `inclusiveLeft`/`inclusiveRight` setting. How bad an idea is this? https://codepen.io/anon/pen/VgbgKY

marijnh:
Markers can't be changed during their lifetime—that hack might work, but it's not something the library expects, so it's likely to break in subtle ways. The only way to do this would be to remove the mark, make the changes, and recreate it in the proper new position.

More generally, if you don't want inclusive behavior, maybe the mark shouldn't have those options enabled?

thomasjm:
You're right that I don't want inclusive behavior. What I want is for the line to be collapsed, so it's impossible to put the cursor on the line with the marker. But as the manual says, "In atomic ranges, `inclusiveLeft` and `inclusiveRight` have a different meaning—they will prevent the cursor from being placed respectively directly before and directly after the range." So I'm forced to use the inclusive options to get the effect I want. Perhaps the real source of the trouble is how the meaning of `inclusiveLeft`/`inclusiveRight` is overloaded this way.

Maybe it would be good to separate the "no cursoring before or after" meaning into a separate option called `preventCursorLeft`/`preventCursorRight` (which has no effect on non-atomic ranges)?

thomasjm:
This was closed by #5773

PR #5773, opened 2019-02-06
Add preventCursorLeft/preventCursorRight options for atomic markers

thomasjm:
Per issue #5763. Hopefully it's this simple, seems to work for me. 

Would probably be good to add tests that the precedence works right when both options are provided. I'm having trouble running the tests locally though, `npm run test` produces errors like `ReferenceError: Can't find variable: addDoc (localhost:3000)`.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -1674,13 +1674,21 @@ editor.setOption("extraKeys", {
         <dt id="mark_inclusiveRight"><code><strong>inclusiveRight</strong>: boolean</code></dt>
         <dd>Like <code>inclusiveLeft</code>,
         but for the right side.</dd>
+        <dt id="mark_selectLeft"><code><strong>selectLeft</strong>: boolean</code></dt>
+        <dd>For atomic ranges, determines whether the cursor is allowed
+        to be placed directly to the left of the range. Has no effect on
+        non-atomic ranges.</dd>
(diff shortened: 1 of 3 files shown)


marijnh:
Some of the errors in the travis run seem spurious (I can't see why they'd occur, and they don't occur for me locally), but the `core_atomicMarker` test is definitely failing for me (both in Firefox and Chrome).

Also, I prefer to avoid negatives in option names (because you get a double negative when you set them to false), and this applies not just to cursors but to any selection endpoints, so I'd prefer to rename these to `selectLeft` and `selectRight`, and inverse their meaning (making sure they now default to the inverse of the corresponding `inclusive` option).

Finally, you failed to update the anchor name in the docs for `preventCursorRight`.

thomasjm:
Thanks for the review! I'll make those changes.

Can you give me a hint on how to run the tests? I've tried `npm install && npm run build && npm run test` and gotten the exact same failures as Travis (which doesn't include a `core_atomicMarker` failure). I also just tried rebasing on the latest master and got the same result.

marijnh:
`npm test` is indeed how you run the tests, and they are passing on the master branch for me.

thomasjm:
Sorry for the delay @marijnh -- I just took another look at this. Responded to your review, fixed up the tests and added specific tests for the fallback behavior. What do you think?

marijnh:
Thanks! Merged.

PR #5780, opened 2019-02-13
Update to dark themes

firephreek:
Lightening the blue link color (cm-link) on vibrant-ink and lesser-dark so that links show up better against the dark background.

theme/vibrant-ink.css     Before/After
vibrant-ink-before
vibrant-ink-after

theme/lesser-dark.css    Before/After
lesser-dark-before
lesser-dark-after


marijnh:
Thanks!

Issue #5781, opened 2019-02-14
Delete key does not work in command mode for Vim


bobbydavid:
In Vim, if you press the Delete key it should delete the current character, like how the 'x' command behaves.

bobbydavid:
Possibly the fix is as simple as adding a mapping in command mode from <Del> to 'x' ?

PR #5783, opened 2019-02-17
Add proper emulation for forward-delete in Vim.

bobbydavid:
In Vim's command mode, forward-delete (Delete on PC or fn-delete on Mac) should
behave the same as 'x': delete the character under the cursor. Due to a bug in
CodeMirror's emulation, this wasn't happening before.

This change adds the proper support, along with tests for forward-delete.

Relevant issue: #5781


marijnh:
Thanks!

PR #5786, opened 2019-02-19
Typo: `npm build` should be `npm run build`

jankeromnes:
Hi! 👋

Wondering why I didn't have a `/lib/codemirror.js`, I grepped through the repository, and found this instruction:

```
npm build
```

However, running it doesn't work:

```
$ npm build
npm WARN build `npm build` called with no arguments. Did you mean to `npm run-script build`?
```

I then guessed the correct instruction, `npm run build`. I suppose this is a typo, so I fixed the changelog and release notes accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -646,7 +646,7 @@ Line endings for pasted content are now normalized to the editor's [preferred en
 
 ### New features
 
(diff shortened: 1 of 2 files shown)


marijnh:
Oops! Thanks for noticing that. Merged.

PR #5792, opened 2019-02-27
Don't % jump to angled brackets

takamori:
Missed in prior commit 3675309aa9a9950a34c0e5cdc5a79e88d4ef9bc5

Repro is in the google_modes index.html test page for C++ editor if vim keyMap is used; the { in line 34 in the editor should not match against line 36's > character, but rather to line 38's } character.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -1978,7 +1978,7 @@
           }
         }
         if (ch < lineText.length) {
-          var matched = cm.findMatchingBracket(Pos(line, ch), {bracketRegex: /[(){}[\]<>]/});
+          var matched = cm.findMatchingBracket(Pos(line, ch), {bracketRegex: /[(){}[\]]/});
           return matched.to;
         } else {
           return cursor;


marijnh:
This was intentionally left in in commit 3675309 , in order to make jumping to angle brackets work in the vim bindings.

@yoongu Apparently, supporting this has the side effect of treating `<` and `>` as mismatched brackets when scanning for other kinds of brackets. In languages where they don't act as brackets, and will thus appear on their own, that is a problem. Can you think of a workaround? Maybe only including those in the string passed to `findMatchingBracket` when actually matching an angle bracket makes sense?

heppe:
Is there a good reason why unmatched brackets of one type need to interfere at all with matching brackets of another?

Also, I never quite figured out why matching the braces in { a-> } would be broken but in { > } wouldn't be.

marijnh:
> Is there a good reason why unmatched brackets of one type need to interfere at all with matching brackets of another?

`findMatchingBracket` was written for bracket-matching, which highlights mismatches as well as matches.

heppe:
Should line 1981 then be replaced with the following?:

          // Only include angle brackets in analysis if they are being matched.
          var bracketRegex =
              (ch === '<' || ch === '>') ? /[(){}[\]<>]/ : /[(){}[\]]/;
          var matched = cm.findMatchingBracket(Pos(line, ch), {bracketRegex});

marijnh:
That looks reasonable—if the vim tests continue to pass with that change I'd be happy to merge it.

marijnh:
Done in #5798

PR #5802, opened 2019-03-11
Handle scanning for user provided brackets, e.g. "

torgeir:
Here's a PR to fix `scanForBracket` support for user provided "brackets", e.g. `"`, that fails with type error.

Editor content with cursor position indicated by |:

```
"wo|o"
```

and calling 

```js
editor.scanForBracket(editor.getCursor(), 1, "string", { bracketRegex: /"/ })
```

results in type error: 

```js
matchbrackets.js:108 Uncaught TypeError: Cannot read property 'charAt' of undefined
    at scanForBracket (matchbrackets.js:108)
    at CodeMirror.scanForBracket (matchbrackets.js:216)
    at <anonymous>:1:8
```

This fixes the issue at hand, making the call return the following, which is what you'd expect.

```json
{
  "pos": {
    "line": 0,
    "ch": 4,
    "sticky": null
  },
  "ch": "\""
}
```

### DIFF
--- a/addon/edit/matchbrackets.js
+++ b/addon/edit/matchbrackets.js
@@ -69,7 +69,7 @@
         var ch = line.charAt(pos);
         if (re.test(ch) && (style === undefined || cm.getTokenTypeAt(Pos(lineNo, pos + 1)) == style)) {
           var match = matching[ch];
-          if ((match.charAt(1) == ">") == (dir > 0)) stack.push(ch);
+          if (match && (match.charAt(1) == ">") == (dir > 0)) stack.push(ch);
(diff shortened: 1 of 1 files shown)


marijnh:
The patch seems innocent enough, so I've merged it, but I don't think matching characters where the code can't see the difference between opening and closing character is going to work—being able to do that is an assumption here.

torgeir:
Yep, that was my conclusion as well, after som further investigation. Well, it works for `scanForBracket`, but probably not for `findMatchingBracket` et.al.. Agreed.

PR #5803, opened 2019-03-11
panda-syntax theme: use monospaced Source Code Pro instead of Source Sans Pro

cben:
https://codemirror.net/demo/theme.html#panda-syntax

#### before:
https://adobe-fonts.github.io/source-sans-pro/ is a proportional font family by Adobe (the "Source" in the name refers to being open source).
image

#### after:
https://adobe-fonts.github.io/source-code-pro/ is a matching monospace font family.
image

(screenshots in chrome on fedora)

cc @siamak

### DIFF
--- a/theme/panda-syntax.css
+++ b/theme/panda-syntax.css
@@ -7,7 +7,7 @@
 	background: #292A2B;
 	color: #E6E6E6;
 	line-height: 1.5;
-	font-family: 'Operator Mono', 'Source Sans Pro', Menlo, Monaco, Consolas, Courier New, monospace;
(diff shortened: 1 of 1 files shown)


marijnh:
Hm, right, since the other fonts are also monospace I guess this fits the spirit of that line. Merged!

cben:
I also see doc/docs.css references 'Source Sans Pro' (and pulls it from google) but that's in context of `body { font-family: 'Source Sans Pro', Helvetica, Arial, sans-serif; }` so clearly deliberate :heavy_check_mark:

PR #5805, opened 2019-03-12
Add keymap for JetBrains

kdekooter:
Please consider adding this keymap for the JetBrains IDE family (IntelliJ, WebStorm, PHPStorm, PyCharm etc.)

### DIFF
--- a/keymap/intellij.js
+++ b/keymap/intellij.js
@@ -1,0 +1,497 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: https://codemirror.net/LICENSE
+
+// A rough approximation of Sublime Text's keybindings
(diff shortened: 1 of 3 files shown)


marijnh:
Unfortunately, the scope of the main distribution has crept outward far enough, and we're no longer accepting contributions of new modules like this, which could also be distributed as separate packages. So that's what I recommend—put it into its own repository and publish it on npm as something like `codemirror-keymap-jetbrains`

kdekooter:
That's too bad. It would be a real selling point to support the most popular key mappings. 
Are there any examples of a separate keymap library?

marijnh:
I don't think so. There are a number of [modes](https://www.npmjs.com/search?q=codemirror%20mode) on npm, though.

PR #5812, opened 2019-03-15
Improves closebrakcets

### LINKED ISSUE #5512: closebrackets addon: configure the end of statement characters
As you can observe in the [demo](https://codemirror.net/demo/closebrackets.html) by typing `(` right before `;` on any line, it won't add the closing `)`, which is quite inconvenient in cases where you switch from a function reference to a function invocation, for example. That is, from `foo = bar;` to `foo = bar(whatever);`

Currently the only way to circumvent this is to add `;;` to the `pairs` option i.e. pass `()[]{}''"";;` - quite a hack I'd say, especially if you want to allow all punctuation characters for some language.

What are you thoughts on solving this? Let us use the above hack? A separate endOfStatement string option? A custom callback function? Checking whether we're typing at the end of a *token*? I don't mind trying to submit a PR if needed.

hanzhaodeng:
Resolves #5512 
Improves the current behavior which auto closes every opening bracket. Improved behaviors are observed from VSCode.

- hardcodes a shuoldClose list of characters that if the next character is in the list, opening a bracket should be auto-closed.
The list is : ")]}'\":;>"
Basically the closing characters in pairs  ; : and >.

- when opening a bracket, it will be auto-closed if:
1. there is no next character,
2. the next character is a white space character.
3. the next character is in the shouldClose list.

### DIFF
--- a/addon/edit/closebrackets.js
+++ b/addon/edit/closebrackets.js
@@ -11,6 +11,7 @@
 })(function(CodeMirror) {
   var defaults = {
     pairs: "()[]{}''\"\"",
+    shouldClose: ")]}'\":;>",
     triples: "",
     explode: "[]{}"
   };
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for looking into this. We're targeting older runtimes that don't have an `includes` method—could you use `indexOf` instead?

Also, I think `closeBefore` would be a clearer name for the option.

hanzhaodeng:
Added a commit. Thanks 👍

marijnh:
Wonderful!

PR #5814, opened 2019-03-18
Update keywords, builtin for MIME x-pgsql. Added missing pgsql keywords

adityatoshniwal:
Keywords based on latest Postgres version 11.2. No change in builtins, checked with the docs.
Also, pgsql keywords which were missing were added. For ref - [](https://github.com/postgres/postgres/blob/REL_11_2/src/pl/plpgsql/src/pl_scanner.c)


marijnh:
Thanks!

PR #5815, opened 2019-03-18
Added a new theme: yonce

thomasmaclean:
Adapted from the popular Bash-It/Visual Studio Code/iTerm/Slack & Alfred theme, Yoncé, by @minamarkham. I say "adapted" as VSCode themes are very complex and allow for a bit more nuance, so I did the best I could. Along those same lines, only ASCII characters seem to work with the demo page, for whatever reason, so I changed the `é` to an `e`. 

Licence from the original VSCode theme:
https://github.com/minamarkham/yonce-vscode/blob/master/LICENSE

More info:
https://yoncetheme.com/


marijnh:
Nice, thanks. Merged as 2c123a4a

PR #5816, opened 2019-03-18
Fix clike mode to only match "_t" at the end of an identifier

vincentwoo:
Currently, clike mode will assign token "type" to C++ variable decls like:

```cpp
int m_total_population;
```

because the regex `.+_t` matches the `_t` part of total. The rule should only apply to names that end in `_t`, as referenced by the link in the comment above the function.


marijnh:
Thanks!

PR #5820, opened 2019-03-19
[xml-hint addon] Optionally match in middle

srguiwiz:
For example, typing "happy" to match completions "very-happy", "not-so-happy", "happy".

Useful for XML with longer, composite names.


marijnh:
That's a reasonable feature. Merged!

PR #5824, opened 2019-03-21
Cancel selection before swapping document

### LINKED ISSUE #5821: Unhandled exception when swapping document during an active selection
:wave: lovely CodeMirror maintainers,

In GitHub Desktop we use CodeMirror in read-only mode to render our syntax highlighted diffs. In an effort to improve performance we recently altered our approach for updating the editor when we detect that a diff has gone stale and we have a new one to present.

Previously we used `setValue` but now we've switched to creating the next Document, doing a little bit of work on it, and then updating the editor using `swapDoc`. After shipping we started seeing crashes due to unhandled exception being thrown in response to mouse movements.

```
Error: There is no line 143 in the document.
    at getLine (app/node_modules/codemirror/lib/codemirror.js:401:34)
    at bidiSimplify (app/node_modules/codemirror/lib/codemirror.js:7543:1)
    at extendTo (app/node_modules/codemirror/lib/codemirror.js:7486:1)
    at extend (app/node_modules/codemirror/lib/codemirror.js:7504:1)
    at CodeMirror.<anonymous> (app/node_modules/codemirror/lib/codemirror.js:7530:13)
    at HTMLDocument.<anonymous> (app/node_modules/codemirror/lib/codemirror.js:3905:12)
```

From what I've been able to gather the cause of this is that people will click on our application (usually hitting the CodeMirror editor's surface area) in order to focus it after having made some changes to their code. When they press their mouse button over the editor CodeMirror will [initiate a selection](https://github.com/codemirror/CodeMirror/blob/5.33.0/src/edit/mouse_events.js#L196).

After that, Desktop reacts to being brought into the foreground, refreshes the contents of the diff, and ultimately calls `swapDoc`. Oftentimes this happens quickly enough that the user hasn't let go of the mouse button by the time we call `swapDoc` and if they move their pointer just a little CodeMirror will receive the [mousemove event](https://github.com/codemirror/CodeMirror/blob/5.33.0/src/edit/mouse_events.js#L321) and attempt to update the selection.

The `leftButtonSelect` function will [capture the current editor document](https://github.com/codemirror/CodeMirror/blob/5.33.0/src/edit/mouse_events.js#L196) into the closure and reuse that for _most_ of the logic related to extending the selection when the mouse moves and, I suspect, then gets into a problem when trying to update the selection if the new document has significantly less lines such that the calculated selection can't fit within the new document.

### Possible solutions

I'm not familiar enough with the code to be certain but it sure looks like there are many attempts at guaranteeing that a selection is clipped to the boundary of the document throughout a selection. It might be as easy as removing the closed-over document reference and ensuring that all methods reacting to mouse events go through the editor to get the current document.

Another, possibly safer, approach would be to cancel any active selection when `swapDoc` get called. This is the workaround I'm planning on including in Desktop to mitigate the problem. I have included the workaround in the reproduction case below. It's obviously not one I'm very happy with as it requires reaching into undocumented internal CodeMirror state (I'm not aware of any exposed API to cancel an active mouse selection).

## Reproduction case

I put together a small-ish reproduction scenario with a hacky workaround, instructions are in the document.

https://jsbin.com/penanesuto/1/edit?html,output

### Environment information

**CodeMirror version**: 5.33.0 (verified to be present in 5.44.0)
**Operating system**: Platform independent
**Browser version**: Electron 3.1.6 (Chromium 66.0.3359.181)

niik:
This change attempts to cancel any active selection prior to swapping the document in the `swapDoc` method in order to avoid conflicting information about the selection range. See the problem described in #5821.

Fixes #5821

### DIFF
--- a/src/edit/mouse_events.js
+++ b/src/edit/mouse_events.js
@@ -305,8 +305,10 @@ function leftButtonSelect(cm, event, start, behavior) {
   function done(e) {
     cm.state.selectingText = false
     counter = Infinity
-    if (e) e_preventDefault(e)
-    display.input.focus()
+    if (e) {
+      e_preventDefault(e)
+      display.input.focus()
+    }
     off(display.wrapper.ownerDocument, "mousemove", move)
     off(display.wrapper.ownerDocument, "mouseup", up)
     doc.history.lastSelOrigin = null
--- a/src/edit/mouse_events.js
+++ b/src/edit/mouse_events.js
@@ -305,7 +305,7 @@ function leftButtonSelect(cm, event, start, behavior) {
   function done(e) {
     cm.state.selectingText = false
     counter = Infinity
-    e_preventDefault(e)
+    if (e) e_preventDefault(e)
     display.input.focus()
     off(display.wrapper.ownerDocument, "mousemove", move)
     off(display.wrapper.ownerDocument, "mouseup", up)
--- a/src/edit/mouse_events.js
+++ b/src/edit/mouse_events.js
@@ -305,6 +305,9 @@ function leftButtonSelect(cm, event, start, behavior) {
   function done(e) {
     cm.state.selectingText = false
     counter = Infinity
+    // If e is null or undefined we interpret this as someone trying
+    // to explicitly cancel the selection rather than the user
+    // letting go of the mouse button.
     if (e) {
       e_preventDefault(e)
       display.input.focus()
--- a/src/edit/methods.js
+++ b/src/edit/methods.js
@@ -423,6 +423,8 @@ export default function(CodeMirror) {
     swapDoc: methodOp(function(doc) {
       let old = this.doc
       old.cm = null
+      // Cancel the current text selection if any (#5821)
+      if (this.state.selectingText) this.state.selectingText()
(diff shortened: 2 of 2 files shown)


src/edit/mouse_events.js
```diff
@@ -305,8 +305,13 @@ function leftButtonSelect(cm, event, start, behavior) {
...
     cm.state.selectingText = false
     counter = Infinity
-    e_preventDefault(e)
-    display.input.focus()
+    // If e is null or undefined we interpret this as someone trying
+    // to explicitly cancel the selection rather than the user
+    // letting go of the mouse button.
+    if (e) {
+      e_preventDefault(e)
+      display.input.focus()
```
niik: @marijnh We didn't discuss this in the issue but I felt it'd be weird if `swapDoc` started fiddling with focus all of a sudden. If there's a better way to work around this I'm all ears.

marijnh: Right, that's slightly kludgey but I agree it makes sense, and with the comment it isn't too confusing

marijnh:
Looks great! Thanks for implementing this.

PR #5825, opened 2019-03-22
[xml-hint addon] Prevent extraneous quote

srguiwiz:
If typing `attribute="` and by `closetag` it becomes `attribute=""` and user presses `Ctrl-Space` to autocomplete and the completion is `"x"` then by this fix  it nicely gives `attribute="x"` instead of obnoxious `attribute="x""`.

Intentionally coded to cause least change to previous behavior, to deal with the frequent most obnoxious case.

Resisted temptation to make changes affecting other cases too.


marijnh:
Thanks!

Issue #5829, opened 2019-03-23
F16 is considered to be Delete key


YellowAfterlife:
If you have any input devices that emit F16 keystrokes, you can find that F16 acts like a Delete key in CodeMirror.

Happens in Chrome 77, Firefox 66, and IE 11.648. Doesn't happen in Edge as it simply doesn't emit any keyboard events for any Fn keys above F15.

The cause seems to be this line:
https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/src/input/keynames.js#L6
127 is a `<DEL>` charcode, but not a Delete keycode (usually 46) - Fn keys start at 112 (F1) and go up to 135 (F24). 

This is consistent across different browsers ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)), except for specific versions of Gecko on Linux, which just don't believe in existence of F13+.

I could submit a PR for removing that key-value pair, but I'm not sure if anything else relies on these mappings.

Example dump from devtools on Chromium:
```js
altKey: false
bubbles: true
cancelBubble: false
cancelable: true
charCode: 0
code: "F16"
composed: true
ctrlKey: false
currentTarget: null
defaultPrevented: false
detail: 0
eventPhase: 0
isComposing: false
isTrusted: true
key: "F16"
keyCode: 127
location: 0
metaKey: false
path: (4) [body, html, document, Window]
repeat: false
returnValue: true
shiftKey: false
sourceCapabilities: InputDeviceCapabilities {firesTouchEvents: false}
srcElement: body
target: body
timeStamp: 497141.3900000043
type: "keydown"
view: Window {postMessage: ƒ, blur: ƒ, focus: ƒ, close: ƒ, parent: Window, …}
which: 127
```

marijnh:
Some of these key codes come from some table that I found in 2008, and it is likely that some of them only happened in antique browsers that no one uses anymore (especially early Opera versions were very creative with key codes). So yes, I think a pull request to remove that code should be safe.

marijnh:
Resolved via #5830

PR #5830, opened 2019-03-23
Do not consider key-127 (F16) to be Delete

YellowAfterlife:
As per https://github.com/codemirror/CodeMirror/issues/5829


marijnh:
Thanks!

PR #5845, opened 2019-04-05
Improve support for minification

SueMaurizio:
When this file is minified, the absence of a ";" after the definition of variable "punctuation" can lead to a run-time error. In my case, I'm using the minification engine embedded in the DNN platform (https://www.dnnsoftware.com).


marijnh:
Sorry, but we don't change our code to address bugs in minifiers. Use a proper one like [babili](https://github.com/babel/minify) if you can, or if you have to use a broken one for reasons outside your control, apply this patch in your own build process.

SueMaurizio:
Sorry, you're right: I'll patch the minifier instead.

Il ven 5 apr 2019, 09:46 Marijn Haverbeke <notifications@github.com> ha
scritto:

PR #5849, opened 2019-04-11
Swift: properly handle empty strings

### LINKED ISSUE #5793: [Swift] empty string highlighting problem
if you go to https://codemirror.net/mode/swift/index.html and put in this:

```swift
print("")
print("")asdfadsfkjdsf//dsfsdfsdf
```

You'll find that the close parens and all the characters afterwards are highlighted like strings, which seems wrong:

image

graue:
Fixes #5793, a bug that was introduced in PR #5675. I'm working for and contributing this code on behalf of @CoderPad.

While writing this, I discovered a couple other bugs with Swift strings that I don't plan to work on at the moment (pre-existing bugs, not regressions; let me know if you want me to file a separate issue for these):

-The mode does not flag it as an error if you fail to include a newline after the opening `"""` or before the closing `"""` of a multiline string
-The mode doesn't properly highlight strings which use [extended string delimiters](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html#ID606)

### DIFF
--- a/mode/swift/swift.js
+++ b/mode/swift/swift.js
@@ -73,8 +73,9 @@
       stream.match("..")
       return "punctuation"
     }
-    if (ch = stream.match(/("{3}|"|')/)) {
-      var tokenize = tokenString(ch[0])
+    var stringMatch
+    if (stringMatch = stream.match(/("""|"|')/)) {
+      var tokenize = tokenString.bind(null, stringMatch[0])
       state.tokenize.push(tokenize)
       return tokenize(stream, state)
     }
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks!

> -The mode does not flag it as an error if you fail to include a newline after the opening """ or before the closing """ of a multiline string

Error reporting isn't really something that is a mode's responsibility, so I don't consider this much of an issue.

> -The mode doesn't properly highlight strings which use extended string delimiters

A pull request for that would be wonderful. No one is actively maintaining this mode, so an issue would likely just sit there.

vincentwoo:
SHOUTOUT TO MY BOY

graue:
Thanks for the quick response and merge, @marijnh! Do you have an idea when you might package this into a release on npm? Weighing whether to patch our copy or not.

marijnh:
Releases happen monthly—the next one is happening late next week.

PR #5855, opened 2019-04-18
Add 1C language to project

sikuda:
Add new language 1C (https://1c-dn.com) to project
like as
- mode: "text/x-1c",
- theme: "1c"

Code for example:
var editor = CodeMirror.fromTextArea(codeArea, {
...
  mode: "text/x-1c",
  theme: '1c'
});

if you want work with russian language in full power - use regexp like this [0-9a-zA-Zа-яёА-ЯЁ&]
Full of examples - code1c.sikuda.ru


marijnh:
Hi Sergey. We are no longer adding new modes to the main distribution. The recommended way to distribute you mode is to put it into its own repository and publish it separately on npm, like the [elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir).

Issue #5854, opened 2019-04-18
Hint are mispositioned if hintOptions.container is defined


Nouzbe:
`hintOptions.container` let users define in which DOM element the hint overlay should be appended.

When this container is given, then the hints `left` and `top` style attributes should be computed relative to it.

[They currently relie on cursorCoords](https://github.com/codemirror/CodeMirror/blob/master/src/measurement/position_measurement.js#L366) which doesn't take their container into account and computes a position relative to the whole page. This results in it being positioned to far down and right in their container.

I can open a PR if needed.

marijnh:
The idea behind the `container` option is that you'd set the dialog's position to something other than `absolute`, so that you can put it in some fixed place. What would be the use of controlling the container element if you continue to position absolutely?

Nouzbe:
Then indeed, I am not using `container` for what it was first intended.:

I contribute to a components library that uses Codemirror along with some CSS to theme its editors. I would like applications using this library to be able to use Codemirror as well, without necessarily being impacted by this CSS. A solution could be to place the library's hints into a div with a class name to which this CSS would be scoped. I was thinking of using the parent element of the editor as `container` for the hints.

marijnh:
That makes sense. A pull request that subtracts the top left corner of the proper `offsetParent` when `container` is given would be welcome.

PR #5857, opened 2019-04-22
[hint addon] Offset the hint to position it correctly when hintOptions.container is given

Nouzbe:
See #5854.

I tested it on the `complete` demo but did not commit changes to the demo file since I believe that having a `container` is not the most common usage. Let met know if you would prefer that this demo uses a container.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -229,14 +229,26 @@
       elt.hintId = i;
     }
 
+    var container = completion.options.container || ownerDocument.body;
     var pos = cm.cursorCoords(completion.options.alignWithWord ? data.from : null);
     var left = pos.left, top = pos.bottom, below = true;
-    hints.style.left = left + "px";
-    hints.style.top = top + "px";
+    var offsetLeft = 0, offsetTop = 0;
+    if (container !== ownerDocument.body) {
+      // We offset the cursor position because left and top are relative to the offsetParent's top left corner.
+      var isContainerPositioned = ['absolute', 'relative', 'fixed'].indexOf(parentWindow.getComputedStyle(container).position) !== -1;
+      var offsetParent = isContainerPositioned ? container : container.offsetParent;
+      var offsetParentPosition = offsetParent.getBoundingClientRect();
+      var bodyPosition = ownerDocument.body.getBoundingClientRect();
+      offsetLeft = (offsetParentPosition.left - bodyPosition.left);
+      offsetTop = (offsetParentPosition.top - bodyPosition.top);
+    }
+    hints.style.left = (left - offsetLeft) + "px";
+    hints.style.top = (top - offsetTop) + "px";
+
     // If we're at the edge of the screen, then we want the menu to appear on the left of the cursor.
     var winW = parentWindow.innerWidth || Math.max(ownerDocument.body.offsetWidth, ownerDocument.documentElement.offsetWidth);
     var winH = parentWindow.innerHeight || Math.max(ownerDocument.body.offsetHeight, ownerDocument.documentElement.offsetHeight);
-    (completion.options.container || ownerDocument.body).appendChild(hints);
+    container.appendChild(hints);
     var box = hints.getBoundingClientRect(), overlapY = box.bottom - winH;
     var scrolls = hints.scrollHeight > hints.clientHeight + 1
     var startScroll = cm.getScrollInfo();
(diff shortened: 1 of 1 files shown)


addon/hint/show-hint.js
```diff
@@ -229,14 +229,17 @@
       elt.hintId = i;
     }
 
+    var container = completion.options.container || ownerDocument.body;
+    var positionOffset = (container.offsetParent || ownerDocument.body).getBoundingClientRect();
     var pos = cm.cursorCoords(completion.options.alignWithWord ? data.from : null);
-    var left = pos.left, top = pos.bottom, below = true;
+    // We offset the cursor position because left and top are relative to the offsetParent's top left corner.
+    var left = pos.left - positionOffset.x, top = pos.bottom - positionOffset.y, below = true;
```
marijnh: `.x` and `.y` don't exist on IE and Edge, so you'll want to use `.left` and `.top` instead.

marijnh:
Did you also test on a scrolled document? It seems that subtracting the top of `document.body` where nothing was subtracted before would break something.

Nouzbe:
You are right, the previous solution did not work with IE/Edge, nor on a scrolled document (whether `container` was given or not).

Here is a new attempt where all of this should be fixed:
- Used `.left` and `.top` in place of `.x` and `.y`
- Didn't change the existing behavior when `hintOptions.container` is `undefined`.
- Used the position offset between the `offsetParent` and `body` in order to not inadvertently offset the scroll, as was previously the case.

addon/hint/show-hint.js
```diff
@@ -229,14 +229,24 @@
       elt.hintId = i;
     }
 
+    var container = completion.options.container || ownerDocument.body;
     var pos = cm.cursorCoords(completion.options.alignWithWord ? data.from : null);
     var left = pos.left, top = pos.bottom, below = true;
+
+    if(container !== ownerDocument.body) {
+      var offsetParentPosition = container.offsetParent.getBoundingClientRect();
```
codeStryke: Hi, I tried this out locally and feel like this line should be `container.getBoundingClientRect();` to get the hint to show up at the exact same position as before. I maybe wrong, but it didn't work for me as is.

marijnh: I guess that depends on whether `container` is positioned or not—I guess we'd want the `offsetParent` of a _child_ of the container.

Nouzbe:
As you pointed out, the previous did not work well if the `container` is positioned.

Here is a new attempt, that should work in this case too.

marijnh:
That should work better, but now you're introducing another layout cycle (changing the DOM first, then measuring it). I think using `getComputedStyle` to figure out whether the container is positioned would be a more efficient approach.

Nouzbe:
Here's a new possible option. It works well too as far as I tested and does not change the points at which the script interacts with the DOM.

Also, I realized that other style updates (aiming at repositioning the hint if it overflows) reassigned `left` and `top` and we need to apply the same offset there too. My bad for not seeing it earlier. Better late than never though.

addon/hint/show-hint.js
```diff
@@ -229,30 +229,41 @@
...
 
+    var container = completion.options.container || ownerDocument.body;
     var pos = cm.cursorCoords(completion.options.alignWithWord ? data.from : null);
     var left = pos.left, top = pos.bottom, below = true;
-    hints.style.left = left + "px";
-    hints.style.top = top + "px";
+    var offsetLeft = 0, offsetTop = 0;
+    if(container !== ownerDocument.body) {
+      // We offset the cursor position because left and top are relative to the offsetParent's top left corner.
+      var offsetParent = ['absolute', 'relative', 'fixed'].includes(parentWindow.getComputedStyle(container).position) ? container : container.offsetParent;
```
marijnh: Another nitpick: `includes` doesn't exist on all browsers that we support. Use `indexOf` instead, or a regexp.

Also, a space after `if` would make the code look more like the rest of the codebase.

Nouzbe:
Sorry about all these roundtrips, and thanks for your patience @marijnh 🙏 

Here's a new one, including your last remarks.

marijnh:
Thanks for bearing with me!

codeStryke:
@Nouzbe thanks for doing this. I was facing the same problem and now I can use this.
@marijnh thanks for the reviewing the changes

PR #5864, opened 2019-05-02
Add a config object for tags and a tag context.

axellew:
Add a config object for tags. This lets us declare properties for tags like:
- How the tag should be indented.
- Which tags creates a new variable scope.
- Which inner mode should be used for the tag.
- Which parser state should be used.
- If the tag has an end tag.

A new context object is used to store the state of the current tag (only for tags which can be closed).

Some basic error handling for when the end tag doesn't match the current open tag.


marijnh:
Thanks!

PR #5866, opened 2019-05-03
Fix matchhighlight vim cursor when background-color is set for cm-matchhighlight

takamori:
This makes the cursor z-index in vim the same as other keymaps.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -57,7 +57,7 @@
   background: #7e7;
 }
 .cm-fat-cursor div.CodeMirror-cursors {
-  z-index: 1;
+  z-index: 3;
 }
 .cm-fat-cursor-mark {
   background-color: rgba(20, 255, 20, 0.5);


marijnh:
Does this solve a specific problem? I can't immediately identify an issue with the patch, but the layering of the selection relative to other stuff is so subtle and fragile that I'm hesitant to make changes to it.

takamori:
Sorry- I was mistaken that it affects midnight theme. If an opaque background-color style is set for cm-matchhighlight, then the cursor becomes invisible with vim keymapping.
 (demo: https://stackblitz.com/edit/js-yu9slo ; look at the CSS file)

Relatedly, it also looks to me like the midnight theme has a typo, btw (it's specifying CodeMirror-matchhighlight instead of cm-matchhighlight).

I suspect we can work around this issue for our usage by making the background not opaque, but it seems like it would be a good idea to address the inconsistency of the cursor z-index values between vim and non-vim.

marijnh:
The matchhighlight theme was indeed using the wrong class name. Attached patch fixes that.

I would propose leaving things as they are until an actual problem is identified—I should have written more comments around this, since there's a lot going on with the layering and I don't remember the details. But as it is, I really don't want to touch it without good reason.

takamori:
Fair enough; I'll drop this then; but then, shouldn't the midnight theme use a translucent background color, rather than an opaque one, to avoid running into the issue I've noted here?

takamori:
(or maybe that style should just be dropped, if nobody has complained about it being missing :D )

takamori:
(hoping that closing the PR doesn't prevent further responses)

marijnh:
Good point. Attached patch drops the styles.

PR #5881, opened 2019-05-15
[ruby] bug fix for curly brace indentation

MyIsaak:
There is a particular annoyance when going through an interactive web tutorial which lies in the auto-indenting of your code -- and by auto-indenting I mean to avoid pressing TAB.

In your interactive Ruby tutorial, and playground this was the case. To reproduce this bug, take for example the following uncompleted ruby playground environment, note the use of -> to mark indents.

```
def main
->5.times {|i|
->->puts i
-># ...
end
```
When I add the second curly brace } after the third line puts i, what I expect is one indent like ->} but what I get is the following
```
def main
->5.times {|i|
->->puts i
->->} # unnecessary indent
end
```
There is an extra indent, or in other words, a missing or faulty auto formatting.

I have provided a simple bug fix that solves the indenting issue for curly braces.

### DIFF
--- a/mode/ruby/ruby.js
+++ b/mode/ruby/ruby.js
@@ -280,7 +280,16 @@ CodeMirror.defineMode("ruby", function(config) {
       if (state.tokenize[state.tokenize.length-1] != tokenBase) return CodeMirror.Pass;
       var firstChar = textAfter && textAfter.charAt(0);
       var ct = state.context;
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for reporting that. I've pushed a slightly simplified patch which doesn't use ES6 features (modes don't have a build process at the moment). See attached.

PR #5882, opened 2019-05-15
[emacs] Add "redo" keybinding in Emacs keymap

aiguofer:
Since "redo" is generally assigned to Ctrl+Y, we lose this binding
when using the Emacs keymap. The Emacs default for "redo" is a bit
complicated, but since we're using Ctrl-Z for "undo", it makes sense
to have Ctrl+Shift-Z for "redo".


aiguofer:
Also see https://github.com/ipython-contrib/jupyter_contrib_nbextensions/issues/1331

PR #5896, opened 2019-05-30
Autocomplete/Hint: Selection of multiple options with ctrl + click

antoniohdez:


marijnh:
This seems like a lot of code for a very obscure feature. What would be a situation in which it is useful?

antoniohdez:
This could be useful for SQL hints, when selecting multiple columns or table names for joins.
I know it may seem like a lot of code, If you want I can make a code refactoring to make it cleaner.

marijnh:
Sorry, but I think I'd rather not have this in the library—almost nobody will know how to use it, and in most circumstance it'll just produce invalid code. You're going to have to use your own fork of show-hint.js.

PR #5900, opened 2019-06-06
Add U+FFFC as a default replaced special character

mtaran-google:
It doesn't have a very useful meaning per https://en.wikipedia.org/wiki/Specials_(Unicode_block), but more notably it currently shows up as an invisible character and thus it's very easy to mess up documents with it.


marijnh:
Good idea.

marijnh:
Any reason not to include the whole `\ufff9-\uffff` range?

mtaran-google:
No particular reason, just that was the one we ran into...

PR #5901, opened 2019-06-09
Remove 'float' as Java keyword

### LINKED ISSUE #5899: float is both a keyword and type in Java
In the clike mode, "float" is specified to be both a keyword and a type for Java. While this is technically a Java keyword, none of "int", "double", etc. are listed as keywords. Should this just be a type?

dklilley:
This is a change to the Java mode to resolve #5899.


marijnh:
Thanks!

PR #5913, opened 2019-06-19
[liquid mode] Add

vrana:
See documentation at https://shopify.github.io/liquid/.

### DIFF
--- a/mode/liquid/index.html
+++ b/mode/liquid/index.html
@@ -1,0 +1,48 @@
+<!doctype html>
+
+<title>CodeMirror: Liquid mode</title>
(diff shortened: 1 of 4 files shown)


marijnh:
Thanks, but we're not accepting any further modes into the main distribution. The recommended way to manage an additional mode is to keep it in its own repository and publish it to npm under a name like `codemirror-mode-liquid` (see for example [the elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir)).

vrana:
Published at https://github.com/HlidacStatu/codemirror-mode-liquid.

marijnh:
Thanks!

PR #5914, opened 2019-06-19
Added Svelte component support

gomezcabo:
Added Svelte component files support based on Vue.js components mode.


marijnh:
Thanks, but we're not accepting any further modes into the main distribution. The recommended way to manage an additional mode is to keep it in its own repository and publish it to npm under a name like `codemirror-mode-svelte` (see for example [the elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir)).

PR #5916, opened 2019-06-21
Mode klartext

drunsinn:
Hello everyone,

this is a new mode for CodeMirror to support a language used on some CNC controls.
The file extensions is 'H' so in theory there might be a problem with header files for C if a upper case H is used.


marijnh:
Hi. We aren't accepting new modes into the main distribution anymore. The recommended way to distribute your mode is to put it in its own npm package, and name it something like codemirror-mode-klartext

PR #5920, opened 2019-06-27
Add Flow language mode

goodmind:
https://flow.org

### DIFF
--- a/mode/javascript/flow.html
+++ b/mode/javascript/flow.html
@@ -1,0 +1,62 @@
+<!doctype html>
+
+<title>CodeMirror: Flow mode</title>
+<meta charset="utf-8"/>
+<link rel=stylesheet href="../../doc/docs.css">
+
+<link rel="stylesheet" href="../../lib/codemirror.css">
+<script src="../../lib/codemirror.js"></script>
+<script src="../../addon/edit/matchbrackets.js"></script>
+<script src="javascript.js"></script>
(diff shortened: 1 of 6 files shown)


goodmind:
I also found bug with `typeof X` in `extends` clause in TypeScript, so I added failing tests here

marijnh:
Hi, thanks for the contribution. But we aren't including new modes into the main distribution anymore. The recommended way to distribute code like this is to put it into its own npm package with a name like codemirror-mode-flow

marijnh:
Ah, I see you extended the javascript mode, rather than adding it as a new mode. I'll look into this in more detail next week, though my first thought is that that mode is already way too complex as it is.

marijnh:
Though it does add more complexity to an already big mode, this is a very solid PR implementing a useful feature, so I guess I have no choice but to merge it 😄

marijnh:
On closer look, it seems that a lot of the failing tests are not intentional. Could you take care of those first?

goodmind:
There's bug with `<Y extends typeof X>(x: Y) <- x becomes variable instead of def` that I don't know how to fix

marijnh:
Right, but there's a bunch of tests failing that don't use that construct.

goodmind:
You mean with indentation?

goodmind:
I didn't really understand how it works and couldn't make it work for exact objects, so it's really off
image
image

Not sure why regular objects failing

Issue #5926, opened 2019-07-04
[javascript mode] numeric separators are not supported


hashseed:
[Numeric separators](https://github.com/tc39/proposal-numeric-separator) is a new language proposal that is already being supported in Node.js and Chrome. I would like to add support to enable correct syntax highlighting in Chrome DevTools (which uses CodeMirror).

PR #5927, opened 2019-07-04
[javascript mode] Support numeric separators

hashseed:
Note that the new regular expression is more permissive than
the spec proposal, e.g. `0_._1` or `1__2` are not valid numbers.

However, filtering out these cases would make the tokenizer
unnecessarily complex.

Fixes: https://github.com/codemirror/CodeMirror/issues/5926


marijnh:
Thanks!

PR #5928, opened 2019-07-05
[vim mode] match vim char escape substitution behavior

ilyakharlamov:
currently in vim mode 
replacements behavior does not match vim's
for substitutions like :%s/\//slash/g supposed to replace "/" -> "slash"
however currently in codemirror it shows notification "no matches for \\"
the provided PR fixes the behavior in vim pcre mode to match vim's


marijnh:
Thank you!

PR #5933, opened 2019-07-10
Reflect tab unit and size for defaultTab command

paullewis:
I've tracked a Chromium bug (https://bugs.chromium.org/p/chromium/issues/detail?id=977394) back to the fact that CodeMirror assumes tabs should be done with insertTab execCommand rather than using indentLine. This PR changes the behavior such that the tab unit and size are used instead.


marijnh:
We're not going to change the default behavior of the library to align it with some downstream user's expectation. You can bind another command to tab if you want to customize what it does.

paullewis:
I see. Thanks for reviewing.

PR #5937, opened 2019-07-11
[sparql mode] Non-ASCII variable names

### LINKED ISSUE #5935: SPARQL variables names with non-ASCII characters are not colorized
In the mode SPARQL, variables names are colorized up to the first non-ASCII character, although some are valid in the SPARQL standard. For instance try to add an " é " on the variable " ?family " in the [SPARQL demo](https://codemirror.net/mode/sparql/index.html).

The SPARQL standard precisely defines the acceptable characters for a [variable name](https://www.w3.org/TR/sparql11-query/#rVARNAME).

I will submit a pull request in some hours.

Seb35:
This regex is ES6-compliant and SPARQL-compliant, but it not ES3 or ES5 compliant.

This should be preferred if compatibility with old JavaScript (pre-ES6) is not required. Else the other PR #5936 should be preferred. Both PR are excluding.

Closes #5935

### DIFF
--- a/mode/sparql/sparql.js
+++ b/mode/sparql/sparql.js
@@ -41,7 +41,7 @@ CodeMirror.defineMode("sparql", function(config) {
       if(ch == "?" && stream.match(/\s/, false)){
         return "operator";
       }
-      stream.match(/^[\w\d]*/);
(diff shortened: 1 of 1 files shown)


marijnh:
We're still targeting ES3 browsers.

PR #5944, opened 2019-07-18
Dart class tokens

johnpryan:
This adds additional hooks to the Dart mode. This uses the `variable-2` token type for tokens matching `Uppercase` or `_UppercaseWithUnderscore` where `_lowercaseUnderscore` are assumed to be normal variables.

The Dart mode uses the `clike` mode, so this uses the `hooks` feature from that to avoid writing a  custom `token()` function.

### DIFF
--- a/mode/dart/dart.js
+++ b/mode/dart/dart.js
@@ -112,7 +112,7 @@
   function className(stream) {
     stream.backUp(1);
     if (stream.match(/[_$]*[A-Z][a-zA-Z0-9_$]*/, true)) {
-      return "class";
+      return "variable-2";
     }
 
     // variable starting with an underscore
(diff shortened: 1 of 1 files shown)


marijnh:
Couldn't the `token` hook (as supported by the clike mode) do this in a less awkward way?

johnpryan:
@marijnh I wasn't aware of that, thanks for the suggestion. Could you take another look?

marijnh:
Still some potential improvements:

 - Since the `token` hook is called after the main tokenizer has already categorized the token, you could look at its third argument and compare that to `"variable"` to avoid re-doing the same work (matching your regexp when the tokenizer didn't match a variable token)

 - `String.match` doesn't take a second argument (you're passing it `true` for some reason)

johnpryan:
I think I confused `String.match()` with `stream.match()`. There was also an issue using `String.match()`  so this now uses `RegExp.test()` to test that the entire token matches the isUpper regex.

Thanks for the feedback!

marijnh:
Thanks!

PR #5946, opened 2019-07-18
Improve performance of website

jMuzsik:
Hello!

Well, I had the idea to minify several of the largest and most used JS files being passed around in HTML to improve the performance of the code mirror website as well as faster loading scripts for a few other HTML files.

Lighthouse score changes for the homepage on my side:

Screenshot 2019-07-18 at 5 07 19 PM

And what it is currently for me (codemirror.net):

Screenshot 2019-07-18 at 5 07 46 PM

I also did some improvements for SEO and Accessibility.

But I did touch a lot of code, which makes this daunting to view. I got caught in a rabbit hole, not thinking there was so many `index.html` files. I can remove any alterations on files that are not on the main website if you'd prefer that. I did make sure that all paths are correct though.

Tests pass, lint passes. I only touched HTML files, and minified several JS/CSS files.

Thanks! And thanks for the things you do for the community. I loved Eloquent JS when I first began learning to code.


marijnh:
Thanks for working on this, but it is not something I'm interested in at this time. The current website is intentionally primitive so that it can be opened directly from the git repository. I am not interested in including minified scripts (or any kind of build artifacts) to the repository (it means noisy, big diffs whenever you make a change to their source files).

The next major release will involve a completely overhauled website, as a separate repository, with a build process, at which point there won't be so many independent html files, and it'll be easier to set up a minification step.

PR #5947, opened 2019-07-19
[javascript mode] fix tokenizing of underscore properties

hashseed:
The regexp for parsing numbers with numeric separators is too
broad. This way, we mistake properties starting with underscore
to be numbers.


marijnh:
Thanks!

PR #5953, opened 2019-07-24
remove unmaintained codev.it

g5li:
The domain http://codev.it is Inaccessible,  the [latest twitter](https://twitter.com/codevit) was post on Sep 2015.
It's better to remove it.


marijnh:
Thanks!

PR #5955, opened 2019-07-29
Foldgutter: make it possible to redefine the `widget` prop

edoroshenko:
Currently if I want to specify a custom widget for folded ranges it would not work when I fold a range using the gutter. This happens because `onGutterClick` function passes through only `rangeFinder`  option.
The solution is to pass all options, which is supported by the `foldCode` method.


marijnh:
That's reasonable. Thanks.

PR #5957, opened 2019-08-01
[handlebars mode] add support for triple mustache tags

nknapp:
This PR adds support for triple mustaches like 

```hbs
{{{rawHtml}}}
```
that can be used in Handlebars to avoid HTML-escaping the output.


nknapp:
if there is anything else I should to the make this PR mergeable, please let me know.

marijnh:
Looks good!

PR #5963, opened 2019-08-06
Make ".CodeMirror pre" selector more specific

thomasjm:
Currently, `lib/codemirror.css` contains a broad selector for `.CodeMirror pre` in a couple places. It seems intended to match the `pre` tags corresponding to `.CodeMirror-line` and also the ones for measurement.

However, if you want to create a widget or line widget that contains a `pre` tag, it ends up inheriting a bunch of styles you don't expect.

Since `pre` tags seem to only be created in a few places, I searched through the code and made sure they always have an appropriate classname. Then I modified the selector to only target the desired CodeMirror elements.

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -13,7 +13,8 @@
 .CodeMirror-lines {
   padding: 4px 0; /* Vertical padding around content */
 }
-.CodeMirror pre {
+.CodeMirror pre.CodeMirror-line,
+.CodeMirror pre.CodeMirror-measure-elem {
   padding: 0 4px; /* Horizontal padding of content */
 }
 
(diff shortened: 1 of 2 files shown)


marijnh:
You've added all of lib/codemirror.js to the repository in that patch. Could you remove that again?

thomasjm:
Oh whoops -- fixed

marijnh:
Thanks!

thomasjm:
Oh I thought there might be some discussion about this @marijnh -- I've played around with the editor and everything seems fine but I'm not sure how to 100% tell this doesn't break anything. But if you're comfortable with it I'm happy :)

marijnh:
It looks harmless enough to me (I can imagine some people accidentally or intentionally, relying on getting these styles in embedded `<pre>` tags, but that seems obscure enough to not worry about too much).

PR #5967, opened 2019-08-08
Fix placeholder styling after <pre> change

thomasjm:
Turns out my previous PR #5963 did actually break something -- the `placeholder.js` addon was rendering wrong without these styles. This PR re-adds them for the `pre.CodeMirror-placeholder` class.

The placeholder appears to be the only addon that does this.

### DIFF
--- a/addon/display/placeholder.js
+++ b/addon/display/placeholder.js
@@ -39,7 +39,7 @@
     var elt = cm.state.placeholder = document.createElement("pre");
     elt.style.cssText = "height: 0; overflow: visible";
     elt.style.direction = cm.getOption("direction");
-    elt.className = "CodeMirror-placeholder";
+    elt.className = "CodeMirror-placeholder CodeMirror-line-like";
     var placeHolder = cm.getOption("placeholder")
     if (typeof placeHolder == "string") placeHolder = document.createTextNode(placeHolder)
     elt.appendChild(placeHolder)
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -14,7 +14,7 @@
   padding: 4px 0; /* Vertical padding around content */
 }
 .CodeMirror pre.CodeMirror-line,
-.CodeMirror pre.CodeMirror-measure-elem {
+.CodeMirror pre.CodeMirror-line-like {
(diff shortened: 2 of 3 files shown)


marijnh:
Couldn't we just have the placeholder addon attach a `CodeMirror-line` class to its element instead?

thomasjm:
Well, there are a few styles in `codemirror.css` that are for `.CodeMirror-line` specifically, so that could introduce a danger of picking up some of those (although I don't think it will at the moment).

That solution seems a bit messy to me, since the placeholder isn't really the same as the other `.CodeMirror-line`s--I'm used to thinking of those as the children of the `.CodeMirror-code`, but that's not even where the placeholder gets attached.

I think the cleanest solution would be for the placeholder addon to set the styles it needs by itself when it's building the `<pre>`.

marijnh:
Maybe `CodeMirror-measure-elem` should be renamed to something more generic that both the measure element and any other code that wants to create a line-like element can use. How about `CodeMirror-line-like`?

thomasjm:
Hmm, it seems like overkill to create a whole new "line-like" concept if the placeholder addon is the only thing that will use it. But if there are other situations where the line-like class can be useful then it could be good. 

It also feels a little fragile to me to define "line-like" as "this bucket of styles that comes from `codemirror.css`". But this sort of decision may be above my pay grade :P

I'm a big fan in general of moving styling from CSS directly into JS, where there's often less distance between the code and the intended result. To address the common need of lines, measurement elems, and placeholder to all be laid out exactly the same way, why not enshrine the concept of "make me a `<pre>` tag that has this line-like set of styles" into a JS function `lineLikePreElt` that they can use to create the element? (It seems like `codemirror.css` is for things the user might conceivably want to override, and these sorts of styles are best moved out of there.)

marijnh:
This version of CodeMirror isn't going to move to a CSS-in-JS style approach anymore.

Adding a special-cased for a specific plugin to the main CSS file seems messier than having a vaguely named `line-like` class that both that and the measure element use.

thomasjm:
Okay, I just updated the PR with the `CodeMirror-line-like` class

PR #5971, opened 2019-08-14
significant speed optimisation when large amount of lines are folded

wai-fu:
UI begins to lag when a block of text over 10000 lines are folded. Browser will freeze for minutes when line count increased to 100K.
The isFolded function only requires looking forward, optimised the code by passing the -1 argument to the TextMarker.find method.
Code folding is now instantaneous even with 100K lines involved.


marijnh:
Thanks!

PR #5973, opened 2019-08-16
Add passive event listeners to Chrome audit recommended events

mfix22:
Chrome recommends that using `passive` event listeners for wheel, mousewheel, touchstart, and touchmove events ([source](https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners?utm_source=lighthouse&utm_medium=devtools))

Including CodeMirror causes this warning in the Chrome audit as well:
Screenshot 2019-08-15 18 12 23

If this change is not desired, feel free to just close this PR 👍

### DIFF
--- a/src/util/event.js
+++ b/src/util/event.js
@@ -10,7 +10,8 @@ const noHandlers = []
 
 export let on = function(emitter, type, f) {
   if (emitter.addEventListener) {
-    emitter.addEventListener(type, f, false)
+    const passive = type === 'wheel' || type === 'mousewheel' || type === 'touchstart' || type === 'touchmove'
+    emitter.addEventListener(type, f, passive ? {passive} :false)
   } else if (emitter.attachEvent) {
     emitter.attachEvent("on" + type, f)
   } else {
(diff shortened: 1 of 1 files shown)


marijnh:
The wheel handlers _do_ sometimes call `preventDefault`, so there this isn't appropriate.

marijnh:
Oh, and touchdown can also call `preventDefault` (via `clickInGutter`), so it seems only `touchmove` could benefit from this.

mfix22:
@marijnh makes sense! 

I am not sure it is worth adding this then. Feel free to close if you think the same.

mfix22:
@marijnh out of curiosity, do you know why `preventDefault` is called for `mousewheel` events?

marijnh:
Because on horizontal scroll, the editor needs to realign the gutter before the content is scrolled (or it'll jank around)

I agree that this might not be worth it then, and adding such an exception in a general-purpose utility function (which client code may also be using) seems problematic in other ways as well. Going to close the PR.

PR #5982, opened 2019-08-23
Stop using the `keypress` event.

joelburget:
This change was motivated by noticing that vim-mode behaves differently
on my mac vs linux boxes. In particular, when holding down a movement
key (eg 'j' or 'k') on linux, the movement repeats, so I can move 20
lines with one keypress. However, on my mac, on keypress translates into
one movement, no matter how long I hold the key. Note: I'm running
Chrome 76 on both machines.

I noticed that the actual event handling is being done via the
`keypress` event, which is deprecated, and I believe we should be able
to do the same handling on the `keydown` event.

We kept the `onKeyPress` handler, though it's not triggered by
`keypress` events anymore. This is to maintain backwards compatibility
with `CodeMirror.triggerOnKeyPress` (which could be deprecated).

We now get the character from `e.key` rather than
`String.fromCharCode(charCode == null ? keyCode : charCode)`. In my
testing I saw in instance where I pressed 'j' and
* `e.keyCode == 74` (though `String.fromCharCode(74) == 'J'`, it should
  really be 106 for `j`)
* `e.charCode == 0`
* `e.key == 'j'`

So I decided to just go with `.key`, which is well supported according
to Mozilla
(https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent#Browser_compatibility).

Testing:
* I ran the test suite (test/index.html) locally. There were three
  failures (core_move_bidi_Όȝǝڪȉۥ״ۺ׆ɀҩۏ ҳ, core_move_bidi_ŌӰтقȤ؁ƥ؅٣ĎȺ١
  Ϛ, and core_issue_4878), but these also appear as failures on master.
* I verified that I can type and that holding movement keys repeats
  movement on the Vim bindings demo (demo/vim.html).
  - Chrome 76 on linux
  - Chrome 76 on mac
  - Firefox 60 on linux
* Ran `bin/lint`

### DIFF
--- a/src/edit/CodeMirror.js
+++ b/src/edit/CodeMirror.js
@@ -18,7 +18,7 @@ import { bind, copyObj, Delayed } from "../util/misc.js"
 
 import { clearDragCursor, onDragOver, onDragStart, onDrop } from "./drop_events.js"
 import { ensureGlobalHandlers } from "./global_events.js"
-import { onKeyDown, onKeyPress, onKeyUp } from "./key_events.js"
+import { onKeyDown, onKeyUp } from "./key_events.js"
 import { clickInGutter, onContextMenu, onMouseDown } from "./mouse_events.js"
 import { themeChanged } from "./utils.js"
 import { defaults, optionHandlers, Init } from "./options.js"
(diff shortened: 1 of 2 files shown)


marijnh:
Did you see the list of [supported browsers](https://codemirror.net/#browsersupport)? IE8 definitely doesn't have `.key`, and if I remember correctly it's pretty broken on IE9-10.

joelburget:
Thanks for pointing out the supported browsers (which I hadn't seen). I'm curious, would you be open to using a [polyfill](https://github.com/cvan/keyboardevent-key-polyfill)?

marijnh:
[CodeMirror 6](https://github.com/codemirror/CodeMirror.next) (obviously) doesn't use the keypress event anymore. I don't think it's worthwhile making drastic changes (with the chance of regressions that entails) to version 5 anymore for this. Browsers will continue to fire keypress for the foreseeable future.

joelburget:
Makes sense 👍

PR #5992, opened 2019-09-05
Schedule a redraw on changes instead of change

alur:
The setTimeout/clearTimeout overhead of re-scheduling on every change in a large operation can be significant. This cuts the time of a particularly large operation I was testing from 5s to 2.5s (replacing some text in ~5000 lines). The majority of the rest of the time coming from rescheduling of the highlight worker inside the startWorker call.


alur:
Actually, I missed foldgutter earlier. Doing the same thing there got the time down to 1.88s.

I also tweaked Delayed in util a bit and got it to down to 450ms. I'll clean that up a little and send it as a separate PR.

marijnh:
Thanks, that's a good idea.

PR #5993, opened 2019-09-06
Reduce setTimeout/clearTimeout calls from Delayed

alur:
This reduces the time the operation I mentioned in https://github.com/codemirror/CodeMirror/pull/5992 takes to ~450ms.

### DIFF
--- a/src/util/misc.js
+++ b/src/util/misc.js
@@ -37,18 +37,19 @@ export class Delayed {
   }
   onTimeout(self) {
     self.id = 0
-    if (self.time < Date.now()) {
+    if (self.time < +new Date) {
       self.f()
     } else {
-      setTimeout(self.handler, self.time - Date.now())
+      setTimeout(self.handler, self.time - +new Date)
     }
   }
   set(ms, f) {
     this.f = f
-    const time = Date.now() + ms
+    const time = +new Date + ms
     if (!this.id || time < this.time) {
       clearTimeout(this.id)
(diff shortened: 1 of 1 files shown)


marijnh:
This doesn't look like it's been tested—`onTimeout` is referred to without `this`, which seems like it should definitely break the code.

alur:
Sorry about that. I originally wrote it in the compiled codemirror.js and thought I could rewrite it correctly in the source :)

marijnh:
I had no idea setTimeout was _that_ expensive.

We still support IE8, so `Date.now` can't be used. Does this still perform as well using `+new Date`?

tophf:
setTimeout is indeed very expensive, in Violentmonkey we've switched to using just one setTimeout call per event loop cycle in 99% of cases  instead of hundreds ([code](https://github.com/violentmonkey/violentmonkey/blob/master/src/common/cache.js)) so when the timer function runs it performs housekeeping on the entire cache and optionally reschedules itself to the closest expiry time.

tophf:
>suspect using the performance tab in Chrome's devtools adds some extra overhead

Just to be sure you can measure the time without devtools by using console.time('foo') and console.timeEnd('foo'). 

<details><summary>Or run the low-overhead version of the devtools profiler</summary> 

chrome_2019-10-08_17-46-36

It also has a flamechart mode similar to the full profiler.
</details>

alur:
Yeah, I took those measurements using console.time :)

PR #6010, opened 2019-09-23
[foldgutter] reuse old markers if indicatorOpen/Folded is string

tophf:
On large documents foldgutter spends a lot of time (0.2s, 0.5s, or even a second) unnecessarily forcing CodeMirror to recreate the unchanged mark elements:

before

With this PR on a large document we'll spend 10+ times less on folding+rendering:

after

* The built-in indicator element for open/folded states differs only by its className so we can optimize for this case. This PR leaves the old marker unchanged if its className contains the new class.

* With a user-provided element the only case we can optimize for is to skip removing the marker when there were none. Such a check is implemented in this PR.

* `classTest` function was copied from [/src/util/dom.js](https://github.com/codemirror/CodeMirror/blob/master/src/util/dom.js) because AFAICT we can't use [classList.contains](https://developer.mozilla.org/en-US/docs/Web/API/element/classList) in some browsers still supported by CodeMirror.

### DIFF
--- a/addon/fold/foldgutter.js
+++ b/addon/fold/foldgutter.js
@@ -71,23 +71,35 @@
   }
 
   function updateFoldInfo(cm, from, to) {
-    var opts = cm.state.foldGutter.options, cur = from;
+    var opts = cm.state.foldGutter.options, cur = from - 1;
     var minSize = cm.foldOption(opts, "minFoldSize");
     var func = cm.foldOption(opts, "rangeFinder");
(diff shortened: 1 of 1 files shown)


marijnh:
Looks good. Why did you move the `++cur` statement? It doesn't look like the old approach would change anything about the code.

tophf:
I moved it because I've added several early `return`s so I had to start with `from - 1` as well. I still think this is the shortest solution.

marijnh:
That makes sense. Thanks.

PR #6016, opened 2019-09-28
adjustments for eclipse java

sterlp:
- added void as keyword
- used correct color annotations

I haven't found a CSS class telling that it is java - please check the changes and provide advice if where is a better solution.

# Open:
`
.cm-s-eclipse span.cm-type + span.cm-variable {
	color: #00f !important;
}
`
primitive types have a wrong color too:
`
jQuery( ".cm-s-eclipse span.cm-type:contains('boolean')" )
  .css({"color": "#7F0055", "font-weight": "bold"});
});
`
# Result the changes
image

## Fixed
- void has the correct color
- boolean return type has the correct color // JS local fix
## Needs work
- Class name has still the wrong color
- String as a primitive type as still a wrong color
- methods have a wrong color

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -470,7 +470,7 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
                     "do else enum extends final finally for goto if implements import " +
                     "instanceof interface native new package private protected public " +
                     "return static strictfp super switch synchronized this throw throws transient " +
-                    "try volatile while @interface"),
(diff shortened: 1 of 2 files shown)


marijnh:
> added void as keyword // so that it is like public and class

Despite what Eclipse does here, `void` is a type so I think highlighting it as such is reasonable.

In general, creating exactly the same highlighting as Eclipse is not a goal for this project.

PR #6018, opened 2019-10-04
Add alt attribute to codemirror logo images

oscarlofwenhamn:
Having images without an "alt" causes bug reports from certain scanning tools such as SonarQube. Adding an alt is a small effort to fix this bug, this fixes all instances related to the CodeMirror logo.

### DIFF
--- a/mode/apl/index.html
+++ b/mode/apl/index.html
@@ -12,7 +12,7 @@
 	.CodeMirror { border: 2px inset #dee; }
     </style>
 <div id=nav>
-  <a href="https://codemirror.net"><h1>CodeMirror</h1><img id=logo src="../../doc/logo.png" alt="CodeMirror logo"></a>
+  <a href="https://codemirror.net"><h1>CodeMirror</h1><img id=logo src="../../doc/logo.png" alt=""></a>
 
   <ul>
     <li><a href="../../index.html">Home</a>
--- a/mode/apl/index.html
+++ b/mode/apl/index.html
@@ -12,7 +12,7 @@
 	.CodeMirror { border: 2px inset #dee; }
     </style>
 <div id=nav>
-  <a href="https://codemirror.net"><h1>CodeMirror</h1><img id=logo src="../../doc/logo.png"></a>
+  <a href="https://codemirror.net"><h1>CodeMirror</h1><img id=logo src="../../doc/logo.png" alt="CodeMirror logo"></a>
 
   <ul>
     <li><a href="../../index.html">Home</a>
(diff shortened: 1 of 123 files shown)


marijnh:
Wouldn't an `alt=""` be more appropriate, since the logo doesn't carry any useful information?

oscarlofwenhamn:
I think describing the image as the logo at hand is, albeit "on the nose", appropriate. `alt=""` would still be considered non-compliant to e.g. SonarQube, as per the following:

> The alt attribute provides a textual alternative to an image.
> 
> It is used whenever the actual image cannot be rendered.
> 
> Common reasons for that include:
> * The image can no longer be found
> * Visually impaired users using a screen reader software
> * Images loading is disabled, to reduce data consumption on mobile phones
>
> Empty alt attributes are not allowed, since purely decorative images should be specified using CSS, not using the img tag.

marijnh:
Here's a W3C page recommending empty alt attributes: https://www.w3.org/WAI/tutorials/images/decorative/ . Where does the text you quoted come from?

oscarlofwenhamn:
https://rules.sonarsource.com/html/RSPEC-1077

I managed to find this article on the subject, by the creators of SonarQube. It does differ from the recommendations given by the bug report in that on this page, it is stated that

> It is also very important to not set an alt attribute to a non-informative value. For example logo is useless as it doesn't give any information to the user. In this case, as for any other decorative image, it is better to use a CSS background image instead of an  tag. If using CSS background-image is not possible, an empty alt="" is tolerated. See Exceptions bellow.
> ...
>
> Exceptions
> \ tags with empty string alt="" attributes won't raise any issue. However this technic should be used in two cases only:
> 
> When the image is decorative and it is not possible to use a CSS background image. (...)
> When the image is not decorative but it's alt text would repeat a nearby text. For example, images contained in links should not duplicate the link's text in their alt attribute, as it would make the screen reader repeat the text twice.
```
<a href="flowers.html">
    <img src="tulip.gif" alt="" />
    A blooming tulip
</a>
```

Arguably this could be a case of the latter, since the `<h1>CodeMirror</h1>` is right next to it, thought it is not as clear cut as the example above with the tulip. 

Honestly I would be fine with an `alt=""` especially if there is a source confirming it is ok praxis, though if it were up to me (which it isn't) I would have preferred a non-empty attribute.

marijnh:
I prefer to follow the official recommendation over what a some tool prefers, so yeah, if you'd update the PR to use an empty attribute I'll merge that.

marijnh:
Thanks!

oscarlofwenhamn:
Likewise!

PR #6020, opened 2019-10-06
[yaml-frontmatter mode] Allow pandoc-style closing with `...`

seifferth:
This allows to end yaml-frontmatters with either `---` or `...`, as
suggested [on the discussion board][1]. The old behaviour was to just
accept `---`.

[1]: https://discuss.codemirror.net/t/pandoc-markdown-style-yaml-frontmatter/2182


marijnh:
Shouldn't this also apply to the opening marker?

seifferth:
Only the closing marker allows optional dots in pandoc's markdown. See <https://pandoc.org/MANUAL.html#extension-yaml_metadata_block>.

marijnh:
Oh, weird. Merging!

PR #6022, opened 2019-10-07
[continuecomment] skip some edge cases, respect indentWithTabs

tophf:
When pressing <kbd>Enter</kbd>:
* line comment will always be continued if the next line is a line comment
* line comment won't be continued if cursor is at EOL or inside trailing space(s) at EOL
* line comment won't be continued if cursor is at position 0 and the line comment starts at 0 too. An empty line will be added above just like before but the original line won't change, whereas previously it would be mangled like `// // comment`.
* block comment won't be continued if it's inside a line comment e.g. `// /*`
* the indentation of block continuation respects `indentWithTabs` option.

### DIFF
--- a/addon/comment/continuecomment.js
+++ b/addon/comment/continuecomment.js
@@ -9,6 +9,8 @@
   else // Plain browser env
     mod(CodeMirror);
 })(function(CodeMirror) {
+  var nonspace = /\S/g;
+  var repeat = String.prototype.repeat || function (n) { return Array(n + 1).join(this); };
   function continueComment(cm) {
     if (cm.getOption("disableInput")) return CodeMirror.Pass;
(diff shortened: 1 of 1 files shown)


addon/comment/continuecomment.js
```diff
@@ -9,6 +9,8 @@
   else // Plain browser env
     mod(CodeMirror);
 })(function(CodeMirror) {
+  var nonspace = /\S/g;
+  var repeat = String.prototype.repeat || function (n) { return Array(n).join(this); };
```
marijnh: The fallback doesn't quite work—should be `Array(n + 1)` since join will only join _between_ the elements.

tophf: Indeed!
Fixed.

PR #6039, opened 2019-10-25
Useless/bad string escapes

brettz9:
lgtm.com-inspired fixes: Change useless string escapes into useful or non-confusing ones; escape backslashes when performing an escape that adds backslashes


marijnh:
These are almost all wrong—they change the meaning of strings by turning ignored backslashes into escaped ones, or in some cases do even stranger things like replacing a forward slash with a backslash. As it is, this is not mergeable.

brettz9:
Yes, I am just trying to draw attention to the fact that there are bugs here. If you aren't concerned enough, go ahead and close. I don't have the time myself to delve into the code; just wanted to flag what appear to be likely bugs rather than inadvertent stray backslashes.

marijnh:
Okay, closing—a lot of these are superfluous, but harmless.

PR #6050, opened 2019-11-03
[ CSS Hint pseudoClasses] Update pseudo Classes Variable 

HasanDelibas:
Added new CSS Pseudo Selector from https://www.w3schools.com/cssref/css_selectors.asp

### DIFF
--- a/addon/hint/css-hint.js
+++ b/addon/hint/css-hint.js
@@ -11,14 +11,14 @@
 })(function(CodeMirror) {
   "use strict";
 
-  var pseudoClasses = {"active":1, "after":1, "before":1, "checked":1, "default":1, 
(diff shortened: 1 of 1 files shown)


marijnh:
Hi. You can't do things like `not(selector)` here, since the pseudo-class isn't called `"not(selector)"`. That'll never match. To parse something like that you'll have to extend the parser to actually handle that syntax.

You could just remove the selectors with parentheses from this pull request in order to move forward with this, and leave the more complicated part for later.

HasanDelibas:
Removed parentheses in pseudoClasses object. Example "not(selector)" to **"not"**, "lang(language)" to **"lang"**

marijnh:
Thanks.

Issue #5931, opened 2019-07-09
phantomjs is an unmaintained project


ossdev07:
Hello Team,

PhantomJS is an unmaintained project and CodeMirror uses phantomjs for its qunit testing.

I am curious to know that will you be interested in replacing phantomjs with some other suitable alternative, as it is better using up to date dependencies.

Please share your thoughts on this and any further suggestion regarding the phantomjs replacement like chrome headless.

ossdev07:
Ping @marijnh 

I would like to introduce puppeteer as the testing framework in replacement to phantomjs,
As in order to achieve the same, I would have to update the test cases and other files having phantomjs support currently.
Will you be interested in accepting such a wide change in the package, could you please share your feedback on this.

marijnh:
What would you need to change about the tests themselves? There's some feature-detection happening where tests that don't run properly on Phantom are turned off, but I think that would be it (and can be ignored in a first iteration and removed later).

I'd be interested in a Puppeteer-based test runner. There's a bunch of tests that fail on current Chrome due to some details of the way it handles bidi text changing between the ancient version in PhantomJS and current versions, but I'm okay with just disabling those until someone has time to look into them.

marijnh:
Closed via #6051

PR #6051, opened 2019-11-04
Codemirror: Replaced PhantomJS Dependency with Puppeteer

ossdev07:
Phantomjs is an unmaintained project. Removed phantomjs dependency from package.json and run.js.
Added support for puppeteer to run the test suite on chrome.

Closes https://github.com/codemirror/CodeMirror/issues/5931


marijnh:
This was rather broken, in that the runner didn't notice failing scripts, and the patch, for some reason, inserted a broken and unneccesary new line in `test.js`, silently disabling all the tests in that file.

I've merged it as 94b08d6fc and fixed it in dde0e5cb51b

PR #6053, opened 2019-11-05
fix #5995

### LINKED ISSUE #5995: [Panel] onResize does not work with panel addon and height 100%
we have a div in our webapplication inside which we render codemirror with 100% height.

When not using panels the resize works fine but with a top panel we show for additional script information we face the following issue:

1. Assuming the browser window is so large that 50 lines fit in the editor without a scroll bar.
2. Resizing the window -> only 30 lines fit in the remaining space for codemirror

Expected behaviour: Codemirror should resize and show a scroll bar -> all 50 lines can be accessed

Actual behaviour: Codemirror does not resize, last 20 lines are cropped, no scrollbar appears and when you try scrolling to them the whole editor shifts inside its parent container, where it is stuck forever, even after resizing to original window size.

Additional things I observed:
The additional wrapper div that panel.js adds has no height restriction.
It will overflow the parent div (outermost div of codemirror) thus the height comparison doesn't trigger resize event.

I fiddled around with setting this wrapper to 100% height and codemirror to 100% and it made things better, but I couldn't get it to work reliably and exactly as it should.

The codemirror document in the screenshot actually contains 54 lines, but as you can see, you cannot scroll further and 18 lines are cropped of.

image

nkappler:
There were several issues in the resize logic of the panel addon which would only allow it to work with static sizes:

- the panel setSize function assumed that when newHeight parameter is null, only width had changed. However on window resize event the function will be called without parameters at all.
- the panel setSize function would overwrite the height of the wrapper. If it has a percent value it would be converted to pixels on the first run.
- Panels might change their height upon window resize, leading to miscalculation of the remaining height for the main editor, thus panels need to listen to resize events, too.

- in the version we are still using in our project CodeMirror would also cache the height of the container and would not trigger the setSize function, since the container is replaced by panel plugin and thus the wrong height is compared to the cache. However this cache seems to be removed already in the current version. (`onResize` in `global_Events.js`)

### DIFF
--- a/addon/display/panel.js
+++ b/addon/display/panel.js
@@ -54,6 +54,7 @@
     this.options = options;
     this.height = height;
     this.cleared = false;
+    window.addEventListener("resize", () => this.changed());
   }
 
   Panel.prototype.clear = function() {
@@ -68,7 +69,9 @@
   };
 
   Panel.prototype.changed = function(height) {
-    var newHeight = height == null ? this.node.offsetHeight : height;
+    if (!height && !this.node.parentElement) return; //this.node.offsetHeight would be undefined
+    var newHeight = !height ? this.node.offsetHeight : height;
+    if (newHeight == this.height) return;
     var info = this.cm.state.panels;
     this.cm._setSize(null, info.heightLeft -= (newHeight - this.height));
     this.height = newHeight;
@@ -91,7 +94,7 @@
 
     cm._setSize = cm.setSize;
     if (height != null) cm.setSize = function(width, newHeight) {
-      if (newHeight == null) return this._setSize(width, newHeight);
+      if (!newHeight) newHeight = info.wrapper.offsetHeight;
       info.setHeight = newHeight;
       if (typeof newHeight != "number") {
         var px = /^(\d+\.?\d*)px$/.exec(newHeight);
@@ -100,7 +103,6 @@
         } else {
           info.wrapper.style.height = newHeight;
           newHeight = info.wrapper.offsetHeight;
-          info.wrapper.style.height = "";
         }
       }
       cm._setSize(width, info.heightLeft += (newHeight - height));
--- a/addon/display/panel.js
+++ b/addon/display/panel.js
@@ -1,15 +1,15 @@
 // CodeMirror, copyright (c) by Marijn Haverbeke and others
 // Distributed under an MIT license: https://codemirror.net/LICENSE
 
-(function(mod) {
+(function (mod) {
   if (typeof exports == "object" && typeof module == "object") // CommonJS
     mod(require("../../lib/codemirror"));
   else if (typeof define == "function" && define.amd) // AMD
     define(["../../lib/codemirror"], mod);
   else // Plain browser env
     mod(CodeMirror);
-})(function(CodeMirror) {
-  CodeMirror.defineExtension("addPanel", function(node, options) {
+})(function (CodeMirror) {
+  CodeMirror.defineExtension("addPanel", function (node, options) {
     options = options || {};
 
     if (!this.state.panels) initPanels(this);
@@ -25,8 +25,7 @@
       wrapper.insertBefore(node, options.before.node);
     } else if (replace) {
       wrapper.insertBefore(node, options.replace.node);
-      info.panels++;
-      options.replace.clear();
+      options.replace.clear(true);
     } else if (options.position == "bottom") {
       wrapper.appendChild(node);
     } else if (options.position == "before-bottom") {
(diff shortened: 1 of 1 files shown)


addon/display/panel.js
```diff
@@ -54,6 +54,7 @@
     this.options = options;
     this.height = height;
     this.cleared = false;
+    window.addEventListener("resize", () => this.changed());
```
marijnh: This leaks an event listener. You could make the handler self-unregister when the editor is no longer in the DOM, but it might be cleaner to register a [`"refresh"` event](https://codemirror.net/doc/manual.html#event_refresh) handler, which the editor will call when it is resized or refreshed.

addon/display/panel.js
```diff
@@ -100,7 +103,6 @@
         } else {
           info.wrapper.style.height = newHeight;
           newHeight = info.wrapper.offsetHeight;
-          info.wrapper.style.height = "";
```
marijnh: Why did you remove this line?

addon/display/panel.js
nkappler: thanks for the hint, I asked myself how to clean this up already, but didn't think of CodeMirrors event system...

addon/display/panel.js
nkappler: This line removes the style on the wrapper element and I don't quite get why:

You can see it in the fiddle I linked in the bug report https://jsfiddle.net/xt64L72a/ (Set a breakpoint at this line and hit "run" )

Initially the wrapper has 90%, and it should always be 90%,
however when this line is executed, the 90% are removed, thus the wrapper sets it height to the height of its contents (panel and editor) and will stay at that height forever.

addon/display/panel.js
nkappler: Plugins are not getting compiled, are they?
Using an arrow function is probably a bad idea then, right?

addon/display/panel.js
marijnh: The code is trying to convert a percentage height to a pixel height by first setting it, measuring the DOM, and then removing the style again. So clearing that is intentional.

nkappler: I understand the intention, but it seems to be executed on the wrong DOM element.
If the container already has a percentage value intentionally, it will be removed and is lost.

From my understanding the line where it is set first is not needed either since it should be set already, but I ran into issues during testing so I left it...

nkappler: Ok I do understand that setting it is required when calling setSize with arguments,
However it should not be removed.

Imagine you call setSize(500, "50%"):
The CodeMirror Container will be set to 50%,
this value will be converted to pixels and shoved into _setSize to resize the actual CoreMirror.

Then the container height (50%) will be removed. The Container won't change it's size since its children make up exactly those 50% it is supposed to be, but it will stay this big forever.
The Height of the container has been changed from dynamic 50% to whatever many pixels were 50% at the time the function was called...

marijnh:
So, frankly, this module is a terrible hack. But since existing users might be relying on it, I rather leave it as it is than making the hacks even more convoluted and possibly breaking them in new ways. I don't remember what I was thinking when I first wrote some of this code, but the very concept seems deeply flawed, so I don't think it can be saved.

I recommend you either fork the addon locally or try to replace it with a simpler, less broken approach that serves your use case. Would that work for you?

nkappler:
Well, after digging through this for two days, I think I understand it well enough to say that actually nothing should break.
If you used absolute sizes before this change, it will still work after the change.
If you used percentage values before, it didn't work and now will, and if somebody got it to work, than it was most certainly some kind of event handler that calculates the new sizes manually and calls setSize(). This approach would be unnecessary now but would also still work...

After all, it is your decision and since we don't import CodeMirror via npm, I already fixed it for our product.
My intention was to improve CodeMirror/the panel addon for everyone... (and also we wouldn't have to maintain this change when we update someday 😄 )

marijnh:
Hm, all right, let's iterate a bit more. If the built-in window resize handler already calls the `setSize` method, why do you need to explicitly hook it up through an additional handler? It seems the current PR will always end up calling `setSize` twice for resize events—once through the built-in resize handler, and once via your handler and `changed`.

nkappler:
Yes that's right (at least sometimes). The panels themselves might change their size when the window resize is triggered, either because they have relative sizes themselves or due to line wrapping in their content.
However this is ignored by the current implementation. When `cm.state.panels.heightLeft` is recalculated inside `setSize`, the height of the panel(s) is ignored, assuming they didn't change.
Hence I added the handler inside the panel, which checks if the height has changed and only if it has (the check in line 74 `if (newHeight == this.height) return;` ) `_setSize` will be called again to correct the wrong height from the first call.

This means that with N panels, `_setSize` will be called at least once and up to N+1 times, since the call is triggered by the panel itself.

A better approach would be of course to recalculate the height based on all panels each time in `setSize`.
This would make `cm.state.panels.heightLeft` completely obsolete, resulting in basically a rewrite of the whole module, which I tried to avoid...

marijnh:
This does sound like the overwriting `setSize` method should just always do the measuring, so that the `resize` handler can be dropped.

nkappler:
> 
> 
> This does sound like the overwriting `setSize` method should just always do the measuring, so that the `resize` handler can be dropped.

Done.

marijnh:
Thanks, looks a lot more coherent now.

PR #6075, opened 2019-11-24
Show item count in fold widget for JSON

develohpanda:
## Feature
There is a [feature request](https://github.com/getinsomnia/insomnia/issues/1799) in a consuming project to show the JSON array size when a node has been folded.

This seems like a reasonable request, given the user currently copy-pastes into a separate window to see the count. I looked for an existing issue/request but couldn't find one. Instead of creating a request, I have attempted an implementation after understanding some of the structure.

## Feedback requested
- I tried to decouple the `JSON.Parse` -> `Object.keys` from the `fold.brace` helper, however I still need all the logic within `fold.brace`. If I decouple, I would still have to load the `fold.brace` helper, return extra data (the brace type), to then run with the counting logic. Thoughts?
- Given the existing `widget` can be overridden, I decided to allow the `jsonCountWidget` to also be overridden. It is up to the user how they want to show it. It is `undefined` by default, meaning the behavior is switched off. Thoughts on configuration?

## Demo

Code folding demo within CodeMirror:

Gif from the consuming application:

Behavior with invalid JSON (reverts to `widget`):

## Final words
I loved the fact that there is no need to spin up a web server. Very vanilla, very nice. 🍦 

Do you have any ideas why these tests fail on my W10 machine? They fail on master, but pass in CI. I am using Node v8.11.1 locally.

image

Open to suggestions and feedback 😄

### DIFF
--- a/addon/fold/brace-fold.js
+++ b/addon/fold/brace-fold.js
@@ -70,7 +70,7 @@ CodeMirror.registerHelper("fold", "brace", function(cm, start) {
     } catch(e) { }
   }
 
-  return {from: from, to: to, count: count};
+  return {from, to, count};
 });
 
 CodeMirror.registerHelper("fold", "import", function(cm, start) {
--- a/addon/fold/brace-fold.js
+++ b/addon/fold/brace-fold.js
@@ -59,7 +59,7 @@ CodeMirror.registerHelper("fold", "brace", function(cm, start) {
   var from = CodeMirror.Pos(line, startCh), to = CodeMirror.Pos(end, endCh);
   var count = undefined;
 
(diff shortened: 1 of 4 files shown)


marijnh:
I'm not going to merge this. It introduces special-case code in places that are supposed to be general tools (there no reason for the code-folding addon to know about JSON widgets), and generally looks too ad-hoc.

You could, as an alternative, add a feature to the foldcode addon to allow `widget` to be a function that create the widget for a given range, and pass that the range's start and end. That should, I think, allow you to create this effect without changing a lot of CodeMirror library code.

develohpanda:
Noted, thanks for your feedback! 👍

PR #6081, opened 2019-12-06
Fix syntax highlight for pgsql mode for backslash and escape constant

adityatoshniwal:
Fix syntax highlight for pgsql mode as PostgreSQL does not require backslash to be escaped.
Add support for escape constant as mentioned with ref to:
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE

Below are the sample queries to test:
SELECT 'select c:\data\out\'||oid||'.csv' from pg_class;
select E'\'', '123\';

Attached are the screenshots of working example where the above queries works fine - pgAdmin4.
Screenshot 2019-12-06 at 12 54 36
Screenshot 2019-12-06 at 12 53 51

### DIFF
--- a/mode/sql/sql.js
+++ b/mode/sql/sql.js
@@ -60,14 +60,16 @@ CodeMirror.defineMode("sql", function(config, parserConfig) {
       state.tokenize = tokenLiteral(ch);
       return state.tokenize(stream, state);
     } else if ((((support.nCharCast && (ch == "n" || ch == "N"))
-        || (support.escapeConstant && (ch == "e" || ch == "E"))
         || (support.charsetCast && ch == "_" && stream.match(/[a-z][a-z0-9]*/i)))
-        && (stream.peek() == "'" || (stream.peek() == '"' && support.doubleQuote)))) {
+        && (stream.peek() == "'" || stream.peek() == '"'))) {
       // charset casting: _utf8'str', N'str', n'str'
       // ref: http://dev.mysql.com/doc/refman/5.5/en/string-literals.html
+      return "keyword";
+    } else if (support.escapeConstant && (ch == "e" || ch == "E")
+        && (stream.peek() == "'" || (stream.peek() == '"' && support.doubleQuote))) {
       // escape constant: E'str', e'str'
       // ref: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE
-      state.escapeConstSet = true;
+      state.tokenize = tokenLiteral(stream.peek(), true);
       return "keyword";
     } else if (support.commentSlashSlash && ch == "/" && stream.eat("/")) {
       // 1-line comment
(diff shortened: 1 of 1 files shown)


marijnh:
It looks like you are keeping mode state in a variable, if I understand the role of `escapeConstSet` correctly. That doesn't work—when the more is resumed from some random point in the code, that state will still be leftover from the last time it ran, and things will break.

adityatoshniwal:
Urgh.. Right. Do you have any suggestions on how I can improve it ?

marijnh:
You could put it in the `state` object instead.

adityatoshniwal:
OK, thanks. I'll go through the code once again.

adityatoshniwal:
@marijnh I've updated the code. Kindly review.

mode/sql/sql.js
```diff
@@ -60,10 +60,14 @@ CodeMirror.defineMode("sql", function(config, parserConfig) {
...
     } else if ((((support.nCharCast && (ch == "n" || ch == "N"))
+        || (support.escapeConstant && (ch == "e" || ch == "E"))
         || (support.charsetCast && ch == "_" && stream.match(/[a-z][a-z0-9]*/i)))
-        && (stream.peek() == "'" || stream.peek() == '"'))) {
+        && (stream.peek() == "'" || (stream.peek() == '"' && support.doubleQuote)))) {
       // charset casting: _utf8'str', N'str', n'str'
       // ref: http://dev.mysql.com/doc/refman/5.5/en/string-literals.html
+      // escape constant: E'str', e'str'
+      // ref: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE
+      state.escapeConstSet = true;
```
marijnh: Two more questions:

 - Is it intentional that this is also being set for other types of string prefixes like `n` and `_`? It looks like it should only be relevant for `e`.

 - You don't actually need a new state field, since the state already has a `tokenize` callback that you could replace with a tokenizer function that can parse escaped strings. You could add another argument to `tokenLiteral` that determines backslash behavior, and set `tokenize` to a function that consumes the quote and then replaces itself with `tokenLiteral(quote, ...)` here.

adityatoshniwal: > Two more questions:
> 
> * Is it intentional that this is also being set for other types of string prefixes like `n` and `_`? It looks like it should only be relevant for `e`.
Oh yes, you're right. Will fix this. :/
> * You don't actually need a new state field, since the state already has a `tokenize` callback that you could replace with a tokenizer function that can parse escaped strings. You could add another argument to `tokenLiteral` that determines backslash behavior, and set `tokenize` to a function that consumes the quote and then replaces itself with `tokenLiteral(quote, ...)` here.
I didn't want to repeat the logic of tokenLiteral and wanted to reuse it. I'll try if can avoid the new state field.

Thanks for your time.

marijnh: > I didn't want to repeat the logic of tokenLiteral and wanted to reuse it.

That's what I mean—if you give it an extra argument, you can reuse it.

adityatoshniwal: Cool. Let me try it !! Thanks.

adityatoshniwal: Hi,
I tried to do it without escapeConstSet var, but the problem is, it is before the string stream starts. So I have to tell the tokenLiteral function somehow that the previous char was a escape constant. I can do it in escape constant check, but then I lose the formatting for escape constatnt.

marijnh: After seeing the `e` prefix, you can set `state.tokenize` to a tokenizer function that's specific to that situation, which starts tokenizing an escape-supporting string when it is called.

adityatoshniwal: Hi @marijnh,
This looks a lot better now. :)
Sorry for trouble.

marijnh:
Thanks for implementing this. I've slightly adjusted it further in attached patch, so that the code won't try to skip quotes at the start of continued string lines.

adityatoshniwal:
Thanks a lot :)

adityatoshniwal:
Oh wait.. I had marked backslashes as false in mode settings. Will check again.

PR #6082, opened 2019-12-06
Allow dynamic fold widget via a function

develohpanda:
Hello!

As per your suggestion in #6075, I updated the cold folding `widget` to accept a function with the folding range, to keep the changes within the core CodeMirror library to a minimum.

I have added an example usage for what a custom widget via a function may be used for, but please let me know if you would like me to remove the example in `folding.html`.

image


marijnh:
Thanks, that looks great now. Merged as 3ace5da2

PR #6083, opened 2019-12-08
Add indentSelection doc

mtgto:
indentSelecton is missing in manual.
https://codemirror.net/doc/manual.html#indentLine
https://github.com/codemirror/CodeMirror/blob/5.49.2/src/edit/methods.js#L97-L116

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -2131,6 +2131,10 @@ editor.setOption("extraKeys", {
           <dd>Add (positive number) or reduce (negative number) the
           indentation by the given amount of spaces.</dd>
         </dl></dd>
+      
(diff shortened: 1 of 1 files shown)


marijnh:
I think the idea is that that method is not public, since you can call it in all useful ways through the `indentAuto`, `indentMore`, and `indentLess` commands. What were you trying to do?

mtgto:
@marijnh Thanks for your reply.
I create this PR because our project already uses indentSelection, but I found there is no document.
It is fine for me to exec indentAuto, indentMore, and indentLess.

sq-dr-art2:
😀

PR #6093, opened 2019-12-22
Add sublime shortcut Ctrl-K Ctrl-D: find_under_expand_skip

axellew:
Add the command and shortcut Ctrl-K Ctrl-D which works as Ctrl-D but removes the last selection and selects the next instead.

In the string, with the first foo selected (bold is selected):
"**foo** bar, foo foobar"
Pressing the new command will select the second and remove the first. 
"foo bar, **foo** foobar"


marijnh:
Thanks!

PR #6094, opened 2019-12-22
Soy: Add support for default param value

axellew:
Add support to soy for default values.
{@param foo: string = 'foo')
{@param bar:= 'bar'}
Also removed {@state?} since it has never been a valid statement.


marijnh:
Thanks. Could you try to clean up your commit history before creating a pull request next time?

axellew:
Yes, will do. I haven't worked enough with git, need to read up on fix the commit history, or how to only pick some of the commits for a pull request.

PR #6099, opened 2019-12-30
Fix: Broken line widget removal

fellmann:
Removal of line widgets is broken in 5.50.0 due to invalid classname check.


marijnh:
Thanks!

Issue #6098, opened 2019-12-27
allowDropFileTypes issue


pakhomchik:
Drag and Drop ignores valid files if there are invalid file types in the draggable list when we limit types with `allowDropFileTypes` option.
<details>
        <summary>GIF</summary>
<img src="https://user-images.githubusercontent.com/668493/71504961-5d23af00-28b6-11ea-99e4-0b1a2dda5cdb.gif">
</details>

Example ([jsfiddle](https://jsfiddle.net/podvzbzdnul/kodbt0z6/)):

1.  Add textarea:
```html
<textarea id="textarea"></textarea>
```
2. Create a CodeMirror instance with `allowDropFileTypes` option:
```js
let options = {
  allowDropFileTypes: ['text/css']
}
CodeMirror.fromTextArea(document.getElementById('textarea'), options);
```
3. Prepare three files: `image.png` (actual image), `valid.css` and `invalid.css` (binary file).
4. - Drop the `valid.css` - you'll see its content in editor.
   - Drop the `valid.css` and `invalid.css` - you'll see the `valid.css` content.
   - [**issue**] Drop the `valid.css` and `image.png` - you'll see nothing.

Expected behaviour: we should see the `valid.css` content in editor.

---
OS: macOS Catalina 10.15.2
Browser: Chrome 79.0.3945.88
CodeMirror: 5.50.0

axellew:
Added a pull request for a fix for the issue: https://github.com/codemirror/CodeMirror/pull/6102
The current logic aborts the whole operation if there is any error (expect bad content) in any of the files. The pull request tries to fix this by filtering out any files with errors and then pasting the valid files.

PR #6102, opened 2019-12-31
If valid and invalid files are drag n dropped, paste all valid files

axellew:
Fix for issue [#6098](https://github.com/codemirror/CodeMirror/issues/6098). When multiple files are dropped into the editor and any of these are invalid, then none of the files content is pasted.

With this patch all valid files will be pasted and all invalid rejected.

### DIFF
--- a/src/edit/drop_events.js
+++ b/src/edit/drop_events.js
@@ -28,20 +28,21 @@ export function onDrop(e) {
   // Might be a file drop, in which case we simply extract the text
   // and insert it.
   if (files && files.length && window.FileReader && window.File) {
-    let readFiles = 0, text = []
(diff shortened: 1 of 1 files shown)


marijnh:
This library still targets environments that don't have `Promise`, so this'd have to be written in a way that doesn't depend on those.

axellew:
Too bad, I'll see what can be done. I found some other usages of promises, so I thought it was safe to use. I see now that those either have checks or are only used in tests.

marijnh:
Thanks! Another issue is that this makes the order in which file content is inserted nondeterministic. I'd prefer to stick to the order of the `files` array. Other than that, this looks good now.

marijnh:
Thanks!

PR #6106, opened 2020-01-07
Add solidity mode

elpnt:


jdbruxelles:
Hi. New modes are not accepted into the main distribution anymore. The recommended way to distribute code like this is to put it into its own npm package with a name like codemirror-mode-solidity.

PR #6113, opened 2020-01-13
Add Elixir mode

sabudaye:
Hello.

First of all, thank you for making this great thing! I was using it for years and didn't even know about it :)

Bitbucket Server uses CodeMirror (version 49 in latest available release) to highlight syntax of source files. As an Elixir developer and Bitbucket Server user I wasn't really happy that Bitbucket can't do syntax highlighting for Elixir, after years of waiting I finally decided to do something about it :)

CodeMirror has built-in modes for almost every language but not for Elixir. There is mode from @ianwalter https://github.com/ianwalter/codemirror-mode-elixir but apparently it can't be used in Bitbucket since it's not a part of CodeMirror repo. It also has Vue as a dependency, which is unnecessary in my opinion. This mode also doesn't support few syntax features of Elixir language like sigils, maps, structs.

So I made new Elixir mode based on Elm and Haskell modes, it's probably not perfect but it works. I'd welcome any kind of feedback.

### DIFF
--- a/mode/elixir/elixir.js
+++ b/mode/elixir/elixir.js
@@ -1,0 +1,262 @@
+// CodeMirror, copyright (c) by Marijn Haverbeke and others
+// Distributed under an MIT license: http://codemirror.net/LICENSE
+
+(function(mod) {
+    if (typeof exports == "object" && typeof module == "object") // CommonJS
+      mod(require("../../lib/codemirror"));
+    else if (typeof define == "function" && define.amd) // AMD
+      define(["../../lib/codemirror"], mod);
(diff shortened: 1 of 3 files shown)


jdbruxelles:
Hi, thank you for your interest but unfortunately, new modes are not accepted into the main distribution anymore. The recommended way to distribute code like this is to put it into its own npm package with a name like codemirror-mode-elixir.

ianwalter:
Vue is only a dev dependency. Also PRs to codemirror-mode-elixir are welcome.

PR #6114, opened 2020-01-14
Add Elixir support as Clike mode

sabudaye:
Hello,

My previous attempt to add Elixir support wasn't accepted #6113
But I didn't give up :) 

I saw there were few merged PRs for a new languages added as an extension for existing modes, so I did the same.

Hope it will work this time.

### DIFF
--- a/mode/clike/clike.js
+++ b/mode/clike/clike.js
@@ -932,4 +932,131 @@ CodeMirror.defineMode("clike", function(config, parserConfig) {
     }
   });
 
+  function tokenElixirString(tripleString){
+    return function (stream, state) {
+      var escaped = false, next, end = false;
+      while (!stream.eol()) {
+        if (!tripleString && !escaped && stream.match('"') ) {end = true; break;}
+        if (tripleString && stream.match('"""')) {end = true; break;}
+        next = stream.next();
+        if(!escaped && next == "#" && stream.match('{'))
+          stream.skipTo("}");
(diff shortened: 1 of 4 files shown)


marijnh:
Sorry, no. I realize that there are advantages for you to having this in the core, but this is a significant chunk of code covering a mode that already exists as an external package, and that's exactly the kind of thing we _don't_ want to add to this codebase.

sabudaye:
@marijnh 
I do understand that it's very hard to maintain so many languages and idea of having language modes/grammars outside of core repo makes perfect sense, like it is done in Github's [linguist](https://github.com/github/linguist). Especially in a situation when you are actively working on complete new [version](https://github.com/codemirror/codemirror.next).

But at the same time you are forcing users (like Bitbucket) of your current actively used library to think about languages unsupported by CodeMirror in some special way. They have to add additional dependencies, which they are not going to do (I have official answer from Atlassian). Why they should? Instead of of having one dependency which manages all syntax-highlighting stuff, they need to maintain few more. Why not to change the library to some another without this issue? That's what Gerrit and Gogs did by moving away from CodeMirror to highlight.js.

While there is nothing wrong with code-mirror-elixir, contributing to this mode is not going to resolve my (and many other's) issue that Bitbucket doesn't support Elixir. But it can be resolved by merging this or previous PR.

Can we just brake this circle and help people actually?

PR #6125, opened 2020-01-24
Add theme class name to lint tooltips

ghik:
Currently it's impossible to style lint tooltips in a theme CSS file because tooltip elements are inserted directly into `body` element.

This PR changes `lint.js` addon so that lint elements have theme class name added.


marijnh:
Good idea!

PR #6135, opened 2020-02-04
Append the tooltip to the wrapper element

tejasrivastav:
The tooltip in the existing way append to the body.
When applied the codemirror context inside a shadow root (custom element) this calls append the tooltip out of the root components where the behavior of the tooltip is broken.

Forcing, the user has to monkey patch the appendchild method to override the behavior or modify the code.

This change would append the tooltip to render in its own context.

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -16,7 +16,10 @@
     var tt = document.createElement("div");
     tt.className = "CodeMirror-lint-tooltip cm-s-" + cm.options.theme;
     tt.appendChild(content.cloneNode(true));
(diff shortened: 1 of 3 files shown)


tejasrivastav:
@marijnh if this makes sense, require similar changes in the `tern.js` addon as well

marijnh:
This would have been the right approach from the start (and is how tooltips will work in CodeMirror 6), but unfortunately just changing it now is without a doubt going to break someone's CSS. Could you make this opt-in through an option (to the lint addon)?

marijnh:
Thanks! Merged as 1fed46d79957d924

PR #6137, opened 2020-02-06
Fix test driver

axellew:
Remove hard coded filter which only allowed vim tests to run.

Also improved the unique naming algorithm's time complexity from O(n^3) to O(n). With these 2000 tests this will reduce the iterations needed from 2 000 000 to 2000 iterations (or if all tests had the same name 4 * 10^9 iterations).

### DIFF
--- a/test/driver.js
+++ b/test/driver.js
@@ -1,4 +1,4 @@
-var tests = [], filters = [], allNames = [];
+var tests = [], filters = [], nameCounts = {};
 
 function Failure(why) {this.message = why;}
 Failure.prototype.toString = function() { return this.message; };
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Seems like that check had been there for almost a month since 32da49cfe4389c4 (ahem, @nightwing , though I should have caught it in review as well).

I assume the performance improvement is purely theoretical? An O(n^2) array search where n is a few thousand seems like it'd have a negligible impact compared to the cost of running even a single test. But sure, the new code does look nicer.

axellew:
Yes this purely theoretically, this shouldn't affect the running time that much. When coding java, 10^8 iterations is about 1 sec of computational time on a normal desktop. So when there are 10k tests it will add about 1 second in total. (If all tests had the same name it would already add about 1 minute extra time to run the tests, though).

marijnh:
I think your estimates are a bit pessimistic there—a small node script that does something like the old loop + `indexOf` check for 10k items runs in 100ms, and I'd expect Java to be even faster.

axellew:
Yes, all of it just estimations and the number of tests will never be large enough for it be a problem. I just had an itch to fix it :)

PR #6140, opened 2020-02-07
[Soy] Improve map, list, record and list comprehension highlighting

axellew:
Add support for map, list and records.

Example of code which is now supported:
  {@param name: [test: number]}
  {@param numbers: list<number>}
  {@param strToNumMap: map<string, number>}
  {let $list: [['a'], ['b']] /}
  {let $list2: [$a + 1 for $a in $name if $a >= 3] /}
  {let $record: record(test: 'a', foo: 'bar') /}
  {let $map: map('test': 'a', 4: 'bar') /}

### DIFF
--- a/mode/soy/soy.js
+++ b/mode/soy/soy.js
@@ -156,7 +156,7 @@
       } else if (match = stream.match(/^\$([\w]+)/)) {
         return ref(state.variables, match[1], !state.lookupVariables);
       } else if (match = stream.match(/^\w+/)) {
-        return /^(?:as|and|or|not|in|for|if)$/.test(match[0]) ? "keyword" : null;
+        return /^(?:as|and|or|not|in|if)$/.test(match[0]) ? "keyword" : null;
       }
 
       stream.next();
@@ -375,6 +375,13 @@
               state.soyState.pop();
               state.lookupVariables = true;
               return null;
+            }
+            if (stream.match(/for/)) {
+              state.soyState.push("var-def")
+              return "keyword";
+            } else if (stream.match(/in/)) {
+              state.lookupVariables = true;
+              return "keyword";
             }
             return expression(stream, state);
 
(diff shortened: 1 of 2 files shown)


mode/soy/soy.js
```diff
@@ -125,6 +125,44 @@
       this.scope = scope;
     }
 
+    function expression(stream, state) {
+      let match;
+      if (stream.match(/[[]/)) {
+        state.soyState.push("list-literal");
+        state.lookupVariables = false;
+        return null;
+      } else if (stream.match(/map/)) {
```
marijnh: I suspect you'll need at least a `\b` at the end of this regular expression, to avoid matching any word that starts with "map". (Same for other word regexps.)

mode/soy/soy.js
```diff
@@ -125,6 +125,44 @@
...
+        state.soyState.push("map-literal");
+        return "keyword";
+      } else if (stream.match(/record/)) {
+        state.soyState.push("record-literal");
+        return "keyword";
+      } else if (stream.match(/([\w]+)(?=\()/)) {
+        return "variable callee";
+      } else if (match = stream.match(/^["']/)) {
+        state.soyState.push("string");
+        state.quoteKind = match;
```
marijnh: This issue already existed before this patch, but it seems that the code that actually uses `state.quoteKind` compares it to a string, which only works because of implicit array-to-string conversion. Might be cleaner to change this to `= match[0]`

mode/soy/soy.js
axellew: Yes, makes sense. Added it to all places I could find.

mode/soy/soy.js
axellew: Yep, fixed! Javascript is a bit too forgiving sometimes...

marijnh:
Thanks!

PR #6155, opened 2020-02-22
Markdown: Disallow single dash for a setext header, fixes #6129

### LINKED ISSUE #6129: Feature Request: Option to disable setext headers in markdown mode
I know that setext headers are part of the [commonmark spec](https://spec.commonmark.org/0.28/#setext-headings), but our users find them unexpected and confusing (particularly when only a single character is used), in part because we style `cm-header-2` to match the font size it will be rendered at
image

Would it be possible to add an option to disable setext headers to the markdown modeConfig in the same way other markdown features can be enabled/disabled like `strikethrough`

Alternatively, the spec says 
> If a line containing a single `-` can be interpreted as an empty list items, it should be interpreted this way and **not** as a setext heading underline.

which if implemented would largely solve our issue as well. 

Thanks very much 👍

Zverik:
Reasons expressed by Michael in #6129 seem good enough, and a dash can be used for multiple purposes. I believe many people start writing lists right after an introductory sentence, and may be puzzled at why the previous line suddently becomes a header for a second.


marijnh:
Thanks!

PR #6156, opened 2020-02-22
Fix lint warnings

Zverik:
What it says on the tin. There were 4 lint warnings, I have fixed them.


marijnh:
Thanks! I was sure our CI would complain about these. Going to go and see why it wasn't doing that.

marijnh:
Ah, that was broken in 94b08d6fc . Pushing a patch in a second.

PR #6179, opened 2020-03-17
Disable autocorrect for the vim prompt

alur:


marijnh:
(You have a trailing space in there—could you drop that to make the linter happy?)

alur:
Dropped it

marijnh:
Thank you!

PR #6185, opened 2020-03-23
Fixes #4804 dialog closes when tab loses focus

### LINKED ISSUE #4804: Leave vim's command bar open even after screen losing focus
1. Open https://codemirror.net/demo/vim.html
2. Type ":" and notice the command bar pop up
3. Switch windows or tabs, and notice that the command bar disappears

Expected:
The command bar should remain active even after switching windows. For example often if i'm doing :%s/foo/bar, I will switch windows to figure out what i want foo and bar to be - it's super annoying when it disappears so that I have to start all over with my vim command.

benhormann:
The input field of a dialog loses focus when switching to another tab or dev tools. This fix checks to see if the new focus target is inside the window (i.e. not null) before closing the dialog.

I have also changed the target to be the dialog div, instead of the input field. This better suits sites that add other elements inside the dialog.

Tested and working in Firefox, Chrome, Safari and IE > 8.
IE 8 retains previous behaviour, since `relatedTarget` is undefined.

### DIFF
--- a/addon/dialog/dialog.js
+++ b/addon/dialog/dialog.js
@@ -82,7 +82,9 @@
         if (e.keyCode == 13) callback(inp.value, e);
       });
 
-      if (options.closeOnBlur !== false) CodeMirror.on(inp, "blur", close);
+      if (options.closeOnBlur !== false) CodeMirror.on(dialog, "focusout", function (evt) {
+        if (evt.relatedTarget !== null) close();
+      });
     } else if (button = dialog.getElementsByTagName("button")[0]) {
       CodeMirror.on(button, "click", function() {
         close();


addon/dialog/dialog.js
```diff
@@ -82,7 +82,9 @@
         if (e.keyCode == 13) callback(inp.value, e);
       });
 
-      if (options.closeOnBlur !== false) CodeMirror.on(inp, "blur", close);
+      if (options.closeOnBlur !== false) CodeMirror.on(dialog, "focusout", function (evt) {
+        if (evt.relatedTarget !== null) close();
+      });
     } else if (button = dialog.getElementsByTagName("button")[0]) {
```
jdbruxelles: While you are at it, can you correct this assignment to make it a comparison:
```javascript
} else if (button = dialog.getElementsByTagName("button")[0]) {
```
to 
```javascript
} else if (button == dialog.getElementsByTagName("button")[0]) {
```

marijnh: No, that would not be a good idea, since it would break the code.

marijnh:
Thanks!

addon/dialog/dialog.js
jdbruxelles: Great ! I see, thanks.

PR #6186, opened 2020-03-24
Fix issue with backslashes being highlighted as errors in Powershell code

mcasperson:
Today backslashes are highlighted with the `cm-error` class in Powershell code:

image

This PR treats a backslash as punctuation, removing the error highlight:

image


marijnh:
Thanks!

Issue #5844, opened 2019-04-05
Cursor Jumps around on Mobile


Hawkers:
**Describe the bug**
The cursor does not always remain where the user intends.

**To Reproduce**

1. Open Chrome mobile
2. Type something at the end of the document
3. Tap after a word higher in the document and press space
4. Cursor will jump to end of document

**Expected behavior**
The cursor should remain in place

** Version information **
OS: Android
Browser: Chrome 73

Ge-lx:
* Android 9 on OnePlus 5T (A5010)
* Chrome `80.0.3987.149`
* CodeMirror `v2.8.0` via EasyMDE `v2.8.0`

I am also experiencing this bug and it makes for a very bad user experience. From my testing it seems that the editor is not always correctly setting the internal cursor position position (`setSelection(doc, sel, options)`) when the user taps somewhere in the document.
It works fine, when using the cursor key (via remote debugging) or a mouse to interact with the document. While debugging this I saw, that CodeMirror is listening for `touchend` events to update the cursor position in response to user input on mobile.

The problem is, that sometimes the `touchend` event is not fired.
You should be able to see for yourself by registering event listeners like:
```javascript
window.addEventListener('touchend', () => console.log('touchend'))
window.addEventListener('touchstart', () => console.log('touchstart'))
```
Some googling led me to this [unfixed Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=464579), which could be causing this issue: When the target element of a `touchstart` is removed from the DOM while the touch is in progress, no `touchend` will be fired.

The problem would then be, that CodeMirror removes elements from the DOM, that have been the target of `touchstart` events, so that the corresponding `touchend` never fires. 

As far as I can tell this is caused by `updateDisplayIfNeeded`, which is called on `touchstart` events.   A quick and dirty bugfix for this is the following:

`src/display/update_display.js:90`
```javascript
function updateDisplayIfNeeded(cm, update) {
  var display = cm.display, doc = cm.cod;
  
  ...
  ...

  // <+++ line=97>
  // Bail out if there is a touch event in progress.
  if (!update.force && cm.activeTouch !== null) {
    return false;
  }
  // </+++>

  ...
  ...
}
```

In my limited testing I could identify any problems with this approach, but I do not have the knowledge of CodeMirror to test it in every scenario.
In my deployment I will however deploy this fix, as without it the editor is not very usable on Android.

marijnh:
[...]

PR #6193, opened 2020-03-27
Fix caret position updating with touch events in chrome

Ge-lx:
There has been an [issue in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=418177) for quite a while (> 5 years) now, which breaks changing the caret position via touch events. 

I have provided more details about the bug in #5844.

While the fix is not the most elegant - I have to admit - it should not break anything else since it only interferes with Chrome environments where there is an `.activeTouch`. It did not break any tests for me.

### DIFF
--- a/src/display/update_display.js
+++ b/src/display/update_display.js
@@ -2,7 +2,7 @@ import { sawCollapsedSpans } from "../line/saw_special_spans.js"
 import { heightAtLine, visualLineEndNo, visualLineNo } from "../line/spans.js"
 import { getLine, lineNumberFor } from "../line/utils_line.js"
 import { displayHeight, displayWidth, getDimensions, paddingVert, scrollGap } from "../measurement/position_measurement.js"
-import { mac, webkit } from "../util/browser.js"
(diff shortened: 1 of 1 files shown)


marijnh:
I'm a bit skeptical about this not breaking anything—it looks like it will freeze the display as long as a touch event is active, which could, if you make, for example, a programmatic change while a touch is active, leave the editor displaying the wrong state.

Have you looked into kludges that focus on clearing `activeTouch` more proactively? For example, store the touch target in that object, and when it isn't part of the editor anymore after a display update, clear `activeTouch`?

Ge-lx:
How would that fix the issue of the missing caret position update?

I also just noticed that I referenced the wrong chromium bug. I think [this one](https://bugs.chromium.org/p/chromium/issues/detail?id=464579) is more appropriate and explains the underlying issue a bit better.

Sorry, I don't have the resources to dig into the code that much, so I cannot offer any help on more sophisticated mitigation methods. My hope is that the browser bug gets addressed and we receive at least a 'touchcancel' event to which we can react.

PR #6195, opened 2020-03-28
Update runMode with changes from runmode.js

BrianHung:
This PR updates the `runMode` function in runmode-standalone.js with 3 commits from [runmode.js](https://github.com/codemirror/CodeMirror/blob/master/addon/runmode/runmode.js):
- https://github.com/codemirror/CodeMirror/commit/7e35f03ad1639d92735c8dee1e1b5c86a727c873
- https://github.com/codemirror/CodeMirror/commit/53bc4b1f54670c827662d5210a37b35e63a07c39
- https://github.com/codemirror/CodeMirror/commit/64113ec97bb73c2a740d42d56021acb01ce05a92

The first commit from runmode.js allows the `callback` in runMode to be a function, as mentioned in https://codemirror.net/demo/runmode.html and https://discuss.codemirror.net/t/runmode-standalone-callback/2357. The last two commits fixes a copy-and-paste bug from IE.

### DIFF
--- a/addon/runmode/runmode-standalone.js
+++ b/addon/runmode/runmode-standalone.js
@@ -104,14 +104,18 @@ CodeMirror.defineMIME("text/plain", "null");
 
 CodeMirror.runMode = function (string, modespec, callback, options) {
   var mode = CodeMirror.getMode({ indentUnit: 2 }, modespec);
+  var ie = /MSIE \d/.test(navigator.userAgent);
+  var ie_lt9 = ie && (document.documentMode == null || document.documentMode < 9);
 
-  if (callback.nodeType == 1) {
+  if (callback.appendChild) {
     var tabSize = (options && options.tabSize) || 4;
     var node = callback, col = 0;
     node.innerHTML = "";
     callback = function (text, style) {
       if (text == "\n") {
-        node.appendChild(document.createElement("br"));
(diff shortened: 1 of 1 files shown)


marijnh:
Looks like `callback` could already be a function (functions won't have a `nodeType` property) even before these patches.

But merging them seems harmless enough.

BrianHung:
That makes sense 😄. The minor issue was in fact an error on my part: I passed in an undefined function as the callback (forgot Chrome console wipes functions per session). I rushed into thinking it was a problem with `nodeType` since the [`runmode.node`](https://github.com/codemirror/CodeMirror/blob/629f42d0d52a05b8d493f045857a7d6c052c4f0e/addon/runmode/runmode.node.js#L180), which doesn't take in a DOM element, throws a `Uncaught TypeError: callback is not a function` instead of a `Uncaught TypeError: Cannot read property 'nodeType' of undefined`.

That leads to another realization: should `runMode` in `runmode.node` also be able to take in as a callback parameter a DOM element? There's no `appendChild` or `nodeType`, so I'm guessing it assumes callback is a function.

marijnh:
The node version will usually be ran without even having access to a DOM implementation, so it didn't seem useful to include the DOM-appending functionality there.

PR #6197, opened 2020-03-31
Fix a accessibility issue to allow screenreader to read a label when …

adityatoshniwal:
…CodeMirror focussed.
Added a CodeMirror option srLabel to allow user to set the label for screenreaders.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -421,7 +421,7 @@
       simply <code>true</code>), focusing of the editor is also
       disallowed.</dd>
 
-      <dt id="option_srLabel"><code><strong>srLabel</strong>: string</code></dt>
+      <dt id="option_screenReaderLabel"><code><strong>screenReaderLabel</strong>: string</code></dt>
       <dd>This label is read by the screenreaders when CodeMirror text area is focussed. This
       is helpful for accessibility.</dd>
 
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -422,7 +422,7 @@
       disallowed.</dd>
 
       <dt id="option_screenReaderLabel"><code><strong>screenReaderLabel</strong>: string</code></dt>
-      <dd>This label is read by the screenreaders when CodeMirror text area is focussed. This
+      <dd>This label is read by the screenreaders when CodeMirror text area is focused. This
       is helpful for accessibility.</dd>
 
       <dt id="option_showCursorWhenSelecting"><code><strong>showCursorWhenSelecting</strong>: boolean</code></dt>
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -421,6 +421,10 @@
       simply <code>true</code>), focusing of the editor is also
       disallowed.</dd>
 
+      <dt id="option_srLabel"><code><strong>srLabel</strong>: string</code></dt>
+      <dd>This label is read by the screenreaders when CodeMirror text area is focussed. This
+      is helpful for accessibility.</dd>
+
       <dt id="option_showCursorWhenSelecting"><code><strong>showCursorWhenSelecting</strong>: boolean</code></dt>
       <dd>Whether the cursor should be drawn when a selection is
       active. Defaults to false.</dd>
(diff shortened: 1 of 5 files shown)


src/input/input.js
```diff
@@ -119,9 +119,15 @@ export function disableBrowserMagic(field, spellcheck, autocorrect, autocapitali
...
 
-export function hiddenTextarea() {
+export function hiddenTextarea(label) {
   let te = elt("textarea", null, null, "position: absolute; bottom: -1em; padding: 0; width: 1px; height: 1em; outline: none")
-  let div = elt("div", [te], null, "overflow: hidden; position: relative; width: 3px; height: 0px;")
+  let te_id = 'cm-textarea-' + Math.floor(Math.random() * 9999)
+  te.setAttribute('id', te_id)
+  /* Label for screenreaders, accessibility */
+  let le = elt("label", label, null, 'position: absolute; width: 1px; height: 1px; padding: 0; overflow: hidden; clip: rect(0, 0, 0, 0); white-space: nowrap; border: 0;')
+  le.setAttribute('for', te_id)
```
marijnh: Why aren't you using an `aria-label` attribute directly on the textarea?

src/edit/options.js
```diff
@@ -132,6 +132,7 @@ export function defineOptions(CodeMirror) {
     }
     cm.display.input.readOnlyChanged(val)
   })
+  option("srLabel", 'CodeMirror text input')
```
marijnh: I think I'd prefer to call this option either just `label` or `screenReaderLabel`. The abbreviation seems unnecessary.

marijnh:
Thanks, this would be good to have. See my comments, and you'll need to also add the label to the contenteditable input code somehow.

adityatoshniwal:
Thank you for reviewing @marijnh. I'll fix them.

adityatoshniwal:
Hey @marijnh, is there any example of contenteditable in the manual ? I can't find it.

marijnh:
See the [`inputStyle`](https://codemirror.net/doc/manual.html#option_inputStyle) option. You'll have to edit `src/input/ContentEditableInput.js` to assign a label to the editable `<div>` element that creates.

adityatoshniwal:
Hi @marijnh, I've made required changes. Please review.

doc/manual.html
```diff
@@ -421,6 +421,10 @@ <h2>Configuration</h2>
       simply <code>true</code>), focusing of the editor is also
       disallowed.</dd>
 
+      <dt id="option_screenReaderLabel"><code><strong>screenReaderLabel</strong>: string</code></dt>
+      <dd>This label is read by the screenreaders when CodeMirror text area is focussed. This
```
marijnh: It's "focused" (one _s_)

src/edit/options.js
```diff
@@ -132,6 +132,7 @@ export function defineOptions(CodeMirror) {
     }
     cm.display.input.readOnlyChanged(val)
   })
+  option("screenReaderLabel", 'CodeMirror text input')
```
marijnh: I think it'd make sense to default this to null (and have the input implementations avoid setting the attribute when the option is null). No need to spam screen reader users with text when we don't have anything relevant to add.

Also, it'd be good if this option could be changed dynamically. Instead of integrating the setting of the attribute in the input object creation, maybe make it a method that the editor calls on init, and then again when this option is changed.

marijnh:
Thanks! I added two more remarks.

src/edit/options.js
adityatoshniwal: OK. Got it.

marijnh:
Wonderful, thank you. Merged.

adityatoshniwal:
Hey,
I missed one change. Can you please correct this. :(

`   createField(_display) {
     // Wraps and hides input textarea
-    this.wrapper = hiddenTextarea(this.cm.options.screenReaderLabel)
+    this.wrapper = hiddenTextarea()
     // The semihidden textarea that is focused when the editor is
     // focused, and receives input.
`

marijnh:
Done in attached patch.

adityatoshniwal:
Cool !! Thanks.

PR #6200, opened 2020-04-01
Simple mode implementation for 32bit ARM assembler code.

gerph:
ARM assembler is relatively simple, with a relatively small instruction
set. Over time, the instruction set has added more complex instructions,
which aren't supported within this mode at the current time. The
mode is based on the ARM assembly language as used in the BBC BASIC
assembler, with some extensions taken from ObjAsm.

I'm not sure what needs to be added for testing purposes, or if there's any additional changes that are needed within the source. I created this mode because I needed a variant of the ARM instruction set colouring in an assembler I wrote, and was surprised that there wasn't an assembler mode present already that I could base off.

### DIFF
--- a/mode/assembler/arm.js
+++ b/mode/assembler/arm.js
@@ -1,0 +1,339 @@
+/* Simple mode definition for ARM code.
+ * Based on the example Javascript Simple Mode.
+ */
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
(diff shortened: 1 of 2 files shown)


marijnh:
Hi. Thanks for writing a mode, but unfortunately we are no longer adding new modes to the main distribution (due to maintainer burden concern). The current recommended way to distribute a mode is to put it in its own npm package (see for example the [Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir)).

PR #6201, opened 2020-04-01
Add BBC BASIC text colouring mode.

gerph:
BBC BASIC has dialects that have diverged from the original version produced
on the BBC Micro, with the most common usages being the BB4W and RISC OS
variants.

This mode covers the RISC OS BASIC V dialect in text (not tokenised)
format, with colouring for embedded ARM assembly, and allows lines to
be prefixed with or without line numbers.

The meta description uses an 'extension' check for the suffix ',fd1' on
filenames. This is the non-RISC OS file system representation of the
filetype for detokenised textual BASIC files.

I'm not sure what needs to be added for testing purposes, or if there's any additional changes that are needed within the source. I created this mode because I needed a variant of the BBC BASIC language colouring in an assembler and build system that I wrote, and was surprised that there wasn't any BASIC mode supplied which I could base the code off.

### DIFF
--- a/mode/basic/bbcbasic.js
+++ b/mode/basic/bbcbasic.js
@@ -1,0 +1,290 @@
+/* Simple mode definition for BBC Basic (with ARM assembler)
+ * Based on the example Javascript Simple Mode.
+ */
+
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"), require("../../addon/mode/simple"));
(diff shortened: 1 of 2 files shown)


jdbruxelles:
Hi. New modes are not accepted into the main distribution anymore. The recommended way to distribute code like this is to put it into its own npm package with a name like codemirror-mode-bbcbasic. This is valid for #6200 too.

marijnh:
RIght, see my comment on #6200

PR #6208, opened 2020-04-09
[show-hint] Scroll cursor into view after picking.

fonsp:
**Fixes a minor visual bug:**
The picked hint might push the cursor out of the screen - this scrolls the cursor back into view. (Would normally happen after the next keypress.)

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -89,8 +89,8 @@
       if (completion.hint) completion.hint(this.cm, data, completion);
       else this.cm.replaceRange(getText(completion), completion.from || data.from,
                                 completion.to || data.to, "complete");
-      this.cm.scrollIntoView();
       CodeMirror.signal(data, "pick", completion);
+      this.cm.scrollIntoView();
       this.close();
     },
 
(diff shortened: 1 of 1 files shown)


marijnh:
Could you wrap the new line and the two lines about it in a `cm.operation(function() { ... })` call? That's more efficient (avoids multiple rounds of DOM measuring).

fonsp:
Will do

fonsp:
Sorry, the `scrollIntoView` call should have come _after_ the `"pick"` signal, otherwise using the Enter key to pick a completion would also insert a `\n`. This is fixed in the last commit 👆.

This means that they can no longer be wrapped together - but aren't two rounds of DOM measuring necessary to scroll to the new position?

marijnh:
You can wrap the `signal` call in the operation as well. Two calls may still be necessary, but it's still better to let the editor handle the scheduling.

fonsp:
I'll do it today 📅

fonsp:
Here you go! It worked in my tests - although i have not tested custom `completion.hint` functions.

marijnh:
Thanks!

PR #6214, opened 2020-04-12
docs: Fix simple typo, compatiblity -> compatibility

timgates42:
There is a small typo in addon/search/searchcursor.js, mode/python/python.js.

Should read `compatibility` rather than `compatiblity`.


marijnh:
Thanks!

PR #6218, opened 2020-04-13
Add missing shortcut (Ctrl-K Ctrl+1 to fold all) to Sublime Text keymap

ariasuni:
In Sublime text, `Ctrl-K Ctrl-1` is used to Fold all levels, but CodeMirror is missing it.

It seems the command wasn’t present in a keymap before, so maybe it should be checked.


marijnh:
Thanks! Looks good.

PR #6231, opened 2020-04-17
Add ctrl left & right to Emacs keymap

arj03:
I noticed this was missing compared to my Emacs


marijnh:
Ah, the intention there was, I think to include the default keymap, since there's a bunch of other standard key combinations the emacs mode doesn't explicitly define. Attached patch adds this.

arj03:
Perfect, thanks :)

PR #6238, opened 2020-04-20
[wast] Add new wast mode (for WebAssembly disassembly)

bmeurer:
This adds support for WebAssembly text format, as used in the
browser DevTools (Firefox, Edge and Chrome DevTools all the like).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -97,6 +97,7 @@ Bastian Müller
 belhaj
 Bem Jones-Bey
 benbro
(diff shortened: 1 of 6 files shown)


marijnh:
In general, we're not accepting new modes into the main distribution, but I guess given this one's relevance to browser devtools (and its simplicity), I can make an exception. Merging.

bmeurer:
Awesome, thanks!

PR #6254, opened 2020-04-30
Add support for globalThis in standalone mode

TimvdLippe:
We use the standalone mode in DevTools in a worker, where `window`
does not exist. Instead, we can use `globalThis` which should be
the corresponding global scope in all scenarios. For older browsers
that do not support `globalThis`, we fallback to `window`.

Relevant DevTools change:
https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/2174418/1/front_end/third_party/codemirror/package/addon/runmode/runmode-standalone.js


marijnh:
Makes sense, thanks.

PR #6259, opened 2020-05-04
Add modern ESM output bundle

TimvdLippe:
This adds a new output bundle using ES exports, which would allow
users to import CodeMirror using ES imports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 .idea
 *.iml
 /lib/codemirror.js
+/lib/codemirror.mjs
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,9 +1,6 @@
 import buble from 'rollup-plugin-buble';
 
-export default {
-  input: "src/codemirror.js",
-  output: {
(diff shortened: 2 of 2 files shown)


TimvdLippe:
In DevTools, we would like to import CodeMirror using ES modules, but the lib folder didn't contain an ESM bundle yet.

While this would work when importing CodeMirror by default, I later realized that addons, etc... require the global `CodeMirror`. That would still need to be exported somewhere. Let me know what your thoughts are on this approach.

marijnh:
Hi. Thanks, but as you noticed, this'll be a rather useless bundle—you'll only be able to use it on its own, without addons. So I feel that it being present is more likely to confuse people than to help. Version 6 is ES modules throughout, but for version 5 I don't intend to make the addons ES module compatible (which would involve shipping two versions of all of them).

So I don't think this is a good idea. It's easy enough to run some tool locally if you really want an ES module version of the library (and to do the same for the addons you need).

TimvdLippe:
Ah I wasn't aware of version 6 that addresses this issue. Looking forward to that release, thanks for the reply!

PR #6267, opened 2020-05-12
Add an "id" to all modes

bokub:
Hi,

My suggestion is that all modes in `meta.js` would have an `id` which is
- unique
- mandatory
- short (1 to 4 characters)
- lowercase alphanumeric
- could never be changed

Why?

Right now, the only "unique" field in the modes is the **name**. 

So if I want to store my favorite mode in a database, or have it as a parameter in a URL, or even send in in a API call, I have to use the `name` of this mode

What I don't like is that the name can be subject to changes, it can contain special characters, and it can be long. In a nutshell: I don't like using the name field as an ID.

Please tell me what you think of this idea

(I have added tests for this)


marijnh:
No. As you noticed, the name is unique, and can thus already serve as id. I'm not interested in adding stuff to this file just to accommodate to your database design preferences.

bokub:
All right, no problem, it was just a suggestion

Issue #6268, opened 2020-05-12
Support Dart's new Null Safety type system


srawlins:
Dart's new [Null Safety language feature](https://github.com/dart-lang/language/blob/master/accepted/future-releases/nnbd/feature-specification.md) introduces some new keywords and a new way to specify types:

* `late` and `required` are new keywords
* `Never` is a new type found in `dart:core`
* (almost) all types can be suffixed with a `?` to indicate a nullable type. For example, in:

  ```dart
  f(int x) { ... }
  g(int? x) { ... }
  ```

  `f` is a function which accepts a non-null `int` parameter, and `g` accepts a nullable `int` parameter.

In this example:

```dart
late int x;
f({required int? y}) {}
const List<Never> z = [];
```

`late` and `required` should be highlighted as keywords. `int?` should highlight the same as `int`. `Never` should be highlighted like the other built-in types.

PR #6269, opened 2020-05-12
Add `late` and `required` keywords, and `Null` and `Never` types to Dart

srawlins:
Addresses most of #6268 

Types with `?` suffixed remain unsupported. I wasn't sure the best way to add this. Just duplicate (approximately) the list and add `?`?


marijnh:
`?` won't be tokenized as part of the identifier, so adding that to these lists won't help. Isn't treating it as a (type) operator good enough?

srawlins:
Probably good enough :)

PR #6272, opened 2020-05-13
[vim] Allow Ex-Commands with non-word names

benhormann:
I'm attempting to implement the `:!` command, but currently the name `!` isn't fully supported (adding an argument causes the command to not be found). This PR fixes it (for commands listed in [index.txt#ex-cmd-index](https://vimhelp.org/index.txt.html#ex-cmd-index)).

I was going to put `/^(\w+|!!?|@@?|[#&*<=>~])/`, but `!!` and `@@` don't take parameters.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -4456,7 +4456,7 @@
         }
 
         // Parse command name.
-        var commandMatch = inputStream.match(/^(\w+)/);
+        var commandMatch = inputStream.match(/^(\w+|!!|@@|[!#&*<=>@~])/);
         if (commandMatch) {
           result.commandName = commandMatch[1];
         } else {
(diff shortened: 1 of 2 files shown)


marijnh:
Could you add a test? (Since I don't understand vim well enough to know when I break something, I really want tests for this stuff.)

benhormann:
With the current proposal, `:!!` is interpreted as command `!` with arg `!`. You _could_ write code to handle that, but it would probably be nicer to define `!!` as a separate command.
I'll update it and add tests.

benhormann:
I think the test is near enough.
_Could do with something like QUnit's [expect](https://api.qunitjs.com/assert/expect)... but `eq` doesn't keep a count._

marijnh:
Thanks!

Issue #850, opened 2012-09-25
Add support for manual resizing


flack:
Sorry if this is a duplicate, but a quick search of the issue list found nothing that seemed related.

I was wondering if CodeMirror could be made to support the CSS resize attribute. So that I could write something like this:

``` css
.CodeMirror-scroll
{
    resize: vertical;
}
```

In the current release (2.34), this actually renders a resize handle, but I can't move it. Or am I just doing it wrong?

mindplay-dk:
I'd like to just have a simple drag-handle, a bar running along the bottom edge of the editor, that the user can drag to resize vertically.

This seems like a pretty basic requirement? A standard textarea can be made user-resizeable.

Yet, I can't find an add-on or plugin or fiddle or anything else demonstrating this.

Has anyone figured out how to make the window resizable?

(it would be really nice to have a standard add-on listed for this basic feature - many of the listed/standard add-ons are far more advanced features than this...)

marijnh:
The reason this isn't an addon is probably that it's rather easy to implement, and since it involves UI elements, providing a general-enough implementation that fits everyone's needs is more involved than just coding one-off implementations per application.

Given that we're not adding new addons to the main distribution anymore, I'm going to close this as out of scope. But if you want to tackle it in a reusable way and publish it on NPM, that'd be wonderful.

mindplay-dk:
I disagree that anything to do with drag-and-drop in DOM/JS is easy to implement.

It took me an hour just to get [this far](https://jsfiddle.net/mindplay/rs2L2vtb/) and I still have no idea how to make it capture the mouse, so that the drag operation continues if you drag outside the window/frame. But I'm probably doing something wrong. There's a million ways drag-and-drop goes horribly wrong with DOM and JS.

IMO the lib should offer functionality on par with a plain textarea. The question gets asked enough on StackOverflow, and nobody gives any practical answer or an example, just short remarks like "do it yourself", or in one case how to do it with jQuery UI, which isn't going to be the answer most people are looking for in this day and age ;-)

mindplay-dk:
If somebody else needs it (or wants to take point on turning it into an add-on) here's a mostly working [vertical resize bar](https://jsfiddle.net/mindplay/rs2L2vtb/). If you click and drag repeatedly, it still sometimes gets confused and "sticks", but most of the time it works. Sorry, I did my best ;-)

Sphinxxxx:
[...]

PR #6280, opened 2020-05-16
Add support for native manual resizing.

scottfalkingham:
This PR attempts to add support for optional native manual resizing of editor windows as requested in #850.  Feedback appreciated.

### DIFF
--- a/addon/display/fullscreen.css
+++ b/addon/display/fullscreen.css
@@ -3,4 +3,5 @@
   top: 0; left: 0; right: 0; bottom: 0;
   height: auto;
   z-index: 9;
+  resize: none;
 }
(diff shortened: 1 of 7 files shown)


marijnh:
Why does the scrollbar need to always show up in resizeable mode?

scottfalkingham:
Hi @marijnh,

This shouldn't affect the scrollbar showing up.  It merely gives space for the resizable toggle if a scrollbar is set to show up.

See the demo `demo/resizable.html`.

marijnh:
I don't think I want this in the core library. It should be possible to implement it in external code, and that has the advantage that I won't be expected to maintain it in the future.

PR #6282, opened 2020-05-16
Fix bug with "map" in type name.

axellew:
Instead of looking for explicit map or list change to look for the '<' to check for param type parameters.
Example:
`map<string, string>`
`list<string>`

Also removed unused `var peekChar = stream.peek();`


marijnh:
Thanks!

PR #6293, opened 2020-05-27
Fix: autocomplete suggestion on _ $ and # for ODBC

ZeeshanNoor:
it is a fix when  autocomplete hints when table name has _ $ or #.

### DIFF
--- a/addon/hint/sql-hint.js
+++ b/addon/hint/sql-hint.js
@@ -264,7 +264,7 @@
       token.string = token.string.slice(0, cur.ch - token.start);
     }
 
-     if (token.string.match(/^[.`"\w@]+(\$?#?\w*)*/g)) {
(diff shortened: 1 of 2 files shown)


addon/hint/sql-hint.js
```diff
@@ -264,7 +264,7 @@
       token.string = token.string.slice(0, cur.ch - token.start);
     }
 
-    if (token.string.match(/^[.`"\w@]\w*$/)) {
+     if (token.string.match(/^[.`"\w@]+(\$?#?\w*)*/g)) {
```
marijnh, line 267: You have an extra space in front of the `if`, misaligning the indentation.

The regexp looks scary like this (nested repeat operators can easily create exponential running time). Why not just replace the `\w` with `[\w$#]`?

ZeeshanNoor: updated the regex and removed space.

Issue #6296, opened 2020-05-28
vim-mode-change signalled twice


benhormann:
When entering insert mode from visual mode, `vim-mode-change` is signalled twice. Firstly with `mode` parameter as `insert`, then with `normal`.

Expected: it should only signal once changing to insert mode.
Repro: from normal mode press `v`, then `c`. (assuming there is a listener on the event).

`exitVisualMode` is called from `enterInsertMode`. `exitVisualMode` always signals a change to normal mode. `vim.insertMode` is already set to true at that point, so a check could be added there [exitVisualMode:L3206](https://github.com/codemirror/CodeMirror/blob/a66f466e33fd274785b043ce308e745a9c6d08dc/keymap/vim.js#L3206).

I'm working on some unit tests, covering this bug and some other mode transitions.

benhormann:
Fixed in [5.55.0](https://github.com/codemirror/CodeMirror/releases/tag/5.55.0)

PR #6297, opened 2020-05-28
[vim] Fix vim-mode-change event signalled twice

benhormann:
Fix for #6296, and tests.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -3203,7 +3203,7 @@
       vim.visualMode = false;
       vim.visualLine = false;
       vim.visualBlock = false;
-      CodeMirror.signal(cm, "vim-mode-change", {mode: "normal"});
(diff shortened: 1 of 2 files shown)


benhormann:
I found a test that fails:
```
  test('v', 'visual');
  test(':', '');
  test('<Esc>', 'normal'); // failure: 'normal;normal' != 'normal'
```
However if you press those keys on vim.html, it just stays in visual mode.
_For reference, Vim goes to normal mode._

@marijnh What do you reckon?
IMO it's a separate issue, and very unlikely to affect anyone (asides from having to press <Esc> twice).

marijnh:
As long as that isn't a regression caused by this PR, I'm okay with merging as-is.

benhormann:
Not a regression. I'm sure because this keymap doesn't listen to its own events.

marijnh:
Thanks!

PR #6309, opened 2020-06-09
[julia mode] Tweaked indent method

fonsp:
Hey! Thanks for all your work ❤

Here's a GIF describing the issue:
cmindentissue

This PR fixes that issue! 

# Before

image

# After

image

(both are auto indented)


marijnh:
Hi. Thanks for the patch. It looks good except that the library still supports IE11, so you'll have to use a regexp rather than the `startsWith` method to do these tests.

PR #6313, opened 2020-06-10
update runmode-node to match core

BrianHung:
This PR does the following:
- updates StringStream to [`src/util/StringStream.js`](https://github.com/codemirror/CodeMirror/blob/master/src/util/StringStream.js)
- updates runMode to use the oracle interface in updated StringStream
- updates resolveMode to same method defined in [`src/modes.js`](https://github.com/codemirror/CodeMirror/blob/master/src/modes.js)
- includes `copyState` from `src/modes.js` to allow for some languages like Markdown or HTMLMixed to work with runmode.node.js

### DIFF
--- a/addon/runmode/runmode.node.js
+++ b/addon/runmode/runmode.node.js
@@ -164,6 +164,18 @@ exports.getMode = function(options, spec) {
   return modeObj;
 };
 
+exports.copyState = function(mode, state) {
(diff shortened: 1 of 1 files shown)


marijnh:
This is superseded by 001e07fcf8b2cbd7 I believe.

PR #6316, opened 2020-06-13
Add ASL mode

kylon:


marijnh:
Hi. Unfortunately, we aren't accepting more modes into the main distribution, in order to keep the maintenance burden within reasonable bounds. So the recommended way to distribute a new mode is to publish it separately, like for example the [elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir).

PR #6321, opened 2020-06-17
[wast mode] Support some Wasm SIMD instructions

ngzhian:
These keywords can be found at
https://github.com/WebAssembly/simd/blob/master/proposals/simd/BinarySIMD.md.


ngzhian:
Ping for review. Does this change look okay? Is the regex change clear? Thanks!

marijnh:
I put this aside because of the initial "This is not complete yet, more to come" comment.

ngzhian:
Yea that was misleading, because I really meant "there are a lot of SIMD instructions, so let me implement them over multiple PRs. This is not the full set, and hopefully small enough to review easily".
Sorry for the confusion, and thank you for merging!

PR #6330, opened 2020-06-26
[wast mode] Add more SIMD instructions

ngzhian:
Some arithmetic instructions, had to shuffle the regex a bit so that we
will attempt to match i8x16.add_saturate_s first before i8x16.add.

Full list available at:
https://github.com/WebAssembly/simd/blob/master/proposals/simd/BinarySIMD.md


marijnh:
If there's a lot more of these coming, could you combine them into a single PR? That would save some work.

ngzhian:
There were 2 more, and it's easy enough to send a PR containing both, see #6334, thanks!

PR #6336, opened 2020-06-30
Add events to vimAPI documentation

### LINKED ISSUE #6284: Feature request: events on vim mode changes
I wanted to make a simple customization to code mirror -- disabling blinking cursor when in normal mode and enabling it in insert mode. For this, I tried to look through events and vim mode API sections.

I'm aware that my use case can be implemented in other ways -- by hijacking keys which take you between visual, insert and normal modes.

But I think having event hooks for when editor enters one of these could be better and can perhaps have other use cases.

benhormann:
Fixes #6284


marijnh:
Thanks!

PR #6352, opened 2020-07-17
[show-hint addon] Introduced option 'scrollMargin'

nibinqtl:
To show some extra lines before and after the selected line.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -378,7 +378,8 @@
       CodeMirror.signal(this.data, "select", this.data.list[this.selectedHint], node);
     },
 
-    scrollToActive: function(pad=0) {
+    scrollToActive: function(pad) {
+      if (pad === undefined) pad = 0;
       var node1 = this.hints.childNodes[Math.max(0, this.selectedHint-pad)];
       var node2 = this.hints.childNodes[Math.min(this.data.list.length-1, this.selectedHint+pad)];
       var firstNode = this.hints.firstChild;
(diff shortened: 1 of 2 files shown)


marijnh:
Could you rename this to `scrollMargin` to align it with the [top-level option](https://codemirror.net/doc/manual.html#option_cursorScrollMargin) that does the same in the editor itself?

nibinqtl:
> Could you rename this to `scrollMargin` to align it with the [top-level option](https://codemirror.net/doc/manual.html#option_cursorScrollMargin) that does the same in the editor itself?

Done

addon/hint/show-hint.js
```diff
@@ -379,12 +379,15 @@
     },
 
     scrollToActive: function() {
-      var node = this.hints.childNodes[this.selectedHint]
+      var margin = this.completion.options.scrollMargin || 0;
+      if (! Number.isInteger(margin)) margin = 0;
```
marijnh, line 383: We are still targeting IE9, which doesn't have `Number.isInteger`. The library as a whole doesn't do run-time type checking, so just dropping this line would be fine.

addon/hint/show-hint.js
```diff
@@ -379,12 +379,15 @@
...
+      var node1 = this.hints.childNodes[Math.max(0, this.selectedHint-margin)];
+      var node2 = this.hints.childNodes[Math.min(this.data.list.length-1, this.selectedHint+margin)];
```
marijnh, line 385: (Could you follow the whitespace-around-operators style from the surrounding code?)

addon/hint/show-hint.js
nibinqtl: done

addon/hint/show-hint.js
nibinqtl: done

marijnh:
Thanks!

PR #6377, opened 2020-08-08
customs addons for ISL

KinjalHinsu:


marijnh:
Sorry, but we're not adding further contributed modules to the library. The best way to distribute this would be as a separate npm package.

PR #6380, opened 2020-08-12
Improvement of fullscreen.js addon

graphemecluster:
Now it is able to make it to be displayed in the browser fullscreen mode.
All `S` in `FullScreen` or `fullScreen` are changed to lowercase `s`.
This version also remove fullscreen.css.

Available options are listed below:
- `bindDocument`
 Whether to call the browser's `requestFullscreen()`. Default to false.
- `shortcut`
 Whether to bind F11 (toggle) and Esc (exit) to the CodeMirror instance for switching between fullscreen mode. Default to false.
- `documentShortcut`
 Whether to bind F11 and Esc to the document. Default to false.
- `width`, `height`, `position`, `top`, `right`, `bottom`, `left`, `zIndex`
 If any of them are set it will override the default style. If any of them are set to false the style would not be set.
- `className`
 Custom class name in fullscreen. Adding of the class can be disabled by setting it to false. Default to `CodeMirror-fullscreen`.
- `onchange`: function(state)
 Called whenever the state is changed. `state` would be a boolean.
- `onenter`: function()
 Called when entering fullscreen.
- `onexit`: function()
 Called when exiting fullscreen.

Extensions:
- `cm.toggleFullscreen(state)`
  Switch between fullscreen mode. If `state` is unset it is toggled.
- `cm.isFullscreen()`
  Whether it is in fullscreen mode.

I apologize for any poor English. For anything unclear just read the code.

### DIFF
--- a/addon/display/fullscreen.css
+++ b/addon/display/fullscreen.css
@@ -1,6 +1,0 @@
-.CodeMirror-fullscreen {
-  position: fixed;
-  top: 0; left: 0; right: 0; bottom: 0;
-  height: auto;
-  z-index: 9;
-}
--- a/addon/display/fullscreen.js
+++ b/addon/display/fullscreen.js
@@ -1,6 +1,5 @@
 // CodeMirror, copyright (c) by Marijn Haverbeke and others
 // Distributed under an MIT license: https://codemirror.net/LICENSE
-
 (function(mod) {
   if (typeof exports == "object" && typeof module == "object") // CommonJS
     mod(require("../../lib/codemirror"));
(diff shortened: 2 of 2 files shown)


marijnh:
This would be a very backwards-incompatible change, which isn't acceptable for this project. In general, it seems like you're doing something quite a bit beyond the scope of the current addon. It might make more sense to distribute this separately, outside of the main distribution.

graphemecluster:
How about if all actions to `document` are removed?
i.e. Improve it without implanting browser fullscreen?

marijnh:
Both removing the CSS file and renaming exported parts of the interface are breaking changes. If you have specific improvements that you want me to review, you are welcome to submit them as PRs with a smaller scope (preferably a single fix or feature).

graphemecluster:
And, just one more thing, neither the original nor this fullscreen.js works with panel.js.

Issue #6385, opened 2020-08-17
Code Fold Missing Scrollbar Annotation


tokafew420:
The scrollbar annotation doesn't show a match when used with the code folding addon in the following scenarios:
-  If there is a match on the folded line and there are also matches within the folded content then when the fold is collapsed, we expect at least the match on folded line to be annotated on scrollbar. But it's not.
- Matches hidden within a collapsed fold are not annotated on the scrollbar.

See Fiddle: https://jsfiddle.net/tokafew420/1exL7kqb/

Expanded
![image](https://user-images.githubusercontent.com/10535048/90353944-bb53fc80-e015-11ea-854c-43c4f9ba5430.png)

Collapsed
![image](https://user-images.githubusercontent.com/10535048/90353951-bf801a00-e015-11ea-8c67-baa6768ddc2e.png)

I guess technically the second scenario is not a bug but rather just how the AnnotateScrollbar addon works however it's not what I was expecting.

I'm looking into a possible solution however I thought I'd create this issue in case there's already a fix for this or if someone has insights into how it can be fixed.

**Codemirror:** 5.56.0
**Broswer:** Google Chrome 84.0.4147.125

tokafew420:
Closing. Fixed with commit ac4aaa08e271766d0dbe3ad81603a559d4956da5

PR #6388, opened 2020-08-19
Annotate scrollbar when matches are folded

tokafew420:
PR for issue #6385 

The changes to annotatescrollbar.js will check to see if the current match is collapsed ("folded"). And if so, then use the first line of the markText.lines as the current line for the annotation. This addresses the 2 issues:

- If there is a match on the folded line and there are also matches within the folded content then when the fold is collapsed, we expect at least the match on the folded line to be annotated on scrollbar. But it's not.
- Matches hidden within a collapsed fold are not annotated on the scrollbar.

**Disclaimer:**
I'm not familiar enough with the wide scope of this project so I cannot guarantee that this will not have unintended side effects. That being said, I did my best to ensure that all lint/test cases are passing.

### DIFF
--- a/addon/scroll/annotatescrollbar.js
+++ b/addon/scroll/annotatescrollbar.js
@@ -72,10 +72,20 @@
     var wrapping = cm.getOption("lineWrapping");
     var singleLineH = wrapping && cm.defaultTextHeight() * 1.5;
     var curLine = null, curLineObj = null;
+
+    function getFoldLineHandle(pos) {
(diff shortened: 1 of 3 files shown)


marijnh:
Thanks! I followed up with 55d0484 to make the code a bit shorter and more general (also handles nested collapsed ranges).

tokafew420:
Thanks @marijnh for the elegant touchups.
I pulled your changes and verified that it does resolve #6385 
I appreciate your help on this. Thanks again.

PR #6389, opened 2020-08-19
Firefox cmd key and CodeMirror.keyName fix

benhormann:
Command &#8984; (A.K.A. OSLeft/OSRight or Meta) keys' keyCodes in Chrome/Safari are 91 and 93, but both are 224 in Firefox.
This keyCode is listed as `DOM_VK_META` at <https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode>.

In Chrome/Safari, `keyName` returns "Cmd-Mod". In Firefox, `keyName` and `isModifierKey` return false.
With this PR they return "Mod" and true, as expected.

Motive: Unlikely anyone is using it in their keymap, but `CodeMirror.isModifierKey(evt)` is helpful when customising.
Assumption: Nobody is doing `addModifierNames("Cmd", evt)`.


marijnh:
Thanks for looking into that! Merged.

PR #6399, opened 2020-09-01
[vim] Add black hole register

### LINKED ISSUE #6223: [vim] The black hole register ( "_ ) is not handled correctly
In Vim, when writing to the black hole register (`_`), for example with `"_dd`, the default register should not be written to. Likewise, when reading from `_`, the default register should not be read from.

The CodeMirror behavior does not match this.
Tested on https://codemirror.net/demo/vim.html against VIM 8.1:

**VIM (expected)**:
* Issuing `"_dd`, then `p` will paste the whatever was in the default register before the `"_dd` command. 

**CodeMirror (actual)**:
* Issuing `"_dd`, then `p` will paste the deleted text from the `"_dd` command.

The docs on the black hole register can be found with `:help change` and searching "black hole". Web mirror for these docs is [here](https://vimhelp.org/change.txt.html#change.txt).

benhormann:
Fixes #6223

Caveats: In theory you could record a macro into the register and fill it with text that way, or by programmatic means. Then it will show up in `:registers` and allow pasting via `"_p`.
([Line 2363](https://github.com/codemirror/CodeMirror/blob/6fcc49d5321261b525e50f111f3b28a602f01f71/keymap/vim.js#L2361) is where recording can be avoided, but that line is long enough already).

Edit: Or just leave '_' out of validRegisters. Works for change/delete, but not paste. (Pasting from nowhere is about as unusual as recording to nowhere).

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -8,7 +8,7 @@
  * Supported Ex commands:
  *   Refer to defaultExCommandMap below.
  *
- * Registers: unnamed, -, a-z, A-Z, 0-9
+ * Registers: unnamed, -, ., :, /, _, a-z, A-Z, 0-9
  *   (Does not respect the special case for number registers when delete
  *    operator is made with these commands: %, (, ),  , /, ?, n, N, {, } )
  *   TODO: Implement the remaining registers.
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks!

PR #6407, opened 2020-09-10
Add commands for programatically folding code

lukehaas:


marijnh:
"Commands", in CodeMirror's system, are functions of only a single argument that can be used for key bindings and such. Does the `foldCode` method defined with `defineExtension` further up in this file not fit your purpose?

lukehaas:
I see now. Yes, the `foldCode` method does fit this purpose, thanks.

PR #6417, opened 2020-09-17
[lint addon] Use separate CSS classes for common lint styles

Clashsoft:
This changes `lint.css` to be less reliant on the predefined severities (`error` and `warning`), in turn making it easier to define custom ones. Now all that needs to be done in order to define a new severity, e.g. `note`, is to add the following CSS:

```css
/* underline */
.CodeMirror-lint-mark-note {
  background-image: ...;
}

/* icon */
.CodeMirror-lint-marker-note, .CodeMirror-lint-message-note {
  background-image: ...;
}
```

Previously, it was necessary to copy many styles that were only available under the `CodeMirror-lint-*-error` and `CodeMirror-lint-*-warning` classes.

### DIFF
--- a/addon/lint/lint.css
+++ b/addon/lint/lint.css
@@ -25,7 +25,7 @@
   -ms-transition: opacity .4s;
 }
 
-.CodeMirror-lint-mark-error, .CodeMirror-lint-mark-warning {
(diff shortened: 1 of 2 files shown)


marijnh:
Good idea, thanks.

Issue #6425, opened 2020-09-25
Support for markdown-admonitions


ehhc:
It would be nice if a mode would support markdown-admonitions (like these: https://www.npmjs.com/package/markdown-it-admonition).

ehhc:
See https://github.com/codemirror/CodeMirror/pull/6426#issuecomment-698785415

PR #6426, opened 2020-09-25
Mode for GFM with admonitions

ehhc:
Solves #6425


marijnh:
Sorry, but as the CONTRIBUTING.md file states, we aren't adding new extensions to this distribution. So this would best be maintained and distributed as a separate npm package.

ehhc:
I feel stupid for skipping that part while reading the contributing.md file... thanks anyway :)

PR #6427, opened 2020-09-25
fix focus/blur issues during editing

rPraml:
We use the CM field in our app and do a page update if the user leaves the CM field. (onBlur)
Unfortunately, blur events are fired during editing, follwed by an immediate "focus" event.
Especially IE (Version 11) fires a lot of "blur/(re)focus" events during editing.

We had a workaround in our code, that detects quick "blur/focus" events, but this workaround had other issues.
This PR fixes firing unneccessary (and unwanted) focus/blur events.

*Current Behaviour:*
When clicking around inside the CM field, unexpected blur events occur, followed by an immediate focus event

*Expected Behaviour:*
The first time you click in the CM field, a focus event should be fired
All subsequent clicks/text selections/d'n'd operations should not fire a "blur/focus" event.
If you click outside the CM field, a blur event should be fired

The idea behind this fix is, not to fire blur/focus as long as the user holds the mouse button down or is doing d'n'd operations

### DIFF
--- a/src/display/focus.js
+++ b/src/display/focus.js
@@ -15,7 +15,22 @@ export function delayBlurEvent(cm) {
   } }, 100)
 }
 
+export function suppressFocusBlur(cm) {
+  if (!cm.state.focused) { return } // do not suppress first focus
+  cm.state.suppressFocusBlur = true
+}
+
+export function enableFocusBlur(cm) {
+  // IE fires some more events, so enable it delayed
+  if (cm.state.suppressFocusBlur) {
+    setTimeout(function() {
+      cm.state.suppressFocusBlur = false
+    },20)
+  }
+}
+
 export function onFocus(cm, e) {
+  if (cm.state.suppressFocusBlur) { return }
   if (cm.state.delayingBlurEvent) cm.state.delayingBlurEvent = false
 
   if (cm.options.readOnly == "nocursor") return
@@ -35,7 +50,7 @@ export function onFocus(cm, e) {
   restartBlink(cm)
 }
 export function onBlur(cm, e) {
-  if (cm.state.delayingBlurEvent) return
+  if (cm.state.delayingBlurEvent || cm.state.suppressFocusBlur) { return }
 
   if (cm.state.focused) {
     signal(cm, "blur", cm, e)
--- a/src/display/focus.js
+++ b/src/display/focus.js
@@ -5,6 +5,17 @@ import { signal } from "../util/event.js"
 
 export function ensureFocus(cm) {
   if (!cm.state.focused) { cm.display.input.focus(); onFocus(cm) }
+  else {
+    // if we are already focused, skip any blur/focus events and do
+    // a refocus after timeout. This is a fix for #6427, because
+    // IE fires blur/focus events if you reposition the cursor by
+    // a mouse click.
+    cm.state.delayingBlurEvent = true
+    setTimeout(function () { if (cm.state.delayingBlurEvent) {
+      cm.display.input.focus()
+      cm.state.delayingBlurEvent = false
+    } }, 100)
+  }
 }
 
 export function delayBlurEvent(cm) {
--- a/src/display/focus.js
+++ b/src/display/focus.js
@@ -15,22 +15,7 @@ export function delayBlurEvent(cm) {
   } }, 100)
 }
 
-export function suppressFocusBlur(cm) {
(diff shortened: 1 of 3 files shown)


marijnh:
Did you look at the existing `delayingBlurEvent` mechanism? Is it not possible to use that in this case?

rPraml:
I took a look at this mechanism, the problem is, I can only suppress one event with `delayingBlurEvent`. IE keeps on firing blur/focus events during a dns-operation. (The screen/mouse cursor flickers)
If you want I can do some more investigations, how to combine my change with the existing mechanism.

marijnh:
Ah, that makes sense.

But I'm really not prepared to just blindly delay these events around mouse events on all platforms. Especially since it seems very likely that this patch will cause such events to be dropped entirely (if you click and then really blur the editor before your timeout fires, I don't think the blur event will ever happen). So as it stands, this patch is not acceptable.

rPraml:
I've updated the PR and found a less invasive solution.
This covers most of my use cases:
- The most annoying blur/focus firing when just clicking around in the CM-field is gone now
- There are some rare drag'n'drop cases where IE fires a blur event although you did not leave the field. I can live with them as d'n'd is completely broken in IE

marijnh:
`ensureFocus` is also used in another context and changing its behavior like this doesn't look safe.

Patch c0b371cf5 suggests that synchronously focusing at this point is a problem, hence the timeout in the existing code. I don't think it's a good idea to start doing that again (and it doesn't seem necessary—you could lift the `if (cm.state.focused)` into `leftButtonDown` and run the old behavior if the editor isn't already focused).

rPraml:
I think I need more help. Did I understand you right, that I should change the leftButtonDown method this way:
```javascript
  function leftButtonDown(cm, pos, repeat, event) {
    if (ie) {
      if (cm.state.focused) {
        cm.state.delayingBlurEvent = true
      }
      setTimeout(bind(ensureFocus, cm), 0); 
    }
    else { cm.curOp.focus = activeElt(); }
```
This will fix my problem in IE.
I'm not sure if it is a good idea to set `delayingBlurEvent = true` - because I do not know if it is safely set back to false

marijnh:
> because I do not know if it is safely set back to false

Indeed, that's no good either. Does something like https://github.com/codemirror/CodeMirror/tree/delay-blur , which preserves the delayBlurEvent flag during draggin, work for you?

rPraml:
Hello @marijnh 
I tried your branch, unfortunately, this does not fix my problem.

I added 
```javascript
    editor.on("focus", function() { console.log("focus")});
    editor.on("blur", function() { console.log("blur")});
```
to the example in index.html and opened it with IE 11.

- Then I clicked into the CM field. the cursor was placed and "focus" was logged.
- When I click into an other place, the cursor is repositioned to the new location and "blur/focus" was logged. And this is my problem, as the "blur" event triggers some update actions in my application.
- I did not try to do drag/drop operations on your branch, yet

cheers
Roland

marijnh:
Oh right, since you mentioned the main problem was dragging, that's what I was thinking about. Does adding a call to `delayBlurEvent` in `leftButtonSelect`, as in my latest revision of the patch (d5079717) help with that?

rPraml:
Hello @marijnh unfortunately, your last change does also not fix my problem.
Do you have the ability to test it in IE?
As already said, it would be very easy to reproduce. I added the two event handlers and clicked several times into the codeMirror field
grafik
as you can see, there are several focus/blur events.
Doing the same in Firefox/Chrome/Edge gives me only one "focus" event

marijnh:
> Do you have the ability to test it in IE?

I was trying to avoid that, but I gave in and did an IE debugging session. Patch 8e7f6728 resolves all the spurious focus/blur events I could find.

rPraml:
@marijnh great, your patch fixes my main issue in IE. Drag & Drop operations in IE still fire a blur event which is not really a problem for me. Unfortunately, the drag&drop in FireFox/Chrome is broken now:

**How to reproduce**
Select a line in FF/Chrome:
grafik
Drag & drop it
grafik
Try to place the cursor again. You cannot refocus the element in FF/Chrome now (IE works)

Dropping the line fires a blur event now in all browsers (which is more or less unexpected, but I can live with that)

marijnh:
I wasn't able to reproduce the new issue on Firefox, but I did see it on Chrome. Patch 12512d3e should help.

PR #6432, opened 2020-10-05
Adds option to turn off highlighting of non-standard CSS properties

### LINKED ISSUE #6402: Valid css property flagging red ?
I have valid css property "**margin-inline-start:1px**" but code mirror flagging this prop as red ? what is option available to make this prop not to flag red.

nina-py:
Fixes #6402. The new option is true by default (so no change for existing setups) and is scoped to the CSS mode only.


marijnh:
Thanks!

PR #6439, opened 2020-10-15
Improve parsing of IRI atoms (sparql mode)

MarkBoyes:
Do not treat the opening "<" character of an unprefixed Datatype IRI as an operator along with the preceding "^^" characters. Respect the SPARQL 1.1 percent-encoding and backslash-escaping within the suffix of a prefixed IRI (and add colons to the set of allowable IRI characters).  

To address a single issue of everything after the last "#" within an IRI (a common character in IRIs) being highlighted as if it were a comment in some common prefixed and unprefixed cases.

### DIFF
--- a/mode/sparql/sparql.js
+++ b/mode/sparql/sparql.js
@@ -71,7 +71,7 @@ CodeMirror.defineMode("sparql", function(config) {
       return "operator";
     }
     else if (ch == ":") {
-      stream.eatWhile(/[\w\d\._\-]/);
+      eatPnLocal(stream);
       return "atom";
     }
     else if (ch == "@") {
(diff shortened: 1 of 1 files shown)


marijnh:
What exactly is the `^`-related code parsing? `.next()` will _consume_ the character, so as it stands, it looks like it is dropping non-caret characters on the floor with the `ch = stream.next()` call.

MarkBoyes:
Thanks, you have caught a bug. I should have had peek() rather than next(). I will delete this PR and create a new one.
FYI section 2.3.2 of the SPARQL 1.1 spec has a literal which exhibits that change intended here. Although three operator characters appear in a row, only first two should ben highlighted as operators, because the "<" is the start of an atom.

`"42"^^<http://www.w3.org/2001/XMLSchema#integer>`

PR #6447, opened 2020-10-22
[stylus mode] Recognize "url-prefix" token properly

tophf:
Test case:

```css
@-moz-document url-prefix("https") {
  a {} 
}
```

**Expected/fixed behavior:**

  A single `url-prefix` token is recognized.

**Current bugged behavior:**

  `url` and `-prefix` tokens are recognized separately.

The underlying cause was that `documentTypes_` was used to create a regexp in `documentTypesRegexp` so the shorter `url` was always matching first.


marijnh:
Thanks!

PR #6454, opened 2020-10-28
[stylus mode] allow block comments

tophf:
Stylus language [supports block comments](https://stylus-lang.com/docs/comments.html) so this PR adds the missing props copied from the built-in css.js. Now we can use addons/comment.js and <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>/</kbd> hotkey to make block comments!


dnknn:
Great! Thank you very much!

PR #6456, opened 2020-10-28
Add support for Soy Element Composition.

iteriani:
Add support for Soy Element Composition. It has the syntax in the form of

<{foo()}></>

This adds support to pass through allowEmptyTag and to support this mode in closetag.

### DIFF
--- a/mode/htmlmixed/htmlmixed.js
+++ b/mode/htmlmixed/htmlmixed.js
@@ -74,7 +74,8 @@
       name: "xml",
       htmlMode: true,
       multilineTagIndentFactor: parserConfig.multilineTagIndentFactor,
-      multilineTagIndentPastTag: parserConfig.multilineTagIndentPastTag
+      multilineTagIndentPastTag: parserConfig.multilineTagIndentPastTag,
+      allowMissingTagName: parserConfig.allowMissingTagName,
     });
 
     var tags = {};
--- a/mode/xml/xml.js
+++ b/mode/xml/xml.js
@@ -189,7 +189,7 @@ CodeMirror.defineMode("xml", function(editorConf, config_) {
 
   function Context(state, tagName, startOfLine) {
     this.prev = state.context;
-    this.tagName = tagName;
+    this.tagName = tagName || "";
     this.indent = state.indented;
     this.startOfLine = startOfLine;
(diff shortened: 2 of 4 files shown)


mode/xml/xml.js
```diff
@@ -399,7 +399,7 @@ CodeMirror.defineMode("xml", function(editorConf, config_) {
     xmlCurrentContext: function(state) {
       var context = []
       for (var cx = state.context; cx; cx = cx.prev)
-        if (cx.tagName) context.push(cx.tagName)
+        context.push(cx.tagName)
```
marijnh, line 402: This changes the return value of this function, which looks like it will confuse the closetag and xml-hint addons. Why is this change needed?

mode/htmlmixed/htmlmixed.js
```diff
@@ -74,7 +74,8 @@
       name: "xml",
       htmlMode: true,
       multilineTagIndentFactor: parserConfig.multilineTagIndentFactor,
-      multilineTagIndentPastTag: parserConfig.multilineTagIndentPastTag
+      multilineTagIndentPastTag: parserConfig.multilineTagIndentPastTag,
+      allowMissingTagName: parserConfig.allowMissingTagName,
```
marijnh, line 78: It doesn't look like this is used at all.

iteriani: This is passed through to xml (https://github.com/codemirror/CodeMirror/blob/master/mode/xml/xml.js#L55)

mode/xml/xml.js
iteriani: Right now what happens when you do something like

<{foo()}> or even <>, the bit here

https://github.com/codemirror/CodeMirror/blob/master/addon/edit/closetag.js#L131

causes context[context.length - 1] to be context[-1], which is undefined. So the autocomplete would have

<{foo()}></undefined>

mode/htmlmixed/htmlmixed.js
marijnh: Ah, right, my bad.

marijnh:
Thanks for clarifying.

iteriani:
Thanks a lot! This was a great experience.

PR #6461, opened 2020-11-05
 add vlang support

lydiandy:
add vlang(https://vlang.io/) support


marijnh:
As the text you deleted from the textbox when opening this PR clearly state, we aren't accepting new modes into this repository anymore.

PR #6471, opened 2020-11-13
Natively support Soy Element Composition in Soy

iteriani:
Soy Element Composition allows HTML tags of the form:

<{foo()} data-foo="3"></>

Previously, we took advantage of missingTagNames to support this. However, this didn't work if the element had attributes. We instead now don't rely on it and parse the syntax directly in the Soy plugin.

In order to handle autocompletes correctly, we update closeTag.js to use "" instead of "undefined" as the tag if the context is missing a tag. So instead of 

<{foo()}></undefined>

we have 

<{foo()}></>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -208,6 +208,7 @@ David Barnett
 David H. Bronke
 David Mignot
 David Pathakjee
+David R. Myers
 David Rodrigues
 David Santana
 David Vázquez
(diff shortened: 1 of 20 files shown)


marijnh:
The addons have to be written in ES5, hence the test failure on the use of `let`. Could you fix that?

iteriani:
I don't think this is related to this PR. The breakage is from

https://github.com/codemirror/CodeMirror/blob/master/mode/shell/shell.js#L74

Do you want me to fix this in this PR?

marijnh:
Hah, sorry about that. It seems Travis has stopped sending me email when the CI is failing. I've pushed a patch to fix that, and will merge this.

PR #6472, opened 2020-11-14
Document all options for show-hint addon.

nathanlesage:
I recently dug into the `show-hint`-add on, and noticed an odd behaviour: The `close`-event was not fired on the completion object when certain characters were pressed. Upon reviewing the source, I found several undocumented options and one undocumented event in the show-hint add-on.

As I am sure some people are already using the `endCompletion`-event which is being fired on the editor instance, I decided not to touch the source, and instead amend the documentation.

## Changes (manual.html)

- Fix a few typos and grammar in the `show-hint` description
- Document the `closeCharacters` option for the `show-hint` add-on
- Document the `completeOnSingleClick` option for the add-on
- Document the `container` option for the add-on
- Document the `endCompletion` event, which is being fired on the cm-instance on completion.

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -2700,8 +2700,8 @@ editor.setOption("extraKeys", {
       Defines <code>editor.showHint</code>, which takes an optional
       options object, and pops up a widget that allows the user to
       select a completion. Finding hints is done with a hinting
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for filling those in!

PR #6485, opened 2020-11-19
Fix white CodeMirror-scrollbar-filler on dark themes

erosman:
background-color: white; remains white on dark themes which doesn't suit dark background pages. Changing it to transparent to match the theme.


marijnh:
Reverted, see #6599

Scutua (approved):

Issue #6490, opened 2020-11-20
lint: Suggestion to remove duplicate messages


erosman:
There are occasions where the same error or warning is repeated in a line e.g.

![image](https://user-images.githubusercontent.com/6458876/99816624-28b98100-2b61-11eb-9511-779cf7d2f05c.png)

In my personal view, it is not beneficial and distracting. I wrote a patch to remove duplicated messages. 
If it is useful, I can make a PR.

lint.js line 163

```js
  function updateLinting(cm, annotationsNotSorted) {
    clearMarks(cm);
    var state = cm.state.lint, options = state.options;

    var annotations = groupByLine(annotationsNotSorted);

    for (var line = 0; line < annotations.length; ++line) {
      var anns = annotations[line];
      if (!anns) continue;

      // filter out duplicate messages
      const message = [];
      anns = anns.filter(item => message.includes(item.message) ? false : message.push(item.message));

      var maxSeverity = null;
      var tipLabel = state.hasGutter && document.createDocumentFragment();

      for (var i = 0; i < anns.length; ++i) {
        var ann = anns[i];
        var severity = ann.severity;
        if (!severity) severity = "error";
        maxSeverity = getMaxSeverity(maxSeverity, severity);

        if (options.formatAnnotation) ann = options.formatAnnotation(ann);
        if (state.hasGutter) tipLabel.appendChild(annotationTooltip(ann));

        if (ann.to) state.marked.push(cm.markText(ann.from, ann.to, {
          className: "CodeMirror-lint-mark CodeMirror-lint-mark-" + severity,
          __annotation: ann
        }));
      }
      // use original annotations[line] to show multiple messages
      if (state.hasGutter) 
        cm.setGutterMarker(line, GUTTER_ID, makeMarker(cm, tipLabel, maxSeverity, annotations[line].length > 1,
                                                       state.options.tooltips));
    }
    if (options.onUpdateLinting) options.onUpdateLinting(annotationsNotSorted, annotations, cm);
  }
```

Result:
![image](https://user-images.githubusercontent.com/6458876/99817588-5eab3500-2b62-11eb-8ba5-84794304ff9d.png)

marijnh:
Sure, definitely feel free to submit that pull request.

PR #6492, opened 2020-11-20
filter out duplicate messages 2

erosman:
ref: #6490
#6491 failed due The keyword 'const' is reserved (174:6) !!??
Changed to var

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -172,7 +172,7 @@
 
       // filter out duplicate messages
       var message = [];
-      anns = anns.filter(item => message.includes(item.message) ? false : message.push(item.message));
+      anns = anns.filter(item => message.indexOf(item.message) > -1 ? false : message.push(item.message));
 
       var maxSeverity = null;
       var tipLabel = state.hasGutter && document.createDocumentFragment();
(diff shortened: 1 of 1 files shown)


bruchmann:
You don't need to open a new pull request. Just `git push` to the old one.

erosman:
I don't develop on Github so I don't know about `git push` :(

marijnh:
Addons need to run as-is on IE9, so that indeed means no `const`. But the `includes` method is also not available there, so you'd have to `indexOf` instead and compare the result to `-1`.

Updating the branch that the PR refers to updates the PR, so indeed, there's not need to open new ones to add or change patches.

erosman:
Thanks @marijnh for the explanation. Updated the code to use `indexOf`

marijnh:
Thanks. Merged!

erosman:
@marijnh I was thinking, is arrow function supported or should it be changed to normal function?

e.g.
```js
anns = anns.filter(item => message.indexOf(item.message) > -1 ? false : message.push(item.message));
```

Should it be changed to?
```js
anns = anns.filter(function(item) {return message.indexOf(item.message) > -1 ? false : message.push(item.message);});
```

marijnh:
Yes, it should. I've done so in 4f37b1e9

Issue #6489, opened 2020-11-20
Search Addon: Add bottom option to search Dialog


erosman:
Further to [Move Search Dialog to Bottom](https://discuss.codemirror.net/t/move-search-dialog-to-bottom/2719), [dialog.js](https://github.com/codemirror/CodeMirror/blob/master/addon/dialog/dialog.js#L14) has a `bottom` setting  to move the dialog panel to the bottom. However, in `search.js` addon, [`persistentDialog()`](https://github.com/codemirror/CodeMirror/blob/master/addon/search/search.js#L60) & [`dialog()`](https://github.com/codemirror/CodeMirror/blob/master/addon/search/search.js#L70) do not set the option.

I would be useful to have the option and be able to set it e.g.

```js
const editor = CodeMirror.fromTextArea(document.getElementById('code'), {
  mode: 'javascript',
  lineNumbers: true,
  openDialog: {bottom: true},
  extraKeys: {'Alt-F': 'findPersistent'}
});
```

I am not familiar enough with code and the method of passing option to the module yet to make a PR for.

erosman:
I have written the following by looking at other addon codes. I am not sure if this is the best way to do it but it is working, nonetheless.


search.js line 19

```js
})(function(CodeMirror) {
  "use strict";

  // add search panel location default 
  var bottom = false;

  CodeMirror.defineOption("search", false, function(cm, val, old) {
    if (val.bottom) {
      bottom = true;
    }
  });
```

then ..

_Note: I guess [New notations in ECMAScript 2015](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015) is not available in IE9_

```js
  function persistentDialog(cm, text, deflt, onEnter, onKeyDown) {
    cm.openDialog(text, onEnter, {
      value: deflt,
      selectValueOnOpen: true,
      closeOnEnter: false,
      onClose: function() { clearSearch(cm); },
      onKeyDown: onkeydown,
      bottom: bottom
    });
  }

  function dialog(cm, text, shortText, deflt, f) {
    if (cm.openDialog) cm.openDialog(text, f, {value: deflt, selectValueOnOpen: true, bottom: bottom});
    else f(prompt(shortText, deflt));
  }
```

then the initialisation...

```js
const editor = CodeMirror.fromTextArea(document.getElementById('code'), {
  mode: 'javascript',
  lineNumbers: true,
[...]

PR #6493, opened 2020-11-21
Add search panel location option

erosman:
ref #6489

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -19,14 +19,8 @@
 })(function(CodeMirror) {
   "use strict";
 
-  // default search panel location 
-  var bottom = false;
-
-  CodeMirror.defineOption("search", false, function(cm, val, old) {
-    if (val && val.bottom) {
-      bottom = true;
-    }
-  });
+  // default search panel location
+  CodeMirror.defineOption("search", {bottom: false});
 
   function searchOverlay(query, caseInsensitive) {
     if (typeof query == "string")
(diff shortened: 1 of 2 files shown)


addon/search/search.js
```diff
@@ -63,12 +66,13 @@
       selectValueOnOpen: true,
       closeOnEnter: false,
       onClose: function() { clearSearch(cm); },
-      onKeyDown: onKeyDown
+      onKeyDown: onKeyDown,
+      bottom: cm.options.search.bottom
```
marijnh, line 70: This looks like it'll fail with "`cm.options.search` is undefined" when the option _wasn't_ provided.

erosman:
> This looks like it'll fail with "cm.options.search is undefined" when the option wasn't provided.

I tested it and it default to `false` (no error) .eg. `CodeMirror.defineOption("search", {bottom: false});`

Test ...
```js
  function persistentDialog(cm, text, deflt, onEnter, onKeyDown) {console.log(cm.options.search.bottom);
    cm.openDialog(text, onEnter, {
      value: deflt,
      selectValueOnOpen: true,
      closeOnEnter: false,
      onClose: function() { clearSearch(cm); },
      onKeyDown: onKeyDown,
      bottom: cm.options.search.bottom
    });
  }
```

I also need to add code to `jump-to-line.js`
Can I add it to this PR? How?

erosman:
@marijnh The CM manual needs a mention of the new setting. :)

PR #6494, opened 2020-11-21
[hardwrap addon] Minor fix for lines starting with spaces in non-forceBreak mode

kaushikcfd:
Fixes wrapping behavior for lines like:
```
   This line withAVeryLongWordThatIsGreaterThanColumn and words after it.
```


kaushikcfd:
Yes, that works for me. Thank you.

PR #6498, opened 2020-11-25
exclude GitHub props from the release archive

infrastation:


marijnh:
Is this about the 'release' files that github offers for download? Those are pretty useless, and I don't really care what files end up in them. Having them be a precise representation of the repository seems less confusing than creating yet another limited set of files. What motivated this change?

infrastation:
The release zip files available at codemirror.net recently started to have a ".github" directory, which is of no use there. Presuming you generate these archives with "git archive", this change will omit that directory from future archives. This is what used to work well in other projects currently hosted on github.com.

For clarity, I make no claims about the "release" feature of github.com, which is an undefined behaviour.

marijnh:
Ah, no, those aren't generated with `git archive`, but with a custom script (it must build the library and include the build files too). I've adjusted the script to drop `.github` from now on.

PR #6503, opened 2020-12-02
Add missing languages link

benhormann:
Looks like there is only one missing.
bash$ `(cd mode; for f in */; do grep -Fq href=\"$f index.html || echo $f; done)`


marijnh:
Thanks for noticing that!

Issue #6507, opened 2020-12-03
Indentation issue in Scheme mode


jcubic:
I was making modification to Scheme mode to match my Scheme based lips interpreter and found issue with indentation.

`let-syntax`, `letrec-syntax`, `define-syntax` and `syntax-rules` should have 2 space indent. If you're willing to merge I can create a PR. This is how it works in GNU/Emacs (syntax-rules is little bit different in Emacs because it indent 4 spaces if you don't type `()` and press enter just after the keyword, but I think that it's fine to just add all those keywords to 2 indents).

marijnh:
Sure, a PR would be great.

marijnh:
Solved via #6508

PR #6508, opened 2020-12-03
fix indent for scheme #6507

jcubic:
This change fixes indentation of Scheme code that should indent with 2 spaces for few keywords. It also add non standard (by R7RS) `define-macro` and `defmacro` keywords that are supported by a lot of scheme implementation, some Schemes (like BiwaScheme) support only this type of macros.


marijnh:
Thanks!

PR #6509, opened 2020-12-04
[matchbrackets addon] Fix memory leak

ofpiyush:
In `cm.off`,`Array.indexOf` on `_handlers` returns `-1` for a function defined within the scope of another function.

That leads to a memory leak.


marijnh:
Thanks for spotting that!

PR #6511, opened 2020-12-06
fix typos in solarized palette

BurningTreeC:
There were two small typos in the solarized palette...

`ligh -> light`


marijnh:
Thanks, merged (with a typo in the updated commit message, no less—sorry about that)

PR #6519, opened 2020-12-10
Break an inifite loop in text wrapping.

elmar-peise:
This breaks an infinite loop triggered by wrapping a text containing a word longer than the targed width (e.g., a long URL).


marijnh:
Thanks!

Issue #2941, opened 2014-11-23
How to increase line height?


eldargab:
Hi,

I tried to increase the line height via

``` css
CodeMirror pre {
  line-height: 1.5;
}
```

but cursor and selection heights left unadjusted. 

Is there some way to get more space between lines?

lfender6445:
For others looking for same thing try

```
.CodeMirror
{
  line-height:21px;
}

```

PR #6524, opened 2020-12-17
Document singleCursorHeightPerLine option

nathanlesage:
It's me again with an undocumented option! Tested on the most recent CodeMirror version, and works like a charm. (Source: https://github.com/codemirror/CodeMirror/issues/2941)

Thanks again for CodeMirror, and I'm looking forward to switching to CodeMirror.Next ❤️

### DIFF
--- a/doc/manual.html
+++ b/doc/manual.html
@@ -513,13 +513,13 @@
       of the line, looks better</dd>
 
       <dt id="option_singleCursorHeightPerLine"><code><strong>singleCursorHeightPerLine</strong>: boolean</code></dt>
-      <dd>Determines if CodeMirror can expect all lines to be of the
-      same height (<code>true</code>, the default) and the cursor-size
-      can therefore be lazily evaluated. In case your editor contains
-      multiple line-sizes, for instance, if <a href="#addLineClass"><code>addLineClass</code></a>
-      sets classes which contain <code>line-height</code>-rules, you
-      should consider setting this to <code>false</code> to prevent
-      visual artefacts.
+      <dd>If set to <code>true</code> (the default), CodeMirror will
+      calculate the cursor height from the adjacent characters or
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks, but that's not what this option does. It determines whether the cursor height is equal to the highest element on the line, or to the height of the element directly next to the cursor.

nathanlesage:
Ah, alright; I didn't follow the source. But is my impression correct that setting this to true means that it'll lazily evaluate it, and when setting it to false that it'll evaluate it even after, e.g., setting line classes? Then I'll update the PR accordingly.

marijnh:
No, there's no lazy evaluation happening there. The cursor position has to be retrieved from the DOM layout either way. This only changes the way the height of the drawn cursor is set.

nathanlesage:
Alright, I just dug in the code, and it's a little bit difficult to understand due to the complex measurements that are taking place. Did I understand it correctly that when the option is true, the cursor size is being calculated based off the character size at the cursor position, whereas when the option is false, the cursor size is always based off the size of the total bounding rect of the line …?

I'll then adapt the PR accordingly.

marijnh:
Yes, pretty much that, except that for wrapping lines it's just the max height of the boxes that are on the same wrapped line.

nathanlesage:
I've just updated the description with your info, and hope that it now captures what the option does.

On another note, what prompted me to assume this behaviour is that I use `addLineClass` to apply classes to Markdown headings that increase their size, but the cursor will stay on the smaller size until you type anything in the line, only then the cursor height is updated. With setting this option to false, the cursor height is immediately adapted. Is this intended, or might this be an issue that could be fixed? In the latter case: shall I open an issue concerning this?

Thanks!

marijnh:
Merged as abc65fe746384652 and ee414661b9, followed up with a90d0f8e992b6

PR #6540, opened 2020-12-28
Prefer dot notation when possible

Yash-Singh1:
Preferred dot notation:
- Takes less space
- Easier to read


marijnh:
The console.log notation is like that to avoid a lint warning (as the failed CI run shows).

marijnh:
Merged as 4d5da83c

PR #6543, opened 2020-12-29
Extended the `closeOnCursorActivity` flag to be `updateOnCursorActivity`

slepowronski:
I had to slightly change the behavior of `closeOnCursorActivity` flag. In fact, it became `updateOnCursorActivity`.

The reason is that it has been sometimes possible (quite often on Safari) to be stuck with multiple dropdowns.
I'm controlling the opening/closing of the hints from outside but `Completion#cursorActivity` is async as it uses `setTimeout`. As a result, it has been possible to have `new Widget` called twice without closing between so two dropdowns have been displayed (without any way to dismiss the previous one).

Then, I'd prefer to opt-out from the whole "update on cursor activity" functionality (not only closing) as I'm controlling that on my own from outside of the addon.

Thanks for checking this PR out :)

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -61,8 +61,10 @@
     this.startPos = this.cm.getCursor("start");
     this.startLen = this.cm.getLine(this.startPos.line).length - this.cm.getSelection().length;
 
-    var self = this;
-    cm.on("cursorActivity", this.activityFunc = function() { self.cursorActivity(); });
+    if (this.options.updateOnCursorActivity) {
+      var self = this;
+      cm.on("cursorActivity", this.activityFunc = function() { self.cursorActivity(); });
+    }
   }
 
   var requestAnimationFrame = window.requestAnimationFrame || function(fn) {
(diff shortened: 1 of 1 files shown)


marijnh:
I'm not really prepared to continue to track your custom changes to this module and introduce and remove options whenever you do—I think at this point it'd be a good idea for you to just fork the addon.

slepowronski:
@marijnh Yes, I fully understand your concerns. That's the only change that I'd need and it's not because the requirements changed but because the previous implementation of `closeOnCursorActivity` has been faulty to some degree (because it has been using `setTimeout`). Then, I believe that it would be good to fix that also in the main `CodeMirror` repository.

Do you think that we could somehow work out with this PR to make this happen?

I won't be really requiring any further changes and, as I said, this one is not caused by the changed requirements but by the previous one not covering all the edge cases (please take a look at the PR description for more details).

marijnh:
Having a description with the PR helps (it was empty when I first looked at it). What would also help would be rebasing the branch to only show a single, relevant commit. I'll try to take a closer look at this soon.

slepowronski:
Sure, I'll rebase it. Many thanks for your help. I think that it would be beneficial to fix that in the original repo.

That's the faulty case that I'm talking about (incidental on Chrome or Firefox but quite regular on Safari):
Screenshot 2020-12-29 at 14 37 28

slepowronski:
@marijnh Rebase done, we've got only one commit right now.

marijnh:
Merged!

slepowronski:
Amazing, thanks a lot!

PR #6554, opened 2021-01-06
Fix invalid comparison of string with number

TimvdLippe:
In this case, the `unit` is of type string. As such, `"codepoint" > 0`
always produces false, hence the code in question always became -1.
I found this while debugging https://crbug.com/1161083, which turns
out was already reported and fixed with
https://github.com/codemirror/CodeMirror/commit/37d7b2efceb192c94811a13b2b7b3eec4b786608


marijnh:
Thanks for spotting that, but the intention of that code was d19a746e, so I've pushed that instead.

PR #6560, opened 2021-01-11
fix fold issue: when JSON array contains json object, it will fold ob…

LuChenZhi:
When the JSON array contains a json object, such as bellow:
{
    "name": "hello",
    "test": [{
        "a": {
            "a": 123
        }
    }, {
        "b": {
            "b": 123
        }
    }]
}

When user click the fold icon, it will fold the first object in array:
{
    "name": "hello",
    "test": [{ --<-->--}, 
    {
        "b": {
            "b": 123
        }
    }]
}

But actually user wants to fold the array.
{
    "name": "hello",
    "test": [{ --<-->--}]
}

### DIFF
--- a/addon/fold/brace-fold.js
+++ b/addon/fold/brace-fold.js
@@ -31,10 +31,10 @@ CodeMirror.registerHelper("fold", "brace", function(cm, start) {
     }
   }
 
-  var startToken = "{", endToken = "}", startCh = findOpening("{");
+  var startToken = "[", endToken = "]", startCh = findOpening(startToken);
   if (startCh == null) {
-    startToken = "[", endToken = "]";
-    startCh = findOpening("[");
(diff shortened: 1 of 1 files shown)


marijnh:
This looks like it'll make the behavior of lines like `{foo: [` worse, though. Maybe the code should scan for both and use the one that appears earlier in the line?

LuChenZhi:
> This looks like it'll make the behavior of lines like `{foo: [` worse, though. Maybe the code should scan for both and use the one that appears earlier in the line?

Yes you are right. I missed this one you mentioned (`{foo: [`). We are using the js-beautify to reformat the json so we didn't meet this `{foo: [` / . And I think to fold the earlier one in line is perfect for most of the cases.

marijnh:
Attached patch implements that approach.

PR #6565, opened 2021-01-14
Add highlightNonMatching config to matchbracket addon

thiemowmde:
It's not uncommon to have non-matching brackets in MediaWiki wikitext source code. This is not always an error. Another reason why we want to disable this feature is that it's rather distracting while typing. We [temporarily patched our copy of the addon](https://gerrit.wikimedia.org/r/656119), but think it's worth having a proper config flag.

[Bug: T269096](https://phabricator.wikimedia.org/T269096)


marijnh:
Looks reasonable. Merged!

thiemowmde:
9 minutes. 😮 Amazing. Thanks a lot!

PR #6567, opened 2021-01-15
Replace regex with faster string matches

thiemowmde:
Warning: This works only when the match() call is a StringStream.match(), not if it is a native String.match(). Please carefully review this and make sure this doesn't break anything.

### DIFF
--- a/mode/soy/soy.js
+++ b/mode/soy/soy.js
@@ -411,7 +411,7 @@
             return null;
 
           case "list-literal":
-            if (stream.match(/\]/)) {
+            if (stream.match(']')) {
               state.soyState.pop();
               state.lookupVariables = true;
               popcontext(state);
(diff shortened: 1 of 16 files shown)


mode/soy/soy.js
```diff
@@ -517,14 +517,14 @@
             }
             return expression(stream, state);
           case "literal":
-            if (stream.match(/^(?=\{\/literal})/)) {
+            if (stream.match('{/literal}', false)) {
```
thiemowmde, line 520: Looking ahead via `(?=…)` does have the same effect as not consuming the match (therefor the `false`). The only benefit of this change is that it replaces a regex with a faster string match.

marijnh:
There's lots of nonsense code in the modes... I'm generally wary about such changes given that most modes don't even have tests, but I guess these do look pretty harmless. Merging, let's see if anyone reports a regression.

PR #6568, opened 2021-01-15
Add missing ^ anchors to StringStream.match() patterns

thiemowmde:
The method fails anyway (i.e. it acts as if there was no match) when the match is not at the start of the string. Adding the anchor does not change any behavior, but increases performance.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -3483,7 +3483,7 @@
         },
         isComplete: function(state) {
           if (state.nextCh === '#') {
-            var token = state.lineText.match(/#(\w+)/)[1];
+            var token = state.lineText.match(/^#(\w+)/)[1];
             if (token === 'endif') {
               if (state.forward && state.depth === 0) {
                 return true;
--- a/mode/cypher/cypher.js
+++ b/mode/cypher/cypher.js
@@ -21,11 +21,11 @@
     var tokenBase = function(stream/*, state*/) {
       var ch = stream.next();
       if (ch ==='"') {
-        stream.match(/.*?"/);
+        stream.match(/^[^"]*"/);
         return "string";
       }
       if (ch === "'") {
-        stream.match(/.*?'/);
+        stream.match(/^[^']*'/);
         return "string";
       }
       if (/[{}\(\),\.;\[\]]/.test(ch)) {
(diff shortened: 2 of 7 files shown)


keymap/vim.js
```diff
@@ -3483,7 +3483,7 @@
         },
         isComplete: function(state) {
           if (state.nextCh === '#') {
-            var token = state.lineText.match(/#(\w+)/)[1];
+            var token = state.lineText.match(/^#(\w+)/)[1];
```
thiemowmde, line 3486: Note: This is a native `String.match` call. However, we know the string must start with the `#` because of the `nextCh === '#'` check above. As all other changes, this one does not have any effect, but improves performance.

mode/cypher/cypher.js
```diff
@@ -21,11 +21,11 @@
     var tokenBase = function(stream/*, state*/) {
       var ch = stream.next();
       if (ch ==='"') {
-        stream.match(/.*?"/);
+        stream.match(/^[^"]*"/);
```
thiemowmde, line 24: Because the `.*` "catch all" was ungreedy, this was looking for the first `"` in the string. We know there can not be another `"` before. The new pattern does the same, but is faster because it avoids any backtracking.

mode/mscgen/mscgen.js
```diff
@@ -72,11 +72,11 @@
   CodeMirror.defineMIME("text/x-msgenny", {name: "mscgen", language: "msgenny"});
 
   function wordRegexpBoundary(pWords) {
-    return new RegExp("\\b(" + pWords.join("|") + ")\\b", "i");
+    return new RegExp("^\\b(?:" + pWords.join("|") + ")\\b", "i");
```
thiemowmde, line 75: I checked and no caller uses the `[1]` result. That's why it's fine to turn the brackets into a non-capturing group.

mode/php/php.js
```diff
@@ -53,7 +53,7 @@
           [["]", null]]
         ], closing, escapes);
       }
-      if (stream.match(/\-\>\w/, false)) {
+      if (stream.match(/^->\w/, false)) {
```
thiemowmde, line 56: No need to escape `>` or `-`, except they are in a special position, which they are not in this case.

PR #6572, opened 2021-01-19
Fix matchbrackets incompatibility with certain modes

thiemowmde:
When a mode does not apply a CSS class name to basic black text (i.e. the style is `""`), the matchbrackets addon will stop working for no good reason. See https://phabricator.wikimedia.org/T269096#6758178 for a longer explanation why I think this `|| null` is a mistake and should be removed.

This `|| null` was added in 2014 as part of commit 98326d16a1a9fc6e399c9e55bba2f544c1c70731. I can't find an explanation. I suspect it was a mistake, possibly done to make the code more fail-safe, but not considering the empty string.

### DIFF
--- a/addon/edit/matchbrackets.js
+++ b/addon/edit/matchbrackets.js
@@ -38,7 +38,7 @@
     if (config && config.strict && (dir > 0) != (pos == where.ch)) return null;
     var style = cm.getTokenTypeAt(Pos(where.line, pos + 1));
 
-    var found = scanForBracket(cm, Pos(where.line, pos + (dir > 0 ? 1 : 0)), dir, style || null, config);
+    var found = scanForBracket(cm, Pos(where.line, pos + (dir > 0 ? 1 : 0)), dir, style, config);
     if (found == null) return null;
     return {from: Pos(where.line, pos), to: found && found.pos,
             match: found && found.ch == match.charAt(0), forward: dir > 0};


marijnh:
Does changing line 71 from

            if (re.test(ch) && (style === undefined || cm.getTokenTypeAt(Pos(lineNo, pos + 1)) == style)) {

to

            if (re.test(ch) && (style === undefined || (cm.getTokenTypeAt(Pos(lineNo, pos + 1)) || null) == style)) {

also solve the problem for you? I suspect that was added to work around precisely this problem, but for modes which use `null` instead of the empty string for unstyled tokens.

thiemowmde:
Ah, I see. I had another look at the implementation of `getTokenTypeAt` and can see the code path for null now. I missed this before.

However, I think it does not make a difference. See, when calling `getTokenTypeAt` for the opening bracket returns an empty string, why should the same call for the closing bracket return null? This can only happen when a mode uses `""` and `null` the same time. Sure, we can change the patch to support this. Do you want to support such "broken" modes?

marijnh:
Yes, I do. They're not even really broken, and I don't want to break any code that worked before. Pushed b29c728 instead.

thiemowmde:
That's absolutely fine for me. Thanks a lot! Good idea to do the normalization for both sides in the same line of code.

PR #6581, opened 2021-02-04
[vim] fix html in dialogs/prompts

benhormann:
Fixes vim registers command when html tags present.
Fixes vim prompts in Firefox Developer Tools.

Sorry for the gnarly diff, but it had to be done.
I've also inlined the dialog function and searchPromptDesc constant var, which had exactly one reference each.

Background:
Firefox - the cmdline and search prompt has been broken in Firefox for a while, the `<input>` tag gets stripped out (when assigning to innerHTML in the dialog addon). Luckily constructing DOM nodes instead works.
`:registers` - This command prints out text you last copied/changed/deleted along with named/numbers registers. If you have html in your registers, the output is incorrect / interpreted.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -1450,7 +1450,7 @@
               showPrompt(cm, {
                   onClose: onPromptClose,
                   prefix: promptPrefix,
-                  desc: searchPromptDesc,
+                  desc: '(JavaScript regexp)',
                   onKeyUp: onPromptKeyUp,
                   onKeyDown: onPromptKeyDown
               });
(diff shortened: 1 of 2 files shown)


marijnh:
> the `<input>` tag gets stripped out 

Why? Is this some kind of elevated security setting active in the devtools? There's several other uses of `innerHTML` in the library (dialog and lint addons).

benhormann:
> Is this some kind of elevated security setting active in the devtools?

Must be, got this message in Browser Toolbox:
`Flattening unsafe node (descendants are preserved). Element: input.`

> There's several other uses of innerHTML in the library (dialog and lint addons).

True, but Firefox have implemented their own search dialog for default keymap. The only other case I found with an input field is Emacs "Goto line". I can do a PR for it later, or do you want it added here?

I didn't check which elements are *unsafe*, but I'd hazard a guess textarea, script and style would get flattened too.

marijnh:
Fair enough --- merged!

PR #6586, opened 2021-02-09
Chars consume possibility

murashki:
We can test stream with `match(..., false)`. That is, we gain knowledge about the next chars. I do so and i want to consume some next `n` chars. 

`stream.backup(-n)` looks bad.
`stream.pos += n` looks bad two...

May be it would more consistently if we add `consume` method...


marijnh:
I'm afraid you'll have to live with `.pos +=`—I don't want to further change the interface of StringStream.

PR #6587, opened 2021-02-10
[emacs] goto line in firefox devtools

benhormann:
Related to #6581. Firefox Developer Tools flattens input elements when using innerHTML.

The existing test didn't have any assertions. No problem, it does now.


marijnh:
Thanks for the patch!

Issue #5450, opened 2018-06-08
:g/^foo/s/$/bar/ gives "RangeError: Maximum call stack size exceeded"


takamori:
1. Go to https://codemirror.net/demo/vim.html
2. Type in the following two lines as its contents:

word: "a
word: "b

3. Execute the vi command :g/^word/s/$/"/

Result: A JS error occurs "RangeError: Maximum call stack size exceeded"

marijnh:
Should be solved via #6593

PR #6593, opened 2021-02-18
[vim] Fix :substitute and :global commands when splitting or joining lines

### LINKED ISSUE #6521: [vim] split one line into multiple lines
let's say I want to split one line
```
foo bar baz
```
into multiple:
```
foo
bar
baz
```
when typing `:s/ /\r/g` into codemirror, I get
```
foo
bar baz
```
but I get the following in actual vim
```
foo
bar
baz
```
how do I recreate vim's behavior in codemirror?

benhormann:
Fixes #6521 (and #5450).

The main issue was any change to line numbers wasn't taken into account. It's the bulk of this PR.

The other issue was the `index` variable being incremented (L5075) after the recursion (L5072). The test (L4073) was incorrect too.
*Note: a large enough number of matched lines will still cause a stack overflow error (as does recursive mappings and macros).*

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -5268,7 +5268,7 @@
       // Set up all the functions.
       cm.state.vim.exMode = true;
       var done = false;
-      var lastPos, modifiedLineNumber;
+      var lastPos, modifiedLineNumber, joined;
       function replaceAll() {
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks!

PR #6598, opened 2021-02-20
[vim] substitute pattern ending in dollar

### LINKED ISSUE #6580: Escaping is broken for $ during substitution in Vim
image
gives
image

As the case with Vim, `:%s/$/dollar` doesn't work either. However search using `/\$` works.

benhormann:
Fixes #6580

I seriously considered `if (regexPart === '$') {` since patterns like `/x$/` work better without the workaround (matches on last line).

I've opted to remove the useless calls to replace. Works fine as is and the flags ought to be saved, though they aren't currently.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -5097,9 +5097,11 @@
               regexPart = new RegExp(regexPart).source; //normalize not escaped characters
           }
           replacePart = tokens[1];
-          if (regexPart && regexPart[regexPart.length - 1] === '$') {
-            regexPart = regexPart.slice(0, regexPart.length - 1) + '\\n';
-            replacePart = replacePart ? replacePart + '\n' : '\n';
+          // If the pattern ends with $ (line boundary assertion), change $ to \n.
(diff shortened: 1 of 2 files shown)


marijnh:
Thank you.

(Would you, by any chance, be interested in collaborating on abstracting the editor integration in vim.js? Mostly as a way to adapt it to [CodeMirror 6](https://prosemirror.net/docs/ref/#view.EditorProps.clipboardSerializer), but possibly other editors would also be able to make use of it (ACE currently has a [fork](https://github.com/ajaxorg/ace/blob/a505922a01395ab77a878c08da06eb09b7822d23/lib/ace/keyboard/vim.js) of this file, I believe. I'd gladly take care of a lot of the work, but the code can be pretty gnarly in places, and I'm simply not knowledgeable enough about vim to understand what's going on sometimes.)

benhormann:
In short, yes. The challenge is to make it editor agnostic.

marijnh:
Thanks. I'm not entirely sure when I'll get to this, but I'll ping you when I need your input on something.

PR #6601, opened 2021-02-24
Allow the omission of head in setSelections

lishid:
This PR sets the default `head` to `anchor` in `setSelections()`, similar to `setSelection()`.
I think it should have been this way to ensure the API is consistent.

I'm more concerned about someone reading the latest docs but using an older version of CM that doesn't support this. Not sure how this is dealt with generally.


marijnh:
Good idea. The docs simply describe the latest version, that's not been a big problem so far.

PR #6605, opened 2021-02-26
fix #6604. Replace mousewheel/DOMMouseScroll by wheel

### LINKED ISSUE #6604: Replace mousewheel and DOMMouseScroll by wheel
Currently CodeMirror is using `mousewheel` and `DOMMouseScroll` instead of the `wheel` event.

https://github.com/codemirror/CodeMirror/blob/8bc57f76383e62e1a03c7d97c9eac74493fdbedc/src/edit/CodeMirror.js#L193-L195

See https://w3c.github.io/uievents/#events-wheelevents

karlcow:
The standard event implemented in all major browser is now wheel.

### DIFF
--- a/src/edit/CodeMirror.js
+++ b/src/edit/CodeMirror.js
@@ -191,8 +191,7 @@ function registerEventHandlers(cm) {
   })
 
   // Listen to wheel events in order to try and update the viewport on time.
-  on(d.scroller, "mousewheel", e => onScrollWheel(cm, e))
-  on(d.scroller, "DOMMouseScroll", e => onScrollWheel(cm, e))
+  on(d.scroller, "wheel", e => onScrollWheel(cm, e))
 
   // Prevent wrapper from ever scrolling
   on(d.wrapper, "scroll", () => d.wrapper.scrollTop = d.wrapper.scrollLeft = 0)


marijnh:
This library also targets older browsers, though. Is there any browser that doesn't fire mousewheel or DOMMouseScroll events?

karlcow:
@marijnh then Probably we can add the three events with `wheel` being first. 
Each time we leave just the usage of the non-standard properties, it means that browsers will never be able to deprecate the feature and will have to keep the feature implemented for ever. 

> Is there any browser that doesn't fire `mousewheel` or `DOMMouseScroll` events?

So the answer to your question is no, because it's impossible to remove them currently for web compatibility reasons. 😁 At Mozilla, we are trying to [get a usecounter for Firefox usage](https://bugzilla.mozilla.org/show_bug.cgi?id=1241602) and Chromium has already one [showing 4% usage for `mousewheel`](https://www.chromestatus.com/metrics/feature/timeline/popularity/1125) which is huge. 
 
Also according to the bugs on the chromium, `wheel` is slightly more performant. (I don't have hard data on that).

Do you want me to make a PR with the 3 values instead of replacing one by another.

marijnh:
If these can be reliable feature-detected, and there's no changes in the event object that'll confuse the existing handler code, then sure, a PR that prefers the standard event would be welcome.

karlcow:
> there's no changes in the event object that'll confuse the existing handler code

Are there tests already for this part of the code?

marijnh:
> Are there tests already for this part of the code?

I don't think so (you'd need to mock up an uncomfortable amount of context to effectively test wheel scrolling, since the code works by selectively allowing native behavior, and only stepping in when it needs to).

karlcow:
yes I have just discovered this part 
https://github.com/codemirror/CodeMirror/blob/250e4e959f022a864c5855cd87cf8b435bb38f6b/src/display/scroll_events.js#L7-L16

Issue #6312, opened 2020-06-10
vim cursor is not visible on empty lines in command mode


DrXyzzy:
Position the cursor on any empty line and hit escape. The cursor vanishes. The line must be completely empty, i.e. no whitespace or visible characters.

Seen with Chrome Version 83.0.4103.97 (Official Build) (64-bit) and Firefox 77.0.1 (64-bit) on macOS 10.15.5.

You can reproduce this in the Vim bindings demo at https://codemirror.net/demo/vim.html , which is at version 5.54.0 as of this writing.

ppasupat:
Possibly related: #6393 

I had the same issue and found the following:
* CodeMirror as of v5.59.0 has an issue where some non-mobile browsers might be detected as iOS and thus mobile (in `src/util/browser.js`). This is recently fixed in the master branch (See #6532).
* The default `inputStyle` option for mobile browsers is "contenteditable".
* The vim keymap uses a different method for showing the normal mode cursor (fat cursor) in contenteditable: it uses `doc.markText` to mark the current character. An empty line does not have any character, so the mark does not show up.

One workaround is to explicitly set the option `inputStyle: "textarea"` in the CodeMirror initializer. Another is to change the line in `codemirror.js` that says
```
var ios = !edge && /AppleWebKit/.test(userAgent) && (/Mobile\/\w+/.test(userAgent) || navigator.maxTouchPoints > 2);
```
to
```
var ios = safari && (/Mobile\/\w+/.test(userAgent) || navigator.maxTouchPoints > 2);
```

Issue #6393, opened 2020-08-21
[Bug] Vim mode + contenteditable doesn't show any caret if content is empty


thien-do:
Reproduce: go to https://codemirror.net/demo/vim.html and delete all content
Expectation: the fat caret is still shown at the first character position
Actual: no caret is shown

Note:
- If inputStyle is "textarea" instead then the caret is shown even if content is empty
- Related: https://github.com/codemirror/CodeMirror/issues/3552

ppasupat:
Possibly related: #6312

When `inputStyle` is "contenteditable", the vim keymap displays the normal mode cursor (fat cursor) by using `doc.markText` to mark the current character. An empty line does not have any character, so the mark does not show up.

PR #6607, opened 2021-03-01
Allow empty cursor for VIM in contenteditable

Billiam:
Quick change to allow a vim cursor on empty lines in contenteditable mode, by enabling `clearWhenEmpty: false` for the call to markText.

Fixes: #6312 and #6393


marijnh:
Won't this also remove the fat cursor on such lines in non-contenteditable mode?

Billiam:
Doesn't seem to? But it also isn't working the way I thought it was, so I'll close this.

PR #6610, opened 2021-03-02
[lua mode] use electricInput to automatically dedent

ianh:
It's inconvenient to have to manually dedent lines after typing `end` or `else`.  Note that `elseif` is also covered by this pattern, as `else` is a prefix of it.


marijnh:
Thanks!

PR #6612, opened 2021-03-03
[wast mode] Add support for GC proposal

jakobkummerow:
This adds support for the types and instructions added in the [GC proposal](https://github.com/WebAssembly/gc/blob/master/proposals/gc/MVP.md).

It also splits the definition of the giant regexp into multiple lines (instead of
a single 2,000-character line) to make it more maintainable.


bmeurer (approved):
Thanks for cleaning this up and making it easier to maintain!

Issue #6308, opened 2020-06-09
Mobile - scrolling / editing with inputStyle contenteditable jumps around


elsigh:
Here's a gif from one of the demo pages.

![Screen Recording 2020-06-09 at 7 33 30 AM](https://user-images.githubusercontent.com/517210/84162012-c2aec680-aa24-11ea-970b-83ab9efdae98.gif)

I've tried calling `preventDefault()` in a `scrollCursorIntoView` handler. 

When I set `inputStyle: textarea` the scrolling/focus behavior works as expected from a user experience point of view. However with `inputStyle: textarea` it seems we cannot longpress and select text. So kind of stuck between a rock and a hard place =)

CWKevo:
Can confirm that this happens too, using Chrome for Android.

rogdevil:
Hey, I have been assigned on gitpay to resolve this. Can you please assign it to me

rogdevil:
also I need a proper description on how to reproduce the issue it would be a great help

elsigh:
@rogdevil - you can repro like so: 
1) Grab an Android phone
2) Visit any of the demos, but go for https://codemirror.net/demo/html5complete.html
3) Tap into the code's container and then scroll the page down and try to tap into a part of the code and notice that you can't get your tap to put the cursor where you're tapping - sometimes it jumps on me, sometimes it puts the cursor into completely the wrong place, sometimes it scrolls the whole page somewhere else. It's just really hard / frustrating to get your cursor to a specifically tapped spot.

rogdevil:
yes, it does this sometimes. do this same happens on the code in the repo?

elsigh:
@rogdevil I don't understand your question, but you can reproduce this problem with this repository

rogdevil:
yeah sure we can make, I was dumb I didn't saw that ... 🤔

WhatMeansGithub:
This is still an issue. I have this issue on iOS when trying to tap or type or paste anything. It won't type where I want to but jumps somewhere randomly (usually top or middle of a command).

dragonDScript:
Hello there! Can I work on this issue? I have experience with CodeMirror.

PR #6621, opened 2021-03-15
fixxed the jumping cursor problem

rogdevil:
issue number
#6308 

description of the fix -
the problem was that the "touchend" and "touchmove" event was not working properly and causing the bug so i fixed then and a additional fix was that we have to add a inline style to the editor container which is
image

### DIFF
--- a/src/edit/CodeMirror.js
+++ b/src/edit/CodeMirror.js
@@ -146,7 +146,6 @@ function registerEventHandlers(cm) {
     return dx * dx + dy * dy > 20 * 20
   }
   on(d.scroller, "touchstart", function (e) {
-      // alert("touch start", e.touches)
       if (!signalDOMEvent(cm, e) && !isMouseLikeTouchEvent(e) && !clickInGutter(cm, e)) {
         d.input.ensurePolled()
         clearTimeout(touchFinished)
@@ -163,11 +162,7 @@ function registerEventHandlers(cm) {
       }
       finishTouch()
     })
-    // on(d.scroller, "touchmove", function () {
-    //   if (d.activeTouch) { d.activeTouch.moved = true }
-    // })
     on(d.scroller, "touchend", function (e) {
-      // alert("touch end")
       var touch = d.activeTouch
       if (touch && !eventInWidget(d, e) && touch.left != null && !touch.moved && new Date - touch.start < 300) {
         var pos = cm.coordsChar(d.activeTouch, "page"), range
--- a/src/edit/CodeMirror.js
+++ b/src/edit/CodeMirror.js
@@ -145,8 +145,7 @@ function registerEventHandlers(cm) {
     let dx = other.left - touch.left, dy = other.top - touch.top
     return dx * dx + dy * dy > 20 * 20
   }
-  on(d.scroller, "touchstart", function (e) {
-      // alert("touch start", e.touches)
+  on(d.scroller, "touchstart", (e) => {
       if (!signalDOMEvent(cm, e) && !isMouseLikeTouchEvent(e) && !clickInGutter(cm, e)) {
         d.input.ensurePolled()
         clearTimeout(touchFinished)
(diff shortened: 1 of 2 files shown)


src/edit/CodeMirror.js
```diff
@@ -145,40 +145,47 @@ function registerEventHandlers(cm) {
...
-      d.input.ensurePolled()
-      clearTimeout(touchFinished)
-      let now = +new Date
-      d.activeTouch = {start: now, moved: false,
-                       prev: now - prevTouch.end <= 300 ? prevTouch : null}
-      if (e.touches.length == 1) {
-        d.activeTouch.left = e.touches[0].pageX
-        d.activeTouch.top = e.touches[0].pageY
+  on(d.scroller, "touchstart", function (e) {
+      // alert("touch start", e.touches)
```
alexanmtz, line 149: Please remove comments

src/edit/CodeMirror.js
```diff
@@ -145,40 +145,47 @@ function registerEventHandlers(cm) {
...
-      cm.setSelection(range.anchor, range.head)
-      cm.focus()
-      e_preventDefault(e)
-    }
-    finishTouch()
-  })
-  on(d.scroller, "touchcancel", finishTouch)
+      finishTouch()
+    });
+    // on(d.scroller, "touchmove", function () {
```
alexanmtz, line 166: Removing this will not impact another scenario? Please make sure all the scenarios will continue to work.

src/edit/CodeMirror.js
```diff
@@ -145,40 +145,47 @@ function registerEventHandlers(cm) {
...
+    //   if (d.activeTouch) { d.activeTouch.moved = true; }
+    // });
+    on(d.scroller, "touchend", function (e) {
+      // alert("touch end")
```
alexanmtz, line 170: Same here, no need to send comments on the final code. Use the debug just for your local tests

alexanmtz (changes requested):
@rogdevil please remove all comments cause you shouldn't send comments of code in the final PR and please validate many scenarios as possible of the code, to be sure your changes will not affect the current behavior and verify the other examples.

alexanmtz:
@rogdevil please check the build failing for some code validation:
https://travis-ci.org/github/codemirror/CodeMirror/builds/763011194

```
src/edit/CodeMirror.js: Semicolon found (165:7)
src/edit/CodeMirror.js: Semicolon found (187:7)
```

rogdevil:
ok I have removed the comments ... i have tested the code very well and it does not cause any bugs at all :)

marijnh:
What I saw in the short time before I lost all patience was that you added a stray attribute to some demo, changed an arrow function to use `function` syntax, added a bunch of commented-out `alert` lines, and added two random spaces of indentation to several blocks of code. None those appear to be useful or and if you want me to review your patch, please make sure it only contains actual relevant changes.

rogdevil:
hey, i have made the fixes as you told, and also I didn't change any function to "function syntax" instead of arrow funtion syntax I have removed the alert statements. now about the changes that I made,

1. when I was investigating the bug the bug I found out that on android devices in chrome the event which is supposed to fire on "touchstart" was firing properly but the event "touchend" is not firing properly so I made some amends to the function firing on touch start by adding finishtouch() function to the end of touchstart event so that the corsor will be placed properly on the end of touch start event and we dont have to rely upon the touchend event to update the position of the cursor.
2. second thing as you know that "touchend" and "touchmove" events are causing the bug the touchend function fires multiple times on longpress event so the selection text was not working properly. in "touchend" event we have defined onlt 3 condition which are single tap, double tap and triple tap, as there is no condition which satisfies the longpress text selection event so we face problem on selecting the text thats why we are using "cm.setSelectText()" function so that our browser api auto detect the text selection behaviour and we don need to pass the starting point and anchor of the text selection like we were doing in the "cm.setSelection(range.anchor, range.head);" 
3. the third change that I made to was a demo html5 file so that we can reflect the changes made to script and disable the default behaviour of touch-action and execute the touch events that we have defined in the codemirror.js

I would love to know you thoughts on it .... :) @marijnh

marijnh:
> i have made the fixes as you told

That wasn't me, though.

> and also I didn't change any function to "function syntax" instead of arrow funtion syntax

I'm seeing this in the current patches:

```diff
-   on(d.scroller, "touchstart", e => {
+  on(d.scroller, "touchstart", function (e) {
```

Plus of course the whitespace changes and the commented-out lines are still there (seems your [last patch](https://github.com/codemirror/CodeMirror/pull/6621/commits/c9c598448200095079ed94d4897e4f5eff831b74) reintroduced those).

Issue #5622, opened 2018-10-17
Tern tooltip


ilevd:
When using custom tooltip with Tern and tooltip's height is big, then it could be "cropped" by border. It would be better if it changes its position in such cases:

![format3](https://user-images.githubusercontent.com/5524072/47059938-b34a8c00-d1f5-11e8-87ad-d4bffa22aaa3.png)

PR #6629, opened 2021-03-24
Fix: Tooltips can be off screen

nkappler:
Tooltips and the scrollbar inside the hint container can be off screen, which can induce scrollbars on the document body.

image
image
image

Only related bug I could find:
Fix: #5622

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -268,8 +268,10 @@
     var winH = parentWindow.innerHeight || Math.max(ownerDocument.body.offsetHeight, ownerDocument.documentElement.offsetHeight);
     container.appendChild(hints);
 
-    var box = completion.options.moveOnOverlap ? hints.getBoundingClientRect() : new DOMRect();
     var scrolls = completion.options.paddingForScrollbar ? hints.scrollHeight > hints.clientHeight + 1 : false;
+    if (scrolls) for (var node = hints.firstChild; node; node = node.nextSibling)
+      node.style.paddingRight = cm.display.nativeBarWidth + "px"
+    var box = completion.options.moveOnOverlap ? hints.getBoundingClientRect() : new DOMRect();
 
     // Compute in the timeout to avoid reflow on init
     var startScroll;
@@ -300,8 +302,6 @@
       }
       hints.style.left = (left = pos.left - overlapX - offsetLeft) + "px";
     }
-    if (scrolls) for (var node = hints.firstChild; node; node = node.nextSibling)
-      node.style.paddingRight = cm.display.nativeBarWidth + "px"
 
     cm.addKeyMap(this.keyMap = buildKeyMap(completion, {
       moveFocus: function(n, avoidWrap) { widget.changeActive(widget.selectedHint + n, avoidWrap); },
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -268,10 +268,8 @@
     var winH = parentWindow.innerHeight || Math.max(ownerDocument.body.offsetHeight, ownerDocument.documentElement.offsetHeight);
     container.appendChild(hints);
 
-    var scrolls = completion.options.paddingForScrollbar ? hints.scrollHeight > hints.clientHeight + 1 : false;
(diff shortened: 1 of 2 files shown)


nkappler:
The test seems to fail because of this line:
```javascript
    var winW = window.innerWidth || Math.max(ownerDocument.body.offsetWidth, ownerDocument.documentElement.offsetWidth);
    var winH = window.innerHeight || Math.max(ownerDocument.body.offsetHeight, ownerDocument.documentElement.offsetHeight);
```

Apparently using ownerDocument is not wanted?
I copied this line from show-hint.js so I am not sure why this is a problem.
I would be fine with just using `window.innerWidth`

Maybe one of the reviewers has more knowledge about this?

marijnh:
> Apparently using ownerDocument is not desired?

It's not defined—the linter is catching a reference to a variable that doesn't exist.

nkappler:
> It's not defined—the linter is catching a reference to a variable that doesn't exist.

Got it... would you recommend to add the variable 
```javascript 
var ownerDocument = cm.getInputField().ownerDocument;
```
or rather just use `window.innerWidth`?

marijnh:
I'm fine with just using the window size.

nkappler:
thanks for your help :) 
The fix is now complete in my opinion. Let me know if you have any open questions...

addon/hint/show-hint.js
```diff
@@ -268,8 +268,10 @@
     var winH = parentWindow.innerHeight || Math.max(ownerDocument.body.offsetHeight, ownerDocument.documentElement.offsetHeight);
     container.appendChild(hints);
 
-    var box = completion.options.moveOnOverlap ? hints.getBoundingClientRect() : new DOMRect();
     var scrolls = completion.options.paddingForScrollbar ? hints.scrollHeight > hints.clientHeight + 1 : false;
+    if (scrolls) for (var node = hints.firstChild; node; node = node.nextSibling)
+      node.style.paddingRight = cm.display.nativeBarWidth + "px"
```
marijnh, line 273: Why did you move this code? This ordering forces another relayout.

nkappler: Before my change the code would check if the hints container is off screen. If so, it would move it to be aligned with the edge of the screen. After that, it checks if the hints container has a scrollbar and if so, it applies padding to each hint element which causes the container to grow and be off screen again. (see attached gif)

I just changed the order so that the padding for the scrollbar is applied before it is checked if the container is off screen.

marijnh: Yes, but this code has been carefully built to _not_ create unnecessary relayouts, and I don't really want to accept a patch that degrades.

nkappler: I don't get why it should create another relayout, it just moves the container once in any case, or am I missing something?

marijnh: See for example [https://eloquentjavascript.net/14_dom.html#p_LcMNzqCMUS](this link). Your change causes a write to the DOM to happen between two layout reads, which will force an extra relayout.

nkappler: I see... I actually read your book but didn't remember that section.
Didn't think of that! 

I did some measurements and my change would have added around 10ms of rendering time.
I now have a more elegant solution: If there are scrollbars, the amount of padding added to the items is just added to the `overlapX`value, so no re-layout necessary anymore.
My measurements confirm that it adds no additional rendering time this way:

image
image

addon/hint/show-hint.js
```diff
@@ -293,6 +293,7 @@
       }
     }
     var overlapX = box.right - winW;
+    if (scrolls) overlapX += cm.display.nativeBarWidth;
```
nkappler, line 296: as mentioned, the amount of padding used in the hint items when there is a scrollbar is also added to the overlapX, moving the hints container just a smidge to the left, so the scrollbar is not off screen...

marijnh: Much better, thanks!

nkappler: thank YOU for building such awesome tools! :)

Issue #6630, opened 2021-03-26
Emacsy key bindings out of date


lishid:
https://github.com/codemirror/CodeMirror/blob/8bc57f76383e62e1a03c7d97c9eac74493fdbedc/src/input/keymap.js#L32

In modern Macs the `Alt+F/B/D` keys no longer perform the legacy emacs behavior.
Instead they insert the special characters `ƒ` `∫` and `∂`.

I'd know this has been previously reported, and I'm aware we can customize the behavior by deleting the binding at runtime, but I'd still recommend eliminating these from the default emacsy keymap since it's now the default OS behavior.

Related: https://github.com/codemirror/CodeMirror/issues/3254

marijnh:
A PR that moves those from `emacsy` to the proper emacs map in `keymap/emacs.js` would be very welcome.

PR #6631, opened 2021-03-26
Remove option-based emacsy key bindings

lishid:
Fixes https://github.com/codemirror/CodeMirror/issues/6630

Removes `Alt-D`, `Alt-F` and `Alt-B` from the default `emacsy` key bindings, which is enabled by default on macOS.

> A PR that moves those from `emacsy` to the proper emacs map in `keymap/emacs.js` would be very welcome.

It seems like these three are already implemented in `keymap/emacs.js`
https://github.com/codemirror/CodeMirror/blob/8bc57f76383e62e1a03c7d97c9eac74493fdbedc/keymap/emacs.js#L309
https://github.com/codemirror/CodeMirror/blob/8bc57f76383e62e1a03c7d97c9eac74493fdbedc/keymap/emacs.js#L311


marijnh:
Thank you!

PR #6643, opened 2021-04-06
[runmode] Pass the mode object to the callback

mtaran-google:
Without the mode object, it's not possible to call the mode's `indent` method to compute how much a given line should be indented (which we now have a need to do in a standalone way).


marijnh:
Looks reasonable. Merged.

PR #6646, opened 2021-04-08
Send new "updateGutter" signal on gutter resize

adamwight:
Some consumers need to know when the gutter width changes.  For
example, the MediaWiki Visual Editor maintains a transparent overlay
that must line up with the CodeMirror text area.  When line numbering
is enabled, the gutter will automatically resize as content grows
beyond 1,000 lines (and 10,000, etc.).  This shifts the text area,
and we need to recalculate position using
`display.gutters.offsetWidth`.

### DIFF
--- a/src/display/update_display.js
+++ b/src/display/update_display.js
@@ -5,6 +5,7 @@ import { displayHeight, displayWidth, getDimensions, paddingVert, scrollGap } fr
 import { mac, webkit } from "../util/browser.js"
 import { activeElt, removeChildren, contains } from "../util/dom.js"
 import { hasHandler, signal } from "../util/event.js"
+import { signalLater } from "../util/operation_group.js"
 import { indexOf } from "../util/misc.js"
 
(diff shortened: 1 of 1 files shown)


thiemowmde (approved):
Here is the relevant change in our code where we patched our copy of the CoreMirror 5 library to send this new signal: https://gerrit.wikimedia.org/r/618737.

Our questions are:
1. Is this a good place to send this signal?
2. Are there other things that can influence the position, width and height of the edit area?
3. Maybe there is already a signal for this? [The documentation](https://codemirror.net/doc/manual.html#events) mentions `viewportChange`, but this is something else, I believe.
4. Is this even the best possible approach?

marijnh:
That looks like the right place, but I'd use `signalLater` instead of `signal` to avoid reentrancy issues. And I think `"gutterChanged"` might be a more descriptive event name. Other than that this looks good.

marijnh:
You'll want to import `signalLater` from `util/operation_group.js`.

adamwight:
Thanks again!

PR #6651, opened 2021-04-10
Fix shadow root contenteditable cursor edit bug

aomarks:
Fixes bug where the cursor always jumps back to the first character of the line after each key input. This happens when using `contenteditable` mode within a shadow root.

#### Repro link

https://jsbin.com/jujuxux/edit?html,output

#### Explanation

CM5's `contenteditable` mode uses `document.activeElement` to access the current focus state of the editor and put the cursor in the correct place after an edit. However, when the input element is located within a [shadow root](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot), then `document.activeElement` is not the focused editor element itself, instead it is the host element in the document scope that *contains* the focused editor element. This causes some confusion that ends up placing the cursor back on the start of the line after each edit.

This PR replaces use of `document.activeElement` with the existing `activeElt` utility function, which traverses through shadow roots to address this issue, and is already used in all other modules that look for the active element.

#### Repro source

```html
<script type="module">
import 'https://unpkg.com/codemirror@5.60.0/lib/codemirror.js';

const shadowHost = document.createElement('div');
const shadowRoot = shadowHost.attachShadow({mode: "open"});
document.body.appendChild(shadowHost);

const scopedStyle = document.createElement('link');
scopedStyle.rel = 'stylesheet';
scopedStyle.href = 'https://unpkg.com/codemirror@5.60.0/lib/codemirror.css';
shadowRoot.appendChild(scopedStyle);

CodeMirror(shadowRoot, {
  inputStyle: 'contenteditable',
  value: `edit me`,
});
</script>
```

### DIFF
--- a/src/input/ContentEditableInput.js
+++ b/src/input/ContentEditableInput.js
@@ -10,7 +10,7 @@ import { simpleSelection } from "../model/selection.js"
 import { setSelection } from "../model/selection_updates.js"
 import { getBidiPartAt, getOrder } from "../util/bidi.js"
 import { android, chrome, gecko, ie_version } from "../util/browser.js"
-import { contains, range, removeChildrenAndAdd, selectInput } from "../util/dom.js"
+import { activeElt, contains, range, removeChildrenAndAdd, selectInput } from "../util/dom.js"
 import { on, signalDOMEvent } from "../util/event.js"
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks for looking into that. Merged!

PR #6652, opened 2021-04-11
Prevent line numbers from being voiced by screen readers

### LINKED ISSUE #6578: VoiceOver speaks line numbers
I have CodeMirror configured with `inputStyle="contenteditable"` and displaying line numbers and code folding.  On the Mac, when using VoiceOver in Chrome, the contents of the gutter (line number and disclosure) is voiced by the screenreader, which confuses users.  This happens when the entire page is being read, or when there is a multiline selection inside of the editing area.

FWIW, I do not have this problem in Chrome when using the NVDA screen reader on Windows.

As a workaround, I'm hiding the gutter area from screen readers by setting `aria-hidden="true"` on `.CodeMirror-gutter-wrapper` elements, but that feels like a hack.

aomarks:
Prevent line numbers from being voiced by screen readers.

Sets `aria-hidden=true` on `.CodeMirror-gutter-wrapper` elements.

Tested with Voice Over on Chrome.

Fixes #6578


marijnh:
Thanks! I could have sworn we were already doing this, but it appears there wasn't anything like that in the code.

Issue #5020, opened 2017-10-11
getting the pc/mac version of emacs & sublime keymaps at runtime


mtaran-google:
I'd like to be able to get the pc and mac versions of the emacs and sublime keymaps programmatically. Currently, they are written such that they each only produce a single keymap based on the sniffed OS at load time. I'd like to send a PR that changes them to

1. create `macEmacs`/`macSublime` and `pcEmacs`/`pcSublime` keymaps, as well as the current `sublime` and `emacs` ones
1. make the prefixless one point to either the mac or pc version (like you do with `default` + `macDefault`/`pcDefault`)

Would this be acceptable? Or do you have a preferred alternative method for accomplishing this sort of thing?

marijnh:
That sounds like a good idea, yes.

adrianheine:
Fixed in #5022, right?

PR #6659, opened 2021-04-19
Expose emacs commands for easier programmatic access

jpredham:
This change is related to the change in [this commit](https://github.com/codemirror/CodeMirror/commit/1951460e52e2ec1d3615b5240123fe3c284d2e17) and discussed in #5020, except for emacs instead of sublime. It simply exposes the functionality that lived in anonymous functions behind commands that are named according to their corresponding emacs functions (except where they conflict with other commands). No logic was changed - just moved around.

### DIFF
--- a/keymap/emacs.js
+++ b/keymap/emacs.js
@@ -271,11 +271,15 @@
     }
   }
 
-  // Commands. Names should match emacs function names wherever possible.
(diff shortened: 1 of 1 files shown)


marijnh:
Looks good. Merged!

jpredham:
Thanks @marijnh .

PR #6677, opened 2021-05-07
Add initial port of abbott.vim v2.1 theme

bcat:
I've ported the [abbott.vim color scheme](https://github.com/bcat/abbott.vim) (version 2.1) to a CodeMirror theme. It's as faithful a port as I could reasonably make given that CodeMirror's syntax highlighting tokens don't perfectly map onto Vim's highlight groups.

I saw the note in the pull request template to not submit a PR for a new mode or addon. I wasn't sure if the same applied to themes, so if y'all aren't merging new themes upstream either, please just let me know. Thanks! :)


marijnh:
Themes are a bit low-maintenance than modes, so I guess I'm okay with merging this.

PR #6681, opened 2021-05-12
Fixes the syntax highlighting of TypeScript that broke after the exclamation mark

romfrolov:
Syntax highlighting of TypeScript broke after encountering the first exclamation mark (Non-Null Assertion Operator).

Before:
Screen Shot 2021-05-12 at 4 40 21 PM

After:
Screen Shot 2021-05-12 at 5 03 37 PM


eliasgroll:
Awesome :) waiting for this!

PR #6682, opened 2021-05-12
[core] fix n^2 code path in marking text

mtaran-google:
This caused some unnecessary slowdowns and GCs when adding many (thousands) of text markers. I looked around and couldn't find anything relying on new versions of the markedSpans array needing to be referentially different, but I could have missed something.

### DIFF
--- a/src/line/spans.js
+++ b/src/line/spans.js
@@ -28,7 +28,11 @@ export function removeMarkedSpan(spans, span) {
 }
 // Add a span to a line.
 export function addMarkedSpan(line, span) {
-  line.markedSpans = line.markedSpans ? line.markedSpans.concat([span]) : [span]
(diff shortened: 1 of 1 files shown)


marijnh:
I think the history code that handles undoing marks does rely on immutable span arrays.

I suspect a lot of efficiency could be gained by adding a mass-mark-adding interface. Would that help in your case?

mtaran-google:
Yes, we'd definitely use that! We already wrap things in an operation, but a batch method would be even better.

marijnh:
On closer look, the interface doesn't really allow for a very elegant way to define a mass-add method, so attached patch just optimizes adding spans within a transaction. Let me know whether it fixes your slowdown.

Issue #6683, opened 2021-05-13
Perl : Number with underscore as separator are not highlighted properly


XRFXLP:
Almost the same issue as https://github.com/codemirror/CodeMirror/issues/6616

![image](https://user-images.githubusercontent.com/23459568/118092941-5da8e100-b3ea-11eb-8974-69d041f7d875.png)

PR #6684, opened 2021-05-13
Update perl.js

XRFXLP:
Ignore any underscores which lies in middle or end of string, this might fix https://github.com/codemirror/CodeMirror/issues/6683


marijnh:
This change looks like it has never actually been ran. `stream` is a  `StringStream` object, which doesn't have a `replace` method. The regexp also isn't valid in JavaScript.

PR #6690, opened 2021-05-19
fix: [lint] exits early if required state is not present

dimitropoulos:
Over on https://github.com/Kong/insomnia/, we've been getting this error on codemirror.  It appears to be (ultimately) due to a missing check for `state.lint` being present.  Sometimes it's there, sometimes it isn't.

Screenshot_20210519_150136

This PR fixes this bug.

### DIFF
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -140,7 +140,9 @@
   }
 
   function startLinting(cm) {
-    var state = cm.state.lint, options = state.options;
+    var state = cm.state.lint;
+    if (!state) return;
+    var options = state.options;
     /*
      * Passing rules in `options` property prevents JSHint (and other linters) from complaining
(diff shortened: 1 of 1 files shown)


marijnh:
I'm curious how you get `startLinting` to run when linting is disabled—it should only be called from event handlers that are enabled by the lint option. Do you have an example of how to trigger this error?

dimitropoulos:
I've been trying to answer that question, too.  I'm not very experienced with CodeMirror, personally, so all I know is that this is possible to do repeatedly/reliably and that the PR implements a fix.

You can view the code here: https://github.com/Kong/insomnia/tree/develop/packages/insomnia-app/app/ui/components/codemirror for our codemirror implementation.

Additionally, you can reproduce this by running insomnia (dev instructions are on the repo's main readme) and opening any codemirror instance (e.g. click on the first/default collection on the first page when the app starts).

PR #6701, opened 2021-06-07
make getRange(a,b,'') work as expected

stasoid:
Currently `getRange(a,b,'')` works the same as `getRange(a,b)`.


marijnh:
Good catch. Merged!

PR #6704, opened 2021-06-08
feat: Highlight a row when lint caused error

lexer2086:
When linter shows errors, there is no way to highlight the whole row
I'm able to set icon only
image

I would like to have a possibility to highlight the whole row, like:
image

### DIFF
--- a/addon/lint/lint.css
+++ b/addon/lint/lint.css
@@ -69,3 +69,7 @@
   background-position: right bottom;
   width: 100%; height: 100%;
 }
+
+.CodeMirror-lint-row-error {
+  background-color: rgba(183, 76, 81, 0.08);
+}
--- a/addon/lint/lint.css
+++ b/addon/lint/lint.css
@@ -70,6 +70,6 @@
   width: 100%; height: 100%;
 }
 
-.CodeMirror-lint-row-error {
+.CodeMirror-lint-line-error {
   background-color: rgba(183, 76, 81, 0.08);
 }
--- a/addon/lint/lint.js
+++ b/addon/lint/lint.js
@@ -11,6 +11,7 @@
 })(function(CodeMirror) {
   "use strict";
   var GUTTER_ID = "CodeMirror-lint-markers";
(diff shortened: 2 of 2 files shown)


marijnh:
Good idea. Could you:

 - Make this an option that's off by default, so that it doesn't change the behavior of existing setups
 - Use 'line' rather than 'row' in the naming
 - Also enable this behavior when there is no lint gutter

lexer2086:
> Good idea. Could you:
> 
> * Make this an option that's off by default, so that it doesn't change the behavior of existing setups
> * Use 'line' rather than 'row' in the naming
> * Also enable this behavior when there is no lint gutter
If there any way to pass params into lint, like: `lint: {highlightErrorLine: true}` instead of just `lint: true`?
```
{
    lineNumbers: true,
    mode: "javascript",
    gutters: ["CodeMirror-lint-markers"],
    lint: true
  }
```
If I put any params into lint like:
```
{
    lineNumbers: true,
    mode: "javascript",
    gutters: ["CodeMirror-lint-markers"],
    lint: {
      highlightErrorLine: true
   }
}
```

I got warning in console:
```
Cannot display JSHint error (invalid line 0)
Reason: Bad option: 'onUpdateLinting'
```
Some problem with JSHINT

marijnh:
I'm not sure what is happening there, or how JSHint can be doing any kind of type checking at all, but that problem seems outside of the scope of this pull request.

marijnh:
Why do you check for `state.marked.length` in `isHighlightErrorLinesEnabled`?

To iterate over all lines, use `cm.firstLine()` and `cm.lastLine()`. Lines don't necessarily start at 0 and `doc.size` isn't part of the public interface.

lexer2086:
I was in a hurry and didn't see `state.marked.length` left. Actually no need to check for my purpose.
Also I've read about `cm.firstLine()` and realised that I didn't check how is my functionality works there

marijnh:
Thanks. I've merged this and followed up with another patch which changes the option name to `highlightLines` (since it highlights warnings as well as errors) and makes some small tweaks.

PR #6714, opened 2021-06-18
Changed Cursor color in some themes 

ps173:
Changed the cursor color in some theme styles. Changes were made in ayu-dark, ayu-mirage, gruvbox-dark, base16-dark, material-palenight, material , material ocean and oceanic-next. Will be adding more changes for other respective themes

Edit: I am surprised to see the number of changes. Cause I didn't change this much and realized that this was a formatting thing.

### DIFF
--- a/theme/ayu-dark.css
+++ b/theme/ayu-dark.css
@@ -40,3 +40,10 @@
   text-decoration: underline;
   color: white !important;
 }
+
+/* changing cursor color in vim mode */
+.CodeMirror-cursor { background-color: #a2a8a175 !important; }
+
+.CodeMirror-cursors { background-color: #a2a8a175 !important; }
+
+.cm-animate-fat-cursor { background-color: #a2a8a175 !important; }
--- a/theme/ayu-dark.css
+++ b/theme/ayu-dark.css
@@ -42,8 +42,5 @@
 }
 
 /* changing cursor color in vim mode */
-.CodeMirror-cursor { background-color: #a2a8a175 !important; }
-
-.CodeMirror-cursors { background-color: #a2a8a175 !important; }
-
-.cm-animate-fat-cursor { background-color: #a2a8a175 !important; }
+.cm-s-ayu-dark .CodeMirror-cursor { background-color: #a2a8a175 !important; }
+.cm-s-ayu-dark .cm-animate-fat-cursor { background-color: #a2a8a175 !important; }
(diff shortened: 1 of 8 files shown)


ps173:
Also apologies for changing this many files.

marijnh:
Sorry, no, I don't want a bunch of changes that apply some automatic formatting to files. The patches should contain only the necessary changes, nothing else.

ps173:
I am really sorry for being that careless. I totally get your point. I have made the necessary changes now can u look at it.

marijnh:
Thanks for cleaning that up. The next problem is that you're not prefixing the new rules with the theme's class, which will cause it to apply when you _load_ the theme, rather than when you _enable_ it.

Also, it looks like these themes already define a cursor color via a rule [like this](https://github.com/codemirror/CodeMirror/pull/6714/files#diff-48a43418e67bafc8a9784a8c94d9e6b3a052276a202b249d3791e7448cdf5902L11). Why are you adding a background, when the cursor is styled with a border? Giving a background to `.CodeMirror-cursors` is definitely useless (this is a zero-height element that acts as parent node for the cursors).

ps173:
> Giving a background to .CodeMirror-cursors is definitely useless (this is a zero-height element that acts as parent node for the cursors).    

Yup you are right about that I will be removing that. 

> Also, it looks like these themes already define a cursor color via a rule

Yes I did see that and to prevent breaking of any previous code. I made a new class just to be safe

Also the the cursor is styled with border in default keymap but in vim mode it has a background color and no borders. 
Background Color

In above screenshot you can clearly see that it has a background color is present in cursor

ps173:
> Thanks for cleaning that up. The next problem is that you're not prefixing the new rules with the theme's class, which will cause it to apply when you load the theme, rather than when you enable it.

Oh yeah I didn't notice that. Thanks I will be fixing that too

ps173:
I have prefixed the theme classes and removed the `.CodeMirror-cursors` from the code. Can u check it now. Also apologies for late replies

marijnh:
> Also the the cursor is styled with border in default keymap but in vim mode it has a background color and no borders.

The background is coming from the fat-cursor class. So why add one to `.CodeMirror-cursor`?

ps173:
so do I just change `fat-cursor` class? I went for `.CodeMirror-cursor`  as it was shown in dev-tools

marijnh:
> so do I just change fat-cursor class? I went for .CodeMirror-cursor as it was shown in dev-tools

No, you have to add it as a parent rule. I've done so in 4c6759e1 (with your patches merged as bac5a2d37f56)

PR #6721, opened 2021-06-23
COBOL:  highlight single or double quote inside single quotes as closed string

### LINKED ISSUE #6680: COBOL: Quotation marks wrong highlighting in the COBOL mode
**Actual result:**  all text after '"' or between '"' and ''' is highlighted.
**Expected result**: only '"' or ''' is highlighted.

You can check this behavior [here](https://codemirror.net/mode/cobol/index.html).

Screenshot 2021-05-11 at 11 20 37

Screenshot 2021-05-11 at 11 20 30

KQLYYY:
closes #6680 

**After fix:** _(line 9, 13)_
Screenshot 2021-06-23 at 11 06 13

### DIFF
--- a/mode/cobol/cobol.js
+++ b/mode/cobol/cobol.js
@@ -195,8 +195,7 @@ CodeMirror.defineMode("cobol", function () {
       case "string": // multi-line string parsing mode
         var next = false;
         while ((next = stream.next()) != null) {
-          const nextForNext =  stream.string[stream.pos]
-          if ((next == "\"" || next == "\'") && nextForNext !== "\'" && nextForNext !== "\"") {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks. The modes are written in IE8-compatible ES3, so you can't use const. Also, `!stream.match(/['"]/, false)` is a simpler way to do the lookahead.

KQLYYY:
> Thanks. The modes are written in IE8-compatible ES3, so you can't use const. Also, `!stream.match(/['"]/, false)` is a simpler way to do the lookahead.

Thank you for your help, I agree with you, all the necessary changes have been made.

marijnh:
Merged!

PR #6722, opened 2021-06-24
Fix #6660. Created regex to match all uppercase ascii/cyrillic chars

### LINKED ISSUE #6660: isUpperCase and isLowerCase tests for vim mode are ascii-only
https://github.com/codemirror/CodeMirror/blob/dc3952a60b6cf8ea7b2cb66521ae5165a04d444e/keymap/vim.js#L427-L429

This causes some vim commands, e.g. `~` key (case switching) to work incorrectly.

Steps to reproduce:
1. Enable Vim mode.
2. Enter some text in mixed case in cyrillic, e.g. “строчные ЗАГЛАВНЫЕ”.
3. Move cursor to the first letter and press “~” a few times.
4. Notice that smallcase letters turn into uppercase, but uppercase letters don’t change.

WillHernandez:
https://github.com/codemirror/CodeMirror/issues/6660
Created a regex that will match  all uppercase ascii/cyrillic characters. This will improve the accuracy of function isUpperCase() and allow vim shortcut "shift + ~" to convert lowerCaseChars => upperCaseChars & upperCaseChars => lowerCaseChars on cyrillic characters.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -910,6 +910,7 @@ Wesley Wiser
 Weston Ruter
 Will Binns-Smith
 Will Dean
+Will Hernandez
 William Desportes
 William Jamieson
 William Stein
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -420,6 +420,7 @@
     var numbers = makeKeyRange(48, 10);
     var validMarks = [].concat(upperCaseAlphabet, lowerCaseAlphabet, numbers, ['<', '>']);
     var validRegisters = [].concat(upperCaseAlphabet, lowerCaseAlphabet, numbers, ['-', '"', '.', ':', '_', '/']);
+    var upperCaseChars;
 
     function isLine(cm, line) {
       return line >= cm.firstLine() && line <= cm.lastLine();
@@ -434,7 +435,12 @@
       return numberRegex.test(k);
     }
     function isUpperCase(k) {
-      return (/^[A-Z]$/).test(k);
+      try {
+        upperCaseChars = new RegExp("^[\\p{Lu}]$", "u");
+      } catch (_) {
+        upperCaseChars = /^[A-Z]$/;
+      }
+      return upperCaseChars.test(k);
     }
     function isWhiteSpaceString(k) {
       return (/^\s*$/).test(k);


marijnh:
Some of the browsers supported by the library don't support `\p` (or `/u` for that matter) so you'll have to do this conditionally and fall back to the old regexp when it doesn't work.

Also, why `new RegExp(/^[\p{Lu}]$/, 'u')`, and not just `/^[\p{Lu}]$/u`?

WillHernandez:
> Some of the browsers supported by the library don't support `\p` (or `/u` for that matter) so you'll have to do this conditionally and fall back to the old regexp when it doesn't work.
> 
I should've recognized that as a potential issue beyond the browsers that I tested.
A more accurate solution that does not require a regex may be to bring back the original regex to check for uppercase ascii and also check the indexof the below upperCase cyrillic chars string. Ill make sure im not missing any additional chars in that string.

var upperCaseCyrillicChars = "АбБВГДЕЁЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ";
upperCaseCyrillicChars.indexOf(k) !== -1

> Also, why `new RegExp(/^[\p{Lu}]$/, 'u')`, and not just `/^[\p{Lu}]$/u`?
For some reason when running npm test - the later regex would return a SyntaxError: invalid regular expression flag but new Regexp would not.

WillHernandez:
> > For some reason when running npm test - the later regex would return a SyntaxError: invalid regular expression flag but new Regexp would not.
> 
> That's a symptom of the engine support issue I was talking about.
> 
> I'd suggest something like:
> 
> ```
> var upperCaseChars
> try { upperCaseChars = new RegExp("^[\\p{Lu}]$", "u") }
> catch (_) { upperCaseChars = /^[A-Z]$/ }
> ```

Thanks for the suggestions. Added try/catch to fallback to the original regex.

marijnh:
return new RegExp(/^[\p{Lu}]$/, "u").test(k);

Could you make the argument a string rather than a regexp? I guess this, due to subtle type conversions, also works, but it's needlessly obscure. (Don't forget to escape the backslash in the string.)

marijnh:
Also since try-catch isn't necessarily cheap I'd really put the regexp in a module-level variable.

marijnh:
The exception occurs on regexp creation, not when calling a method on it, so this doesn't work. Please just follow the structure I showed before.

WillHernandez:
> The exception occurs on regexp creation, not when calling a method on it, so this doesn't work. Please just follow the structure I showed before.

Looks to be set as requested. Apologies on any confusion

PR #6728, opened 2021-06-29
Add python 3 string prefix `rb` and `rf`

### LINKED ISSUE #6727: Python string prefix `rf` and `rb` does not get highlighted
In python, [multiple string prefix such as  `rf` and `rb` is valid](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals):

```python
rf"hi"
fr"hi"
rb"hi"
br"hi"
```

however codemirror currently does not highlight `rf` and `rb` (it does, however highlight `fr` and `br` correctly).

image

https://jsbin.com/xowoginoqo/1/edit?html,js,output

qria:
Added `rb` and `rf` to python highlighter.

Only added to python 3 because there's no f-string in python2.

Now we support all python string prefix. See : https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

fixes #6727


marijnh:
Thanks!

PR #6740, opened 2021-07-09
Improve folding addon performance

Billiam:
A couple of perforance improvements for the fold code addon

## Don't use scanUp for some fold commands

For the `foldAll` and `unfoldAll` commands, the fold addon already iterates through every line in the file, so a secondary loop upward, for every line, is wasteful.

For the `unfold` command, `doFold` exits after `getRange` anyway, and depends on the marker clearing side effect in `getRange` for unfolding. When scanning up, these ranges are not unfolded anyway because of the `allowFolded` check, so having scanUp enabled just causes a bunch of extra `findMarksAt` calls for no reason.

## Exit `getRange` early if `force` is "fold"

`getRange` iterates over all marks at a position and then checks for `force === "fold"` at each loop iteration. This change exits before the loop instead, saving the call to `getMarksAt` and the iteration.

### DIFF
--- a/addon/fold/foldcode.js
+++ b/addon/fold/foldcode.js
@@ -106,13 +106,13 @@
   CodeMirror.commands.foldAll = function(cm) {
     cm.operation(function() {
       for (var i = cm.firstLine(), e = cm.lastLine(); i <= e; i++)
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks!

PR #6741, opened 2021-07-09
Folding: Allow custom textMarkers

Billiam:
Adds a `marker` option to `foldOptions`, to complement the existing `widget` option.

The `marker` value is expected to be a function that accepts a start and end range, a widget, and boolean for `clearOnEnter`, and returns a TextMarker.

This allows customization of the text marker for inclusive/select left/right properties, custom clear behavior, overriding `clearWhenEmpty`, and can be used as an entry point for code that should be run before adding a fold.

### DIFF
--- a/addon/fold/foldcode.js
+++ b/addon/fold/foldcode.js
@@ -47,11 +47,7 @@
       myRange.clear();
       CodeMirror.e_preventDefault(e);
     });
-    var myRange = cm.markText(range.from, range.to, {
-      replacedWith: myWidget,
-      clearOnEnter: getOption(cm, options, "clearOnEnter"),
-      __isFold: true
-    });
+    var myRange = makeMarker(cm, options, range, myWidget);
     myRange.on("clear", function(from, to) {
       CodeMirror.signal(cm, "unfold", cm, from, to);
     });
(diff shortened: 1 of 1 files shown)


marijnh:
I'm not a fan of this—I don't see any reason to ever change the inclusivity of a fold marker, and adding imperative hooks when creating a marker probably won't end well. What's the specific use case that prompted this patch?

Billiam:
> What's the specific use case that prompted this patch?

Mostly, I needed an immediately-before-added and after-removed hook, and I thought that making the marker available outside of the plugin was a nice bonus.

My use case isn't common, but I've added persistent folding to markdown documents, so that any heading ending with `...` is considered "folded". To do that, I'm adding document content when the fold is created (that will get folded), and removing it when the fold is cleared.

The easiest way to do that was to change the content and alter the fold range at the last second, and to add a custom clear handler to the marker to remove that content. I may be misremembering but the unfold signal occured too late for my use.

marijnh:
Don't the `"fold"` and `"unfold"` events emitted by the foldcode addon provide a more solid way to do something like that?

marijnh:
I see, you need to change the document as a fold is taking place. Have you considered creating your own command that handles folding/unfolding, so that you can first adjust the document and _then_ fire off a fold? Working with changing coordinate systems is unfortunately not very convenient in CodeMirror 5.x.

Billiam:
Thanks. It doesn't sound like this PR is a good match, so I'll close.

PR #6746, opened 2021-07-20
[search addon] Make compatible with Trusted Types

vrana:
https://web.dev/trusted-types/ doesn't allow string innerHTML assignments. This change avoids it by using DOM instead.

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -189,18 +189,50 @@
     if (state.annotate) { state.annotate.clear(); state.annotate = null; }
   });}
 
+  function el(tag, attrs) {
+    var element = document.createElement(tag);
+    for (var key in attrs) {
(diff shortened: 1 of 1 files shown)


marijnh:
This looks good, in general, but would you be open to adding a feature to `el` where you can pass children as extra arguments, so that the code gets a bit less verbose? Something like

```
for (let i = 2; i < arguments.length; i++) {
  let child = arguments[i]
  element.appendChild(typeof child == "string" ? document.createTextNode(child) : child)
}
```

vrana:
Done.

marijnh:
Merged as 68d3399294abd37, updated in 336217837414a81 (there was an undefined reference to `dialogText` in the replace code)

vrana:
Thanks.

PR #6749, opened 2021-07-21
Fix passing in Lint options that are not provided as a default.

### LINKED ISSUE #6748: Unable to provide lint.getAnnotations and lint.async config
The getAnnotations and async properties are dropped from the lint addon config. When constructing the options object in LintState, the properties are ignored because they are not represent in the defaults.

jtangelder:
Fixes #6748

Change the hasOwnProperty check to check the conf object instead of the defaults.


marijnh:
See patch 5180d1e0edc993 instead. This one would break the intent of that code (splitting properties that aren't defined for the addon itself into the `linterOptions` object).

PR #6752, opened 2021-07-23
Add Delta and Delta Live Tables keywords

sandra-sandeep:
This patch adds keywords for Delta and Delta Live Tables.

### DIFF
--- a/mode/sql/sql.js
+++ b/mode/sql/sql.js
@@ -473,7 +473,7 @@ CodeMirror.defineMode("sql", function(config, parserConfig) {
   });
 
   // Databricks
-  CodeMirror.defineMIME("text/x-databricks", {
+  CodeMirror.defineMIME("text/x-databrickssql", {
     name: "sql",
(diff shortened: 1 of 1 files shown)


marijnh:
I think this may be better kept in your own codebase—I haven't heard of Delta SQL before, and I'm hesitant to maintain support for things I don't know anything about.

marmbrus:
@marijnh thanks for your time reviewing this PR!

[Delta](https://delta.io/) is an open source project that is commonly used to extend Spark SQL to support transactions and DDL like `UPDATE`/`DELETE`/`MERGE`. There are already a few Delta specific keywords in the Spark SQL dialect (i.e. `OPTIMIZE`), and we just wanted to expand that to keep pace with the project.

I think the though here was that it would be cleaner to create a separate dialect as we add these, but I think it would also be okay to just add them to Spark SQL.  However, @sandra-sandeep, if we do that it should be a superset of Spark SQL, not it own separate dialect. And I would call this `Delta` not `Databricks` SQL if we do.

We can also just do this in our own codebase, but we thought this might be something the community would appreciate.

marijnh:
I think—since this is a two-liner (plus some definition of `set`)—it's reasonable for you to keep it in your own codebase.

PR #6764, opened 2021-08-13
set attribute [translate="no"]

aokiken:
I think don't need to translate programming code.

ref https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate

### DIFF
--- a/src/display/Display.js
+++ b/src/display/Display.js
@@ -49,6 +49,10 @@ export function Display(place, doc, input, options) {
   // The element in which the editor lives.
   d.wrapper = elt("div", [d.scrollbarFiller, d.gutterFiller, d.scroller], "CodeMirror")
 
+  // This attribute is respected by automatic translation systems such as Google Translate,
+  // and may also be respected by tools used by human translators.
+  d.wrapper.setAttribute('translate', 'no')
+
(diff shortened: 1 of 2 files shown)


marijnh:
It doesn't look like you've tested that code—you're calling `setAttribute` on a `Display` object, not a DOM element.

It would probably make more sense to set it on the input field anyway. Which would be implemented in `TextareaInput` (if translators translate textarea content at all) and `ContentEditableInput`.

aokiken:
Thanks your feedback.

I'll check CONTRIBUTING.md.
https://github.com/codemirror/CodeMirror/blob/master/CONTRIBUTING.md

aokiken:
I looking at https://github.com/codemirror/CodeMirror/commit/0b00c0e12fe919865929053e03b2c9dccda0779c Because I think this time changed seems to be similar to this.

aokiken:
@marijnh 
> It doesn't look like you've tested that code—you're calling setAttribute on a Display object, not a DOM element.
Changed the target from Display to Display.wrapper

> Which would be implemented in TextareaInput (if translators translate textarea content at all) and ContentEditableInput.

I looking at 0b00c0e Because I think this time changed seems to be similar to this.

Programming code inside `[role="presentation"]` does not need to be translated.

Issue #3256, opened 2015-05-12
[Vim] Cursor width too small when use keycap Vim


DelightRun:
Hello, everyone!
I use my code mirror with `keymap Vim`, and when on `Vim Mode`, **my little green cursor's width is just 3.5px(in demo, is 9px)**. So I don't know what's going wrong, please help me.

And here is a screen snap. 
![cut](https://cloud.githubusercontent.com/assets/4102875/7592927/fce4252a-f908-11e4-9eaa-f3e53a181ecc.png)

mightyguava:
Hmm, either the CSS is messed up, or vim is trying to make the cursor 0 characters wide. I can't tell from just the picture. Do you have a reproducible example or a link to whatever you took the screenshot of?

DelightRun:
Of course, the page is on http://changxu.wang/test.html.
Oh, by the way, my browser is Safari with Mac OS X 10.10, and when I use Chrome, the problem disappear.
But the offical demo is OK with my Safari

mightyguava:
Here's the bug:

The cursor is just a space rendered over text. It renders to the same width as the text because CodeMirror uses the default monospace font in the browser.

Text inside your CodeMirror instance is rendering with a non-monospace font. This is pretty obvious if you type something like `lalalalala`. Given that the cursor is a space, it gets a super thin space.

The non-monospace font seems to be caused by the `lang="zh-CN` in your `<html>` tag. I guess whatever font Safari uses for rendering Chinese monospace text isn't actually monospace.

My suggested fix is to have a CSS rule like the below to override the monospace font to whatever you want it to be.

```css
.code-editor-codemirror .CodeMirror {
  font-family: Courier;
}
```

@marijnh do you have any thoughts on a fix we can do on our side?

marijnh:
@mightyguava We could set up the vim mode to force the width of the cursor(s) to be the actual width of the character they are on top of after every selection update. It's more expensive, but should give better results.

DelightRun:
Thank you for your solution, it works! Thank you very much..

And, I’m agree with @Marijn Haverbeke: we can set `font-family` of `.CodeMirror` in `codemirror.css` to a specific monospace font like `Source Pro` or `Courier`.

Should I submit a `Pull Request` for that?

marijnh:
> And, I’m agree with @Marijn Haverbeke: we can set `font-family` of `.CodeMirror` in `codemirror.css` to a specific monospace font like `Source Pro` or `Courier`.

[...]

PR #6765, opened 2021-08-14
[vim] fix fat cursor for non-monospace fonts

KryDos:
This pull request should fix https://github.com/codemirror/CodeMirror/issues/3256

(non monospace font and header size is different from other portion of a text)
image
image

I don't think it's good idea to check if vim mode is enabled directly in `cursorCoords` but I don't really see the better way. 

Could somebody help me with better solution? 

Thanks

### DIFF
--- a/src/display/selection.js
+++ b/src/display/selection.js
@@ -37,6 +37,11 @@ export function drawSelectionCursor(cm, head, output) {
   cursor.style.top = pos.top + "px"
   cursor.style.height = Math.max(0, pos.bottom - pos.top) * cm.options.cursorHeight + "px"
 
+  if (/cm-fat-cursor/.test(cm.getWrapperElement().className)) {
+    let charPos = charCoords(cm, head, "div", null, null);
+    cursor.style.width = Math.max(0, charPos.right - charPos.left) + "px";
+  }
+
   if (pos.other) {
     // Secondary cursor, shown when on a 'jump' in bi-directional text
     let otherCursor = output.appendChild(elt("div", "\u00a0", "CodeMirror-cursor CodeMirror-secondarycursor"))


marijnh:
Making a core method work differently when the vim mode is enabled in order to be able to use it in a way it isn't intended to is way too dodgy. Maybe see if you can use `charCoords` instead.

KryDos:
@marijnh, do you mean I should update `drawSelectionCursor` to call `charCoords` somewhere next to `cursorCoords`? If so, that makes sense and feels better. I'll try.

KryDos:
Ok, new change feels much better. We don't touch measurement functions at all now. `drawSelectionCursor` function is the only affected by the change and the change itself is related to how the cursor is drawn. 

@marijnh thanks for the advice

marijnh:
The core library shouldn't know anything about the vim mode, so testing for it like this doesn't seem like a good idea.

There seems to already be logic in keymap/vim.js that sets a bookmark with a `cm-animate-fat-cursor` class. But somehow it is only firing in visual mode. I don't know the code well enough to figure out why that is, but it seems that an approach like that (having the vim mode use `markText` to display its visual cursor) would be the way to do this.

KryDos:
Yeah, but that part in visual mode doesn't update the cursor width. It just draws one more "fake cursor" and hides the actual cursor. 
The actual cursor isn't well-shaped even in visual mode.

I agree that check for vim mode in code library doesn't look well. 

Vim mode assigns `cm-fat-cursor` class to the wrapper element so we can check that I think.

marijnh:
Hm, that's still a bit of a hack, but I guess it is preferable to the alternatives. Let's try it this way.

PR #6767, opened 2021-08-17
Allow merge editor buttons to be used with keyboard

SachinGupta21:
When the merge editor is used with VoiceOver utility, it is not possible to use "Revert chuck" and "Toggle locked scrolling" buttons with keyboard. With this change, it will be possible.


marijnh:
Thank you!

PR #6768, opened 2021-08-19
Fix emacs scroll-up / scroll-down bindings

a-stewart:
Commit https://github.com/codemirror/CodeMirror/commit/97ee57515599e95972567c41ed9176ddfb1f4fb9 changed the baviour of Ctrl-V and Alt-V and I believe this was unintentional.

Previously, it was set to
```
    "Alt-V": move(byPage, -1), "Ctrl-V": move(byPage, 1),
    "PageUp": move(byPage, -1), "PageDown": move(byPage, 1),
```

So Alt-V and PageUp were the same command and Ctrl-V and PageDown likewise.

I think this is the correct behaviour and is as is described here: https://www.gnu.org/software/emacs/manual/html_node/emacs/Scrolling.html

This PR updates it accordingly.
I have set scrollUpCommand to move forward by 1 page, inverting the functionality of that command as that is close to how the command is described in the emacs docs. (And likewise with scrollDownCommand.)

### DIFF
--- a/keymap/emacs.js
+++ b/keymap/emacs.js
@@ -330,9 +330,9 @@
 
   cmds.previousLine = move(byLine, -1);
 
-  cmds.scrollDownCommand = move(byPage, 1);
+  cmds.scrollDownCommand = move(byPage, -1);
 
-  cmds.scrollUpCommand = move(byPage, -1);
+  cmds.scrollUpCommand = move(byPage, 1);
 
   cmds.backwardParagraph = move(byParagraph, -1);
 
@@ -480,8 +480,8 @@
     "Home": "goLineStart",
     "Alt-V": "scrollDownCommand",
     "Ctrl-V": "scrollUpCommand",
-    "PageUp": "scrollUpCommand",
-    "PageDown": "scrollDownCommand",
+    "PageUp": "scrollDownCommand",
+    "PageDown": "scrollUpCommand",
     "Ctrl-Up": "backwardParagraph",
     "Ctrl-Down": "forwardParagraph",
     "Alt-{": "backwardParagraph",


marijnh:
Thanks for spotting that.

However, this patch looks like you're fixing the problem two times, thus not actually changing the behavior of the bindings (first by flipping the sign in the definition of the commands, and then by also binding the keys to different commands). It looks like only the key bindings should be updated, and the commands were correct before.

a-stewart:
PageUp / PageDown _should_ be unchanged as they worked correctly before - PageDown now calls `scrollUpCommand` which does:
> Scroll forward by nearly a full window

This keeps `scrollUpCommand` consistent with `scroll-up-command` used by emacs.

The change here, is that Ctrl-V still points at `scrollUpCommand`, which has now had it's behaviour flipped.

marijnh:
Are you sure? This is the scroll down command after your patch:

```javascript
cmds.scrollDownCommand = move(byPage, -1);
```

Moving by a negative amount, i.e. towards the start of the document, would be moving up, no?

a-stewart:
It seems backwards but that is what is described by https://www.gnu.org/software/emacs/manual/html_node/emacs/Scrolling.html

> `C-v`
> `PageDown`
> `next`
> Scroll forward by nearly a full window (`scroll-up-command`).
>
> `M-v`
> `PageUp`
> `prior`
> Scroll backward (`scroll-down-command`).

So Ctrl-V and PageDown should both point at the `scroll-up-command`, and the scroll up command, should move the document forward by 1 page.

I think the names used by emacs, refer more to the natural scrolling idea, where scrolling "up" means the content is moving up, but the viewport is moving down the document.

marijnh:
> It seems backwards

It really does, but indeed it checks out—emacs being emacs I guess. Thanks for the explanation.

PR #6781, opened 2021-09-16
Remove text-shadow from solarized theme.

johnspiegel:
The text drop shadow unnecessarily makes text slightly blurry.


marijnh:
The shadow doesn't seem to be in the original theme, so I agree. Merged.

Issue #6788, opened 2021-09-30
Scrolling by wheel move unusably slow on edge with  Chromium 94


laggage:
This problem is probably related to the edge `Smooth Scrolling`  feature. If I disabled this feature via `edge://flags/` page, the editor just works fine

![image](https://user-images.githubusercontent.com/38829279/135413173-6f63a80d-5862-41a4-b416-f45d81e8000f.png)

problem demo in gif
![codemirror滚动问题](https://user-images.githubusercontent.com/38829279/135413432-c31299fb-6ffa-4b96-95bc-6b04b895b813.gif)

I have already have a try to fix the problem: 

```js
// Sync scrolling between fake scrollbars and real scrollable
  // area, ensure viewport is updated when scrolling.
  on(d.scroller, "scroll", (e) => {
    if (d.scroller.clientHeight) {
      setTimeout(() => {
        updateScrollTop(cm, d.scroller.scrollTop);
        setScrollLeft(cm, d.scroller.scrollLeft, true);
        signal(cm, "scroll", cm);
      }, 120);
    }
  });
```

marijnh:
Does the problem persist with CodeMirror 5.63.1?

laggage:
> Does the problem persist with CodeMirror 5.63.1?

yes it does

sfc-gh-acourtney:
@marijnh I also encountered this, and upgraded to 5.63.1 because the description indicates it fixes the issue.
But it didn't fix it, so I dove in a bit with the debugger, and was surprised to find the old event handler still around. 

It looks to me like 5.63.1 was inadvertently tagged at the wrong point, missing the actual crucial diff.
If you look on the [release page for 5.63.1](https://github.com/codemirror/CodeMirror/releases/tag/5.63.1), it has a link to ccdcd15d203a931bbbeb4ca723e379bd5cf2331b as the diff for the release. Looking at that diff, it includes all of the release notes and version upgrades, but no code diff.
 The release page for 5.63.1 also has a link to [1 commit](https://github.com/codemirror/CodeMirror/compare/5.63.1...master) to master _since_ this release (emphasis mine), and that commit seems to have the crucial code diff / bug fix...

Thank you to the maintainers for swinging into action and making a fix to this issue available so quickly. Would be extremely grateful if you would cut a 5.63.2 off of master that includes the critical fix! 🙏

marijnh:
> Looking at that diff, it includes all of the release notes and version upgrades, but no code diff.

[...]

PR #6789, opened 2021-09-30
fix: fix scrolling by wheel move unusably slow on edge with Chromium 94

laggage:
address #6788


laggage:
have been solved in chromium 94.0.992.37

PR #6812, opened 2021-11-01
Add esm as rollup output

luucvanderzee:
I was trying to use `CodeMirror` in a [SvelteKit](https://kit.svelte.dev/) project. Because `SvelteKit` currently only supports es-modules, and `CodeMirror` didn't have that yet, I added it. How does the configuration look? Anything I should change?

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,11 @@
   "version": "5.63.3",
   "main": "lib/codemirror.js",
   "style": "lib/codemirror.css",
+  "type": "module",
+  "exports": {
+    "import": "./lib/codemirror.esm.js",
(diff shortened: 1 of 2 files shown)


marijnh:
This won't really work, because all the modes and addons will still be in UMD format and unable to interface with the ESM module. You can try to integrate [CodeMirror 6](https://codemirror.net/6/) instead, which is more ESM-friendly, or, if you really want to use version 5.x, set up scripts convert all of the files in the project to ESM, and publish/maintain it as a separate package. (This repository isn't going to make major additions like that anymore.)

PR #6829, opened 2021-11-22
fix handling of escaped symbols in Scheme mode #6822

### LINKED ISSUE #6822: Scheme mode issue with escaped symbols
I found an issue in Scheme mode. GitHub has more issues than CodeMirror that only don't handle escaped symbols in the form: `|foo bar|`.

Here is my nasty example:

```
(let ((x #;(foo
             bar ")") 10)
      (|foo )))) bar| 20))
   (print #|
            this is comment ))))
          |# x)
   (print (+ x |foo )))) bar|)))
```
Here `|foo )))) bar|` is a valid symbol but it's not processed as a single token.

I can try to fix the issue and create a PR since I use a modified version that I also need to fix.

Here is a demo with matching parentheses that use my modified version of the mode (I have extensions on top of Scheme that I want to support) https://codepen.io/jcubic/pen/VwKeZob?editors=1010

jcubic:
This PR fixes #6822 and handling of escaped symbols, I've reused handling of strings so escape inside symbols works the same.

Here is an example of how this works, the code for the mode is the same (it includes my modifications, but the added code is exactly the same)

https://codepen.io/jcubic/pen/qBXwGaV

### DIFF
--- a/mode/scheme/scheme.js
+++ b/mode/scheme/scheme.js
@@ -3,7 +3,7 @@
 
 /**
  * Author: Koh Zi Han, based on implementation by Koh Zi Chun
- * Imroved by: Jakub T. Jankiewicz
+ * Improved by: Jakub T. Jankiewicz
  */
 
 (function(mod) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks! Could you correct the typo ('sequance' should be 'sequence')?

jcubic:
done, I've also squashed the commits.

marijnh:
Wonderful, thank you

PR #6833, opened 2021-11-26
Update meta.js: add "dataviewjs" as alias for JS

jnsgruk:
Hoping this doesn't count as contributing a new mode, but feel free to close this if I've misread the guidance! 

I'm a user of the [Obsidian.md](https://obsidian.md) note taking app. There is a popular plugin called [Dataview](https://github.com/blacksmithgu/obsidian-dataview/), which allows you to write queries in Javascript to compose/present notes. I've also been using a [syntax highlighting](https://github.com/deathau/cm-editor-syntax-highlight-obsidian) plugin.

Javascript blocks to be used by dataview are annotated with the type `dataviewjs`. This PR just adds an alias to the Javascript mode such that `dataviewjs` blocks are highlighted the same as Javascript.

### DIFF
--- a/mode/meta.js
+++ b/mode/meta.js
@@ -71,7 +71,7 @@
     {name: "Java", mime: "text/x-java", mode: "clike", ext: ["java"]},
     {name: "Java Server Pages", mime: "application/x-jsp", mode: "htmlembedded", ext: ["jsp"], alias: ["jsp"]},
     {name: "JavaScript", mimes: ["text/javascript", "text/ecmascript", "application/javascript", "application/x-javascript", "application/ecmascript"],
-     mode: "javascript", ext: ["js"], alias: ["ecmascript", "js", "node"]},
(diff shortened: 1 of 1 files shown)


marijnh:
This seems like a *very* tool-specific tweak. Have you tried asking the Obsidian maintainers to inject this on their side? That would probably be a more appropriate place.

jnsgruk:
I can certainly ask - thanks for the quick reply. I agree this seems specific, I went around a couple of ways of achieving this and adding an alias seemed like the most appropriate means. I'll come back here in a while, but obviously feel free to close this if it feels too out of place :)

jnsgruk:
Well, this is embarassing. I was looking at the latest release, not the main branch of the other tool, where someone [has already added this](https://github.com/jnsgruk/cm-editor-syntax-highlight-obsidian/blob/fb9a1a08a19491ceada0ebabad2ed5087885cb3b/mode/meta.js#L62). Thanks again for the quick reply.

Issue #6844, opened 2021-12-09
toggleComment extension is broken for Factor, due to incorrect lineComment


Kacarott:
In Factor, line comments are marked with the symbol `!`. When using the toggleComment extension to comment a line, the following *wrong* symbol is placed instead: `!,#!`. This is because, currently the Factor language mode has its lineComment value set to a list of two symbols `lineComment: [ "!", "#!" ]` instead of the standard single symbol. This results in the string used for toggleComment being the entire array, stringified.

As far as I am aware, Factor uses exclusively `!` as its line comments, and the Factor language mode should be fixed accordingly, to make the toggleComment extension useable.

As a side note, if multiple line comments options are indeed supported by CodeMirror, then the toggleComment extension should additionally be fixed to choose only one of the available options (in cases where there is more than one). However that is not the purpose of this issue.

PR #6845, opened 2021-12-09
Fixed incorrect lineComment (#6844)

Kacarott:
Removed redundant line comment symbol from Factor language mode, fixing #6844 .


marijnh:
Thanks!

PR #6848, opened 2021-12-17
Add BQN mode

dunyakirkali:


marijnh:
Hi. We're no longer adding new modes to the main CodeMirror distribution. The recommended way to distribute additional modes is to publish them as separate npm packages—see for example [the Elixir mode](https://www.npmjs.com/package/codemirror-mode-elixir).

PR #6855, opened 2021-12-23
[css mode] add caseInsensitive hint

Realetive:
Current PR allow to invoke CSS hint for property or value in any case:

Снимок экрана 2021-12-24 в 00 07 14

Снимок экрана 2021-12-24 в 00 07 35

### DIFF
--- a/addon/hint/css-hint.js
+++ b/addon/hint/css-hint.js
@@ -30,7 +30,7 @@
       return {list: ["!important"], from: CodeMirror.Pos(cur.line, token.start),
               to: CodeMirror.Pos(cur.line, token.end)};
 
-    var start = token.start, end = cur.ch, word = token.string.slice(0, end - start);
+    var start = token.start, end = cur.ch, word = token.string.slice(0, end - start).toLowerCase();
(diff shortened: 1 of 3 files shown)


addon/hint/css-hint.js
```diff
@@ -41,7 +41,7 @@
     function add(keywords) {
       for (var name in keywords)
         if (!word || name.lastIndexOf(word, 0) == 0)
-          result.push(name);
```
marijnh, line 44: This definitely won't work for `pseudoClasses` defined above. What is it intended to do?

Realetive: CSS is not case-sensitive. The current solution doesn't work with properties and values that the user will camelCase (ex. [`optimizeSpeed`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-rendering#values) instead of `optimizespeed`). I want to fix this. Do you have any idea how to solve the problem with `pseudoClasses`?

marijnh: No I mean, given that `keySet` is only called with lower-case strings, what is the purpose of making the object values hold another copy of the string?

PR #6865, opened 2022-01-10
[fortran mode] Generate regular expression using wordRegexp

dcshzj:
The regular expression used to highlight operators has issues with certain functions such as "len" and "new_line", especially if there is an opening bracket right before these functions.

This issue was originally reported in https://github.com/carbon-app/carbon/issues/1308 and https://github.com/highlightjs/highlight.js/issues/3450. The original issue is at https://github.com/Beliavsky/FortranTip/issues/24.

### DIFF
--- a/mode/fortran/fortran.js
+++ b/mode/fortran/fortran.js
@@ -20,6 +20,10 @@ CodeMirror.defineMode("fortran", function() {
     return keys;
   }
 
+  function wordRegexp(words) {
+    return new RegExp("^((" + words.join(")|(") + "))\\b");
+  }
+
   var keywords = words([
                   "abstract", "accept", "allocatable", "allocate",
                   "array", "assign", "asynchronous", "backspace",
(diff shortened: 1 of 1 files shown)


mode/fortran/fortran.js
```diff
@@ -20,6 +20,10 @@ CodeMirror.defineMode("fortran", function() {
     return keys;
   }
 
+  function wordRegexp(words) {
+    return new RegExp("^((" + words.join(")|(") + "))\\b");
```
marijnh, line 24: Would this not work without the extra parentheses around every word? I didn't benchmark, but it seems that, when matching such a regexp, the regexp engine would be forced to allocate a bunch of empty match group slots in the result array, which seems wasteful.

mode/fortran/fortran.js
```diff
@@ -112,7 +116,7 @@ CodeMirror.defineMode("fortran", function() {
                      "c_short", "c_signed_char", "c_size_t", "character",
                      "complex", "double", "integer", "logical", "real"]);
   var isOperatorChar = /[+\-*&=<>\/\:]/;
-  var litOperator = new RegExp("(\.and\.|\.or\.|\.eq\.|\.lt\.|\.le\.|\.gt\.|\.ge\.|\.ne\.|\.not\.|\.eqv\.|\.neqv\.)", "i");
```
marijnh, line 115: Any idea what these (uselessly) escaped period characters might have been for? I don't know Fortran, but if I google them I see these operators surrounded by actual periods (which should have been `\\.` in this string).

Also the new code drops the case-insensitive flag. Is that intentional?

dcshzj:
Hi @marijnh, thanks for the review! Unfortunately it seems like I went in the wrong direction when trying to come up with a fix for the issue mentioned. I'll see if I can come up with a proper fix for this.

dcshzj:
Unfortunately I am not able to come up with a proper fix. I have filed #6869 for someone else to work on it.

PR #6866, opened 2022-01-11
Addition to regex to matches numbers with underscores in pearl

WillHernandez:
#6683 
Made an addition to the current regex pearl.js uses to match a 'number'. Numbers that contain underscores as separators are now matched.

### DIFF
--- a/mode/perl/perl.js
+++ b/mode/perl/perl.js
@@ -514,7 +514,7 @@ CodeMirror.defineMode("perl",function(){
                 if(state.chain)
                         return tokenChain(stream,state,state.chain,state.style,state.tail);
                 if(stream.match(/^\-?[\d\.]/,false))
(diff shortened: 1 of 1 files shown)


marijnh:
Making the same input (in this case `\d+`) match multiple arms of a choice operator in a regexp is generally a bad idea. Any reason you didn't integrate this with the `\d+(e[+-]?\d+)?)` case at the end of the regexp?

WillHernandez:
> Making the same input (in this case `\d+`) match multiple arms of a choice operator in a regexp is generally a bad idea. Any reason you didn't integrate this with the `\d+(e[+-]?\d+)?)` case at the end of the regexp?

I made the mistake of assuming adding an or operator with my own small regex would make less of an impact than altering one that already existed and possibly ruining another match.
`|[\d_]+(e[+-]?\d+)?)` Seems to fix the issue unless I missed something. 
Thanks for your help

marijnh:
Attached patch adds this (also to the hex and binary parts of the expression).

PR #6874, opened 2022-02-01
[htmlmixed mode] Fixed escaping in double-quote regex string

jbaicoianu:
The htmlmixed mode's `getTagRegexp()` function composes a regex out of double-quoted strings, which contain the `\s` regex character specifier. However, this is first being interpreted as a string escape code, which ends up being equivalent to `/<\/s*/` (zero or more instances of the 's' character) instead of the intended `/<\/\s*/` (zero or more instances of any whitespace character).

This pull request fixes the escaping to be consistent with that used in the `getAttrRegexp()` function a few lines above.

### DIFF
--- a/mode/htmlmixed/htmlmixed.js
+++ b/mode/htmlmixed/htmlmixed.js
@@ -50,7 +50,7 @@
   }
 
   function getTagRegexp(tagName, anchored) {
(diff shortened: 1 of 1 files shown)


marijnh:
Thanks!

PR #6876, opened 2022-02-07
Allow mapping <Esc> in vim normal mode

### LINKED ISSUE #6234: Allow mapping Escape (<Esc>) key in vim normal mode
Trying to map `<Esc>` key with `map()` or `mapCommand()` doesn't do anything. Mapping it in insert or visual mode doesn't really make much sense but it should be possible to map it in normal mode.

Changing [`handleEsc()`](https://github.com/codemirror/CodeMirror/blob/87aaaed6649bad36218faded697dd69ee18c4850/keymap/vim.js#L881-L892) to below code might be enough to fix this.
```js
function handleEsc() {
  if (key == '<Esc>') {
    // Clear input state and get back to normal mode.
    if (vim.visualMode) {
      clearInputState(cm);
      exitVisualMode(cm);
      return true;
    } else if (vim.insertMode) {
      clearInputState(cm);
      exitInsertMode(cm);
      return true;
    }
  }
}
```

MrQubo:
Resolves  #6234 

This change only affects normal mode. It's still possible to map <Esc> in insert and visual mode, and it will be ignored in favour of the default action of exiting to normal mode.

### DIFF
--- a/keymap/vim.js
+++ b/keymap/vim.js
@@ -865,13 +865,17 @@
         }
         function handleEsc() {
           if (key == '<Esc>') {
-            // Clear input state and get back to normal mode.
-            clearInputState(cm);
             if (vim.visualMode) {
+              // Get back to normal mode.
               exitVisualMode(cm);
             } else if (vim.insertMode) {
+              // Get back to normal mode.
               exitInsertMode(cm);
+            } else {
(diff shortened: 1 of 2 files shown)


marijnh:
Thanks!

PR #6881, opened 2022-02-12
Solarized dark theme

creemy0cheesekake:
Added the solarized dark theme created by FarhadG


marijnh:
Sorry, we're not adding new themes to this distribution anymore. If you want to publish the theme, you could do so as a separate package.

PR #6882, opened 2022-02-18
Fix white CodeMirror-scrollbar-filler on dark themes

erosman:
Previously #6485

### DIFF
--- a/lib/codemirror.css
+++ b/lib/codemirror.css
@@ -19,7 +19,7 @@
 }
 
 .CodeMirror-scrollbar-filler, .CodeMirror-gutter-filler {
-  background-color: white; /* The little square between H and V scrollbars */
+  background-color: inherit; /* The little square between H and V scrollbars */
 }
 
 /* GUTTER */


marijnh:
Won't that still make the element transparent whenever the editor itself is transparent?

erosman:
Any other ideas?

The white square doesn't look good on dark theme, and even if transparent when editor is transparent, it would still look better than the while square. ;)

What about getting its `background-color` from the scrollbar `background-color`?!

marijnh:
I think adding a rule to color this class in the themes would be the best approach. A bit burdensome, but without using CSS variables (which some of our supported browsers don't have), I don't see a more convenient solution.

erosman:
That would require the use of `var()` and `@media screen and (prefers-color-scheme: light/dark)` which is not supported by older browsers.

Even then, the color will not be a perfect match for all dark themes.

It doesn't bother me as I had already added `background-color: inherit;` to my own CSS and it works fine for my browser extension use.

I thought the less than ideal `inherit` might be more adaptable than the status que.

Scutua (approved):

PR #6894, opened 2022-03-10
fixed option(keyMap) to check for undefined keymap

timmaffett:
This is a rather minor fix but it prevents setting unknown keyMap options from crashing editor


marijnh:
This won't prevent `lookupKey` from still crashing later. And more generally, the keymap option is specified to take an object, so if something is passing undefined, the error is on the side of the code doing that, and should be fixed there.

PR #6903, opened 2022-04-04
Don't use italics for styling code boxes (in Joplin)

JOJ0:
I'm coming here from the Joplin forums, this theme is used as the "Solarized Light/Dark" themes over there. Some other users and me agreed on that having italic styled text is not a good choice for monospaced text, it is hard to read. In Joplin this style is used when "editing codeboxes (inside tripple backticks) in markdown)

Have a look at the discussion here and tell me what you think: https://discourse.joplinapp.org/t/why-is-the-markdown-editor-monospace-font-italics/24801/6

Thanks for a great theme and a great editor though!

### DIFF
--- a/theme/solarized.css
+++ b/theme/solarized.css
@@ -60,7 +60,7 @@ http://ethanschoonover.com/solarized/img/solarized-palette.png
 .cm-s-solarized .cm-property { color: #2aa198; }
 .cm-s-solarized .cm-operator { color: #6c71c4; }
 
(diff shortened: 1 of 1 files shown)


marijnh:
How that looks will depend on the font you're using. But if the author of the theme made that italic, I'd say that's how the theme intends it, and that's how it should stay. This is easy enough to override in local CSS if you want to use the Solarized theme but tweak the comment style.

JOJ0:
IMHO monospace font and italic is never a good choice.

Who is the theme author?

Thanks in advance!

PR #6916, opened 2022-05-13
fix the paste operation when the cell is unfocused

richard-fung-db:
I noticed that when the border of the CodeMirror component is clicked but the editor has not received the blinking cursor, that the paste operation can still happen.

Reproduce in master:
This can be seen by copying some text, click the border of the CodeMirror editor exactly (e.g., at https://codemirror.net/demo/html5complete.html), and paste using the system keyboard shortcut. The pasted text appears in the editor without a blinking cursor.


marijnh:
Seems to be a Chrome issue (doesn't happen in Firefox), but this workaround looks safe. Merged.

PR #6918, opened 2022-05-18
Preserve whitespace when commenting

lukehaas:
Currently, when you comment the first line of an empty function you lose the indent.

165898422-09f03ca5-863a-4ff8-90fa-cf6b02fc6781

This PR has a simple fix that means commenting an indented empty line will keep the indent.

### DIFF
--- a/addon/comment/comment.js
+++ b/addon/comment/comment.js
@@ -78,7 +78,8 @@
         var baseString = null;
         for (var i = from.line; i < end; ++i) {
           var line = self.getLine(i);
-          var whitespace = line.trim().length === 0 ? line : line.slice(0, firstNonWS(line));
+          var nonWS = firstNonWS(line);
+          var whitespace = nonWS === 0 ? line : line.slice(0, nonWS);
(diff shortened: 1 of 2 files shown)


marijnh:
I can't reproduce this -- when I input a function like that, put the cursor on the (indented) line inside the braces, and press ctrl-/, the comment token is indented by two spaces.

lukehaas:
@marijnh is that shortcut definitely triggering the `cm.toggleComment({ indent: true });` function?

marijnh:
Oh, I wasn't paying attention and tested with CodeMirror 6 (good news: the problem doesn't exist there). Could you rewrite this to not use the `trim` method, which doesn't exist on some of the browsers this project supports? Maybe just put the result of calling `firstNonWS` in a variable and compare that to the string length.

lukehaas:
@marijnh, thanks, that's updated now.

lukehaas:
@marijnh, thanks, that's updated now.

PR #6920, opened 2022-05-24
Minor changes to make CodeMirror accessible (label and color-contrast)

kudelek:
Minor change made for the project to be a bit more accessible ('label' violation)

### DIFF
--- a/addon/search/search.js
+++ b/addon/search/search.js
@@ -202,11 +202,12 @@
   }
 
   function getQueryDialog(cm)  {
-    const label = el("label", {className: "CodeMirror-search-label"}, cm.phrase("Search:"));
+    const label = el("label", {className: "CodeMirror-search-label"}, 
+                     cm.phrase("Search:"),
+                     el("input", {type: "text", "style": "width: 10em", className: "CodeMirror-search-field", id: "CodeMirror-search-field"}));
     label.setAttribute("for","CodeMirror-search-field");
     return el("", null,
-              label, " ",
-              el("input", {type: "text", "style": "width: 10em", className: "CodeMirror-search-field", id: "CodeMirror-search-field"}), " ",
+              label, " ", " ",
               el("span", {style: "color: #888", className: "CodeMirror-search-hint"},
                  cm.phrase("(Use /re/ syntax for regexp search)")));
   }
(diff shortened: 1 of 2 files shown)


addon/search/search.js
```diff
@@ -202,9 +202,11 @@
   }
 
   function getQueryDialog(cm)  {
+    const label = el("label", {className: "CodeMirror-search-label"}, cm.phrase("Search:"));
+    label.setAttribute("for","CodeMirror-search-field");
     return el("", null,
-              el("span", {className: "CodeMirror-search-label"}, cm.phrase("Search:")), " ",
-              el("input", {type: "text", "style": "width: 10em", className: "CodeMirror-search-field"}), " ",
+              label, " ",
+              el("input", {type: "text", "style": "width: 10em", className: "CodeMirror-search-field", id: "CodeMirror-search-field"}), " ",
```
marijnh, line 209: Using a fixed ID here will break if there are multiple editors in the page. Maybe just wrap the input in the label instead?

src/input/input.js
```diff
@@ -131,5 +131,6 @@ export function hiddenTextarea() {
   // If border: 0; -- iOS fails to open keyboard (issue #1287)
   if (ios) te.style.border = "1px solid black"
   disableBrowserMagic(te)
+  te.setAttribute("aria-hidden","true")
```
marijnh, line 134: Hiding the input element from the screen reader seems a _very_ bad idea. What is the purpose of this line?

addon/search/search.js
kudelek: Done

src/input/input.js
kudelek: Sorry, I probably misunderstood what is this textarea for. Maybe it would be good to just add aria-label for this textarea?
If so, what should be in the aria-label?

marijnh: This creates the main focused element when `inputMode` is `"textarea"`. Its label can be set with [this option](https://codemirror.net/doc/manual.html#option_screenReaderLabel).

kudelek: Ok, I reverted this change and set the label locally. That's all for this pull request

Issue #6935, opened 2022-06-10
Glitches when adding newlines


jcubic:
I'm not sure if this is because of bugs in PEG.js mode but I have weird results when I'm adding newlines to the editor. Somethings all lines are broken sometimes they are fixed.

![PEGJS_mode](https://user-images.githubusercontent.com/280241/173076875-7d29cdf3-9561-4a70-8dcf-669162d95ebe.gif)

This may be because incremental parser and PEG.js have inner JS mode.

Do you think that the problem is in PEG.js mode or in CodeMirror?

jcubic:
I'm not sure but I don't see only

```javascript
state.localState = CodeMirror.startState(jsMode);
```

I don't see anywhere ending the mode in PEG.js (no "}" check anywhere). Maybe that is the problem. But I'm not sure how to turn off the inner mode, there is no `endState` in CodeMirror.

marijnh:
There is no file named PEG.js in the CodeMirror distribution. Is this about pegjs.js, or some 3rd party mode?

jcubic:
About default pegjs.js mode that is included in the library. The name for the parser generator is PEG.js so I used the name of the project not the filename.

marijnh:
Your PR looks rather unfinished and confused, and this issue doesn't provide a bug description that I can reproduce, so at this point I can't really do much. Also, that mode was contributed and I'm not willing to invest a lot of time into it.

jcubic:
Yes, the PR is broken, I thought that I've fixed the pegjs mode but I'm not able to make it work the same as htmlmixed. I'm getting stack overflow when trying to get pegjs mode inside pegjs mode. It's really hard because there is no detailed documentation how to write modes. At least I didn't find any.

PR #6936, opened 2022-06-10
fix PegJS mode

jcubic:
The fix is partial for #6935 it works better but there are still issues when adding a new line.

The code for the token in the state is taken from the htmlmixed mode.

I've added `'\x5d'` instead of `']'` because it was breaking JS mode.

I can try to fix the other issue (when editing code in Rule) but I will need some help. I have no clue what is happening and why the code is changing the styling (similar to the linked issue but later in the code).


jcubic:
I'm not sure why but my code doesn't exist JS mode.

jcubic:
Sorry, I have no idea how to create a nested mode. Every example uses some other mode. Based on htmlmixed mode I'm not able to reference pegjs mode from inside pegjs to make it work like htmlmixed.

PR #6945, opened 2022-06-23
Make sure v5 patches don't override latest on npm

masad-frost:
See https://github.com/codemirror/dev/issues/880

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,5 +43,8 @@
     "dependencies": {},
     "devDependencies": {}
   },
-  "dependencies": {}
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -43,5 +43,8 @@
     "dependencies": {},
     "devDependencies": {}
   },
-  "dependencies": {}
+  "dependencies": {},
+  "publishConfig": {
+    "tag": "legacy"
```
masad-frost, line 48:
```suggestion
    "tag": "version5"
```

PR #6964, opened 2022-07-19
Focus the window in which CodeMirror is rendered.

Me1000:
We ran into some issues where under certain conditions Firefox would lose focus of the editor. In particular if CodeMirror is being rendered in a frame different than the frame in which the code itself is running. This PR ensures that the call to `focus()` on mouse clicks happens on the window in which CodeMirror is being rendered, and not the window running the code.

This PR is similar to [#5309](https://github.com/codemirror/codemirror5/pull/5309).


marijnh:
Seems a bunch of global `window`/`document` uses have ended up in the codebase again. Attached patch fixes them.

PR #6965, opened 2022-07-19
Fix for new function name clashing with local variable.

Me1000:
Text selection was broken in the latest commit because the `extend` function is defined within the `leftMouseSelect` function and that caused `doc()` the function to be shadowned by `doc = cm.doc` the local variable.


marijnh:
Argh. Thanks for spotting that.

PR #6974, opened 2022-08-08
fix: diff merge support \r line separator

liangshuai:
image

image


marijnh:
This doesn't look right — it'll only look for `\r` when there are no `\n` characters in the string, so in `"foo\rbar\n"` it'd ignore the `\r`.

Also, won't CodeMirror have already normalized these to `\n` anyway, since `getDiff` is called on the editor content?

PR #6975, opened 2022-08-09
Jinja2 mode Line statements support

vled12:
According to Jinja2 [template designer manual](https://jinja.palletsprojects.com/en/3.1.x/templates/#line-statements), the line statements are now supported. Test example is included.
Beside that, keyword list is slightly updated.

### DIFF
--- a/mode/jinja2/jinja2.js
+++ b/mode/jinja2/jinja2.js
@@ -79,19 +79,19 @@
           stream.next();
           return "string";
         }
-        else if (state.inbraces != null && state.inbraces > 0 && ch ==")") {
+        else if (state.inbraces > 0 && ch ==")") {
           stream.next()
           state.inbraces--;
         }
-        else if (state.inbraces != null && ch == "(") {
+        else if (ch == "(") {
           stream.next()
           state.inbraces++;
         }
-        else if (state.inbrackets != null && state.inbrackets > 0 && ch =="]") {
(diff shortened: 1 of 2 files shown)


Jay4577:
Pullrequst

Jay4577 (approved):

mode/jinja2/jinja2.js
```diff
@@ -78,7 +78,24 @@
           state.instring = ch;
           stream.next();
           return "string";
-        } else if(stream.match(state.intag + "}") || stream.eat("-") && stream.match(state.intag + "}")) {
+        }
+        else if (state.inbraces > 0 && ch ==")") {
```
marijnh, line 82: Can you make sure `inbraces`/`inbrackets` are initialized in the start state?

vled12: `inbraces / inbrackets` are accessed only in `intag` state.
Currently they are initialized right after the turning `intag` state on. I assume, this should be sufficient.

marijnh: It seems dodgy to compare `undefined > 0` here (though it will in fact produce false)

marijnh:
Sure that works, but why not just initialize them to 0 in `startState` and avoid the question of whether they are defined entirely?

marijnh:
Thanks. Merged.

PR #6978, opened 2022-08-25
improve(highlight): long text skip stream.eol()

xuchangjunjx:
when text is very long, function `runMode`  code snippet `while (!stream.eol()) ` will loop long time.


marijnh:
Doesn't the `if` directly inside the `while` loop take care of precisely this?

xuchangjunjx:
你好，如果你看到我这封邮件，说明我已收到你的信息。若有问题我会抽空回复你。

marijnh:
Yes, if tokens are all one character long it will. This is intentional — it highlights up to `maxHighlightLength`, and skips the rest of the line.

PR #6979, opened 2022-08-26
[Show-Hints] Hint popup exceeds the viewport's height

benjaminr-ps:
The current implementation does not taking care of the top-position of the popup window while calculating the final height of it.

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -291,7 +291,7 @@
       if (curTop - height > 0) { // Fits above cursor
         hints.style.top = (top = pos.top - height - offsetTop) + "px";
         below = false;
-      } else if (box.top + height > winH) {
+      } else if (height > winH) {
         hints.style.height = (winH - 5) + "px";
         hints.style.top = (top = pos.bottom - box.top - offsetTop) + "px";
         var cursor = cm.getCursor();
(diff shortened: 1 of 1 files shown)


marijnh:
That `if` seems to address precisely that — if the popup is higher then the window, it shrinks it. Adding `box.top` doesn't seem reasonable here.

benjaminr-ps:
No, it does not shrink it then, instead it will set the `top` position more to the top.

But therefore, the actual `top` position must be taken into account, here.

benjaminr-ps:
image

marijnh:
Right. But the branch you're modifying here is specifically for the case where the pop-up is itself higher than the screen, so just adding this without adjusting the rest of the logic is not a good idea.

But also, in your screenshots, the current behavior actually seems preferable — with the change, you can no longer see what you are typing because the pop-up covers it.

benjaminr-ps:
Ah, sorry, of course it looks proper to you.
Here is how it is actually rendered, exceeding the viewport:
grafik

benjaminr-ps:
Adapted your suggestion, and only resized the hint's popup window.

You are right, it looks much better, and keeps user-experience high ;)

marijnh:
That only covers the case where the pop-up is below the cursor ... and the existing logic seems seriously dodgy as well. Does attached patch work for you?

benjaminr-ps:
The current version was working fine in case of showing it above.

Nevertheless your implementation works great, and is using the full horizontal space, if needed, as well.

benjaminr-ps:
I will close this pull request, as it is overhauled now.

PR #6985, opened 2022-09-13
Make touch(start/move/end), mousewheel handlers passive

0biWanKenobi:


0biWanKenobi:
nevermind, saw a similar pr that got closed because these events may use `preventDefault`. I went and made the pull request because I saw [this](https://discuss.codemirror.net/t/non-passive-event-listener-warning-in-chrome-console/4752), but oh well..

PR #6991, opened 2022-10-04
[sparql mode] Identify all characters in prefixes

MarkBoyes:
SPARQL specifies prefixes, for example "bio" in the prefix specification below, according to https://www.w3.org/TR/sparql11-query/ Productions for terminals [168]

`PREFIX bio: <http://purl.org/vocab/bio/0.1/>`

sparql.js has a simple last-chance regular expression that is used to match prefixes, operators and keywords. This does not, for example, allow hyphen and period characters in prefixes despite them being permitted (in different places) according to the specification. This has been augmented with the patterns from the specification to identify prefix atoms before the operators and keywords are processed as before.

Note that the mode continues to treat anything after a prefix as being a PN_LOCAL (Productions for terminals [169]). This is correct in the places within a SPARQL statement where a prefix is actually used but a slightly different set of characters is legal at the end of a prefix specification, like the example above.  I have not attempted to address this.

### DIFF
--- a/mode/sparql/sparql.js
+++ b/mode/sparql/sparql.js
@@ -33,6 +33,15 @@ CodeMirror.defineMode("sparql", function(config) {
                              "true", "false", "with",
                              "data", "copy", "to", "move", "add", "create", "drop", "clear", "load", "into"]);
   var operatorChars = /[*+\-<>=&|\^\/!\?]/;
+  var PN_CHARS_BASE =
+  "[A-Z]|[a-z]|[\\u{00C0}-\\u{00D6}]|[\\u{00D8}-\\u{00F6}]|[\\u{00F8}-\\u{02FF}]|" +
+  "[\\u{0370}-\\u{037D}]|[\\u{037F}-\\u{1FFF}]|[\\u{200C}-\\u{200D}]|[\\u{2070}-\\u{218F}]|" +
+  "[\\u{2C00}-\\u{2FEF}]|[\\u{3001}-\\u{D7FF}]|[\\u{F900}-\\u{FDCF}]|[\\u{FDF0}-\\u{FFFD}]|[\\u{10000}-\\u{EFFFF}]";
+  var PN_CHARS_U = PN_CHARS_BASE + "|_";
+  var PN_CHARS = PN_CHARS_U + "|-|[0-9]|\\u{00B7}|[\\u{0300}-\\u{036F}]|[\\u{203F}-\\u{2040}]";
+  var PREFIX_START = new RegExp(PN_CHARS_BASE, "u");
+  var PREFIX_MID = new RegExp(PN_CHARS + "|\\.", "u");
+  var PREFIX_END = new RegExp(PN_CHARS, "u");
 
   function tokenBase(stream, state) {
     var ch = stream.next();
@@ -72,11 +81,19 @@ CodeMirror.defineMode("sparql", function(config) {
       return "meta";
     }
     else {
-      stream.eatWhile(/[_\w\d]/);
-      if (stream.eat(":")) {
-        eatPnLocal(stream);
-        return "atom";
+      if (PREFIX_START.test(ch)) {
+        if (stream.eatWhile(PREFIX_MID)) {
+          stream.backUp(1);
+          if (stream.eat(PREFIX_END) && stream.eat(":")) {
+            eatPnLocal(stream);
+            return "atom";
+          }
+        } else if (stream.eat(":")) {
+          eatPnLocal(stream);
+          return "atom";
+        }
       }
+      stream.eatWhile(/[_\w\d]/);
       var word = stream.current();
       if (ops.test(word))
         return "builtin";
--- a/mode/sparql/sparql.js
+++ b/mode/sparql/sparql.js
@@ -34,14 +34,13 @@ CodeMirror.defineMode("sparql", function(config) {
                              "data", "copy", "to", "move", "add", "create", "drop", "clear", "load", "into"]);
   var operatorChars = /[*+\-<>=&|\^\/!\?]/;
   var PN_CHARS_BASE =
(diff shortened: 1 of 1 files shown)


mode/sparql/sparql.js
```diff
@@ -33,6 +33,15 @@ CodeMirror.defineMode("sparql", function(config) {
                              "true", "false", "with",
                              "data", "copy", "to", "move", "add", "create", "drop", "clear", "load", "into"]);
   var operatorChars = /[*+\-<>=&|\^\/!\?]/;
+  var PN_CHARS_BASE =
+    "[A-Z]|[a-z]|[\\u{00C0}-\\u{00D6}]|[\\u{00D8}-\\u{00F6}]|[\\u{00F8}-\\u{02FF}]|" +
+    "[\\u{0370}-\\u{037D}]|[\\u{037F}-\\u{1FFF}]|[\\u{200C}-\\u{200D}]|[\\u{2070}-\\u{218F}]|" +
+    "[\\u{2C00}-\\u{2FEF}]|[\\u{3001}-\\u{D7FF}]|[\\u{F900}-\\u{FDCF}]|[\\u{FDF0}-\\u{FFFD}]|[\\u{10000}-\\u{EFFFF}]";
```
marijnh, line 39: Could you simplify these a bit by not repeating the `[]` characters? (`[A-Za-Z]` is equivalent to `[A-Z]|[a-z]`).

mode/sparql/sparql.js
```diff
@@ -72,11 +81,19 @@ CodeMirror.defineMode("sparql", function(config) {
...
-      if (stream.eat(":")) {
-        eatPnLocal(stream);
-        return "atom";
+      if (PREFIX_START.test(ch)) {
+        if (stream.eatWhile(PREFIX_MID)) {
+          stream.backUp(1);
+          if (stream.eat(PREFIX_END) && stream.eat(":")) {
+            eatPnLocal(stream);
+            return "atom";
+          }
```
marijnh, line 90: This falls through if the `eat` returns false, despite having consumed characters. What is it testing for, precisely? And does the set of allowed prefix characters really differ from the set of allowed variable characters?

MarkBoyes: It was intentional to eat then fall through, since the remainder of the method used stream.current() and would include those characters: however there was a bug there in that the characters allowed for variables do not include some of those allowed for prefixes so stream.current() will contain characters that should not be styled cm-variable. 

I have rewritten the change to use a more complex regular expression for matching every character after 'ch' (with the existing code for the edge case where there is nothing before the colon left untouched).

mode/sparql/sparql.js
MarkBoyes: I have eliminated many of the square brackets: I have grouped the characters named PN_CHARS_BASE, PN_CHARS_U and PN_CHARS in the specification EBNF together as character classes, so there is still some use of "|".

MarkBoyes:
Thanks for the checks. I have commited fixes for the points raised. Please rereview.

mode/sparql/sparql.js
```diff
@@ -71,20 +79,19 @@ CodeMirror.defineMode("sparql", function(config) {
       stream.eatWhile(/[a-z\d\-]/i);
       return "meta";
     }
-    else {
-      stream.eatWhile(/[_\w\d]/);
-      if (stream.eat(":")) {
+    else if (PREFIX_START.test(ch) && stream.match(PREFIX_REMAINDER, false)) {
+        stream.match(PREFIX_REMAINDER)
```
marijnh, line 83: Calling `match` with `false` as second argument and then calling it again with the same regexp is exactly equivalent to calling it without a second argument, so it looks like this line could go and the `false` could be removed.

mode/sparql/sparql.js
```diff
@@ -33,6 +33,14 @@ CodeMirror.defineMode("sparql", function(config) {
...
                              "data", "copy", "to", "move", "add", "create", "drop", "clear", "load", "into"]);
   var operatorChars = /[*+\-<>=&|\^\/!\?]/;
+  var PN_CHARS_BASE =
+    "[A-Za-z\\u{00C0}-\\u{00D6}\\u{00D8}-\\u{00F6}\\u{00F8}-\\u{02FF}" +
+    "\\u{0370}-\\u{037D}\\u{037F}-\\u{1FFF}\\u{200C}-\\u{200D}\\u{2070}-\\u{218F}" +
+    "\\u{2C00}-\\u{2FEF}\\u{3001}-\\u{D7FF}\\u{F900}-\\u{FDCF}\\u{FDF0}-\\u{FFFD}\\u{10000}-\\u{EFFFF}]";
+  var PN_CHARS_U = PN_CHARS_BASE + "|_";
+  var PN_CHARS = PN_CHARS_U + "|[\\-0-9\\u{00B7}\\u{0300}-\\u{036F}\\u{203F}-\\u{2040}]";
+  var PREFIX_START = new RegExp(PN_CHARS_BASE, "u");
+  var PREFIX_REMAINDER = new RegExp("((" + PN_CHARS + "|\\.)*(" + PN_CHARS + "))?:", "u");
```
marijnh, line 43: CodeMirror still targets browsers that don't support `u` regexps. Do you think the upper, astral character range is important to support here?

mode/sparql/sparql.js
MarkBoyes: Will change as suggested.

mode/sparql/sparql.js
MarkBoyes: I took the RegExp from a project that has to be exactly conformant to the spec (since improved thanks to your feedback on the character classes, thanks!), but for CodeMirror this can be weakened: I will removed the necessity to compile with "u".

MarkBoyes:
I have committed changes answering today's observations. The resulting, limited, change still meets my user's requirement that "-" and "." characters in prefix names are handled correctly. Please rereview.

MarkBoyes:
Thanks for your patience.

PR #7000, opened 2022-11-03
Add favicon on the website

### LINKED ISSUE #6998: Missing favicon on the website
Hi Marijn. I noticed that there is missing the favicon on the v5 website. Can I make a pull request?

jdbruxelles:
Closes #6998

### DIFF
--- a/demo/activeline.html
+++ b/demo/activeline.html
@@ -2,6 +2,7 @@
 
 <title>CodeMirror: Active Line Demo</title>
 <meta charset="utf-8"/>
+<link rel=icon href="../style/logo.svg">
 <link rel=stylesheet href="../doc/docs.css">
 
 <link rel="stylesheet" href="../lib/codemirror.css">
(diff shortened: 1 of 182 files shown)


marijnh:
Oh, right, you'd have to change every single HTML file. I've added a redirect to my server config instead.

jdbruxelles:
I think I don't have access to the server or its code. How did you do that (even if the server is not accessible to me), please?

marijnh:
I added a redirect line to my Nginx config. And no, that's not something you could have done.

jdbruxelles:
I'm just trying to understand what you did...
Ok for the `301` redirection, but how do you add the favicon without the `link` tag? It seems to me that you have not done so.

If I understand well, as long as there is a physical icon (or by redirection) at the `/favicon.ico` location, the `link` tag in the html is not mandatory because browsers try to retrieve `/favicon.ico`, even without having defined it in the html?

marijnh:
Right, I made a forward from /favicon.ico, and since the browser fetches that anyway, even without link tag, it finds the icon.

jdbruxelles:
Good point! Thank you 🤗.

PR #7010, opened 2022-12-07
Use autocorrect and autocapitalize value of on instead of empty string

josephdpurcell:
Fixes https://github.com/codemirror/codemirror5/issues/7009 and ultimately should fix https://github.com/directus/directus/issues/16690.


josephdpurcell:
Even though this doesn't fix the issue, based on the iOS docs the values should be `on` instead of `""` (empty string), which is addressed by this PR. I did not check to see historical compatibility with older browsers or even cross compatibility with other browsers.

PR #7017, opened 2023-01-25
Add pipe operator to mlike keywords

Captain-Quack:
I reason that if languages like Clojure have symbols in purple, languages with the pipe operator should also include one of their more heavily used operators.


marijnh:
This does not work (as you could have rather easily found out), since that table is only consulted for tokens that match word characters. Also, CodeMirror modes in general assign the `"operator"` token style to operators, and use `"keyword"` for actual alphabetic keywords. So I'm going to decline this.

PR #7024, opened 2023-02-25
mode.allowIndefiniteTokenCalls = true 

bandinopla:
Trying to "fix" this issue https://github.com/codemirror/codemirror5/issues/7023

https://github.com/codemirror/codemirror5/blob/658bff7c56b7829aeabb8a914be5ca728d8aba0b/src/line/highlight.js#L161
I propose letting the mode define a `allowIndefiniteTokenCalls` property and set it to `true` to remove that hardcoded limit.

### DIFF
--- a/src/line/highlight.js
+++ b/src/line/highlight.js
@@ -158,7 +158,8 @@ function callBlankLine(mode, state) {
 }
 
 function readToken(mode, stream, state, inner) {
-  for (let i = 0; i < 10; i++) {
+  let maxAttempts = mode.allowIndefiniteTokenCalls? Number.MAX_VALUE : 10;
+  for (let i = 0; i < maxAttempts; i++) {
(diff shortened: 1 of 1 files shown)


marijnh:
Again, it's easy enough to do the looping in your mode's `token` function if you really have to. I don't think the library needs to support this.

bandinopla:
yeah i know the work arround is simple to do, but i also think the 10 cap is arbitrary and based on what you said it was to avoid infinite loops. The solution in my personal opinion was too drastic, you should remove it or improve the error message.

> Modes failing to advance the stream at all was a common source of infinite loops (and thus browser hangs) in the past. Thus, the library cutting it off at some point is intentional.

... because after 10 tries you already assume the stream failed to advance when it could be waiting for the 10+ run... it was too soon to call an infinite loop...

i do agree the need for this new flag is as unnecesary as the cap in that for loop.

PR #7044, opened 2023-06-29
Always render linter's tooltip inside of viewport

benjaminr-ps:


marijnh:
There is already code that does something like this in the library. Figuring out why it doesn't do what you want in this case would be the thing to do. Code that adds another layout cycle to every tooltip update, as this does, is definitely not going to be merged.

benjaminr-ps:
Please let me know if any changes needed for merge, or if I should squash.

marijnh:
Attached patch simplifies this a bit and reduces the number of reflows the code causes. Does it work for you?

PR #7045, opened 2023-06-29
Linter: Do not filter duplicates by message only

benjaminr-ps:
Additionally, consider the positions, since there might be the same error message at different places.

Example:
image


benjaminr-ps:
Thank you!

PR #7073, opened 2023-10-25
Support generics syntax in JSX with TS

lukehaas:
Currently, a function that uses generics will look like this when the mode is JSX with TypeScript.

Screenshot 2023-10-25 at 17 15 14

With this fix, it will now be highlighted correctly:

Screenshot 2023-10-25 at 17 17 19

### DIFF
--- a/mode/jsx/jsx.js
+++ b/mode/jsx/jsx.js
@@ -103,7 +103,7 @@
     }
 
     function jsToken(stream, state, cx) {
-      if (stream.peek() == "<" && stream.string.indexOf(',>', stream.start) < 0 && jsMode.expressionAllowed(stream, cx.state)) {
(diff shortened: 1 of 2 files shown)


marijnh:
Why does this assume there will be a trailing comma in the type parameter list? Is that a thing people use to work around this ambiguity in TSX code?

lukehaas:
@marijnh yes, it tells the typescript compiler to recognize that it's a type, not a JSX tag.

marijnh:
Merged as bcb86262e8a2a606bfa13ed47f8c6171b4d37ac9, followed up by adc4282471ee15e2c91b7da7d5398cd1b8eee978 to make false-positives a bit less likely.

Evike23:
[main](https://github.com/codemirror/codemirror5/actions/workflows/ci.yml)

PR #7075, opened 2023-11-10
[shadow dom] make codemirror work inside closed shadow root

pavelfeldman:
Ref https://github.com/codemirror/codemirror5/issues/7074

This change should be:
- noop for old <2017 browsers
- noop for modern browsers when codemirror is in the document or in the open shadow root
- should be fixing input (focus, pasting) when codemirror is in the closed shadow root

The problem was that we were going from element to ownerDocument and then descending. And closer roots can only be pierced one way. The solution is to not climb to the ownerDocument, but stop at the local root node (nearest shadow root).

Works on Safari.


marijnh:
Thanks, this looks solid.

PR #7080, opened 2023-12-08
Adds Support For HCL

dakota-wagner:
Adds support for HCL.

image


marijnh:
No support for new languages is added to the main distribution anymore. The way to publish this would be to release it as a separate npm package.

PR #7091, opened 2024-03-29
Toggle Comment: Uncomment: Allow uncomment of selection containing an inner comment

davidfstr:
* [x] Preexisting tests pass
* [x] New tests added
* [x] All new code passes local lint rules

Explanation of changed lines inside the "comment" addon:
Screenshot 2024-03-27 at 1 59 29 PM

### DIFF
--- a/addon/comment/comment.js
+++ b/addon/comment/comment.js
@@ -174,11 +174,7 @@
     if (open == -1) return false
     var endLine = end == start ? startLine : self.getLine(end)
     var close = endLine.indexOf(endString, end == start ? open + startString.length : 0);
-    var insideStart = Pos(start, open + 1), insideEnd = Pos(end, close + 1)
-    if (close == -1 ||
-        !/comment/.test(self.getTokenTypeAt(insideStart)) ||
-        !/comment/.test(self.getTokenTypeAt(insideEnd)) ||
-        self.getRange(insideStart, insideEnd, "\n").indexOf(endString) > -1)
+    if (close == -1)
       return false;
 
     // Avoid killing block comments completely outside the selection.
--- a/test/comment_test.js
+++ b/test/comment_test.js
@@ -115,4 +115,16 @@ namespace = "comment_";
     cm.setCursor(1, 0)
     cm.execCommand("toggleComment")
   }, "<!-- foo\nbar -->", "<!-- foo\nbar -->")
+  
+  var beforeToggleComment = "\nAAA\n    <!-- BBB -->\nCCC\n\n";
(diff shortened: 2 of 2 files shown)


marijnh:
Given that these lines were there, they were probably not just randomly typed there, but had a reason. Git blame points at 66c68db0c4fcbc6bb6ce735c2def8f03c3f43478, which points at issue https://github.com/codemirror/codemirror5/issues/4641 , which seems like it would be regressed again by merging this.

Most languages don't even have a concept of nested comments, so I'm not sure it's going to be possible to define the thing you're trying to do here in a solid way.

davidfstr:
Thanks for taking a look @marijnh. -- And sorry for the delay on my end. I was out of town all last week.

> issue https://github.com/codemirror/codemirror5/issues/4641 , which seems like it would be regressed again by merging this.

Reading that issue, I think that the code change I made would not regress the example given in the issue. **I'll write a regression test that covers the example from that issue to confirm, and add it to this PR.**

> Most languages don't even have a concept of nested comments, so I'm not sure it's going to be possible to define the thing you're trying to do here in a solid way.

I'm trying to ensure that the Toggle Comment action, when done twice, ends up at the original state most of the time. VS Code (but not Sublime Text) get back to the original state with the example given in this PR's test code.

davidfstr:
@marijnh I've pushed to this PR a regression test for the issue that you mentioned (https://github.com/codemirror/codemirror5/issues/4641). This PR does not regress that issue.

Beyond that issue, were there any other behaviors you were concerned this PR could break?

marijnh:
There's also https://github.com/codemirror/codemirror5/issues/1975 , which seems at odds with what you're trying to do here. I feel uncommenting probably shouldn't remove things that are not actually comment, syntactically speaking.

marijnh:
> Hypothetically CodeMirror could be extended to recognize a comment nested inside an outer comment, so that the outer comment is fully (and correctly) identified as a comment. 

No, that's the thing—if the language doesn't support nested comments, nested comments shouldn't be highlighted as such, since that will just confuse the user into thinking their code is parsed differently than it actually is. Since this version of the library isn't actively developed, I don't really want to change a lot about how it works and risk regressing things for existing users.

davidfstr:
> Since this version of the library isn't actively developed, I don't really want to change a lot about how it works and risk regressing things for existing users.

Makes sense. I'll just keep the proposed changes in my company's (small) internal fork of CodeMirror 5 then.

I'll open a new PR containing just the new regression tests from this PR, so that at least those can be integrated.

PR #7105, opened 2024-09-05
Dart: support digit separators

srawlins:
Dart is [adding support for digit separators](https://github.com/dart-lang/language/blob/main/working/digit-separators/feature-specification.md), the underscore, like `1_000_000` or `0x33_33_33`. The Dart playground, [DartPad](https://dartpad.dev/), still runs on codemirror5, so this would be a welcome fix.

### DIFF
--- a/mode/dart/dart.js
+++ b/mode/dart/dart.js
@@ -44,6 +44,8 @@
     blockKeywords: set(blockKeywords),
     builtin: set(builtins),
     atoms: set(atoms),
+    // clike numbers without the suffixes, and with '_' separators.
+    number: /^(?:0x[a-f\d_]+|(?:[\d_]+\.?[\d_]*|\.[\d_]+)(?:e[-+]?[\d_]+)?)/i,
     hooks: {
       "@": function(stream) {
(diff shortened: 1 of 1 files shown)


marijnh:
That regexp seems to allow underscores anywhere, whereas the linked proposal only allows them between digits. I think this'll make `_` tokenize as a number, which doesn't seem right.

srawlins:
As I understand it, the `number` pattern is only checked when we see `numberStart`, line 65 in `clike.js`. With manual testing, I see that `_` is styled as a "variable."

As for other spots where `_` is not allowed (e.g. in `1_._0`, neither underscore is allowed), I think it is fine to allow them for the purposes of in-progress typing, and error-recovery. For example, in the Dart parser, we _parse_ `1_._0` as a number literal, and also report an error. This reduces errors-carried-forward. And for a trailing `_` like `100_`, it is fine to allow it for code-in-progress.

If you think we should be more strict here and only parse error-free numbers as numbers, I can amend the PR.

marijnh:
Oh, indeed, there's a separate regexp for the start of the number. Then this is close enough, indeed.

PR #7108, opened 2024-09-17
Update realworld.html to include peerfect

Oussamahamani:
added peerfect in the real-world uses


marijnh:
I'm no longer adding entries here. In fact, I was planning to drop this page entirely.

Oussamahamani:
can i ask why you decided to do that?

marijnh:
Version 5 is legacy at this point, and I expect a lot of the links there had rotted already. I don't feel it's worthwhile to keep maintaining that page.

PR #7113, opened 2024-12-04
Update keywords with the appropriate uppercase option

amarbrkic:
Since a lot of people like to write syntax keywords in uppercase, this would be a nice add-on.


marijnh:
It seems Pascal is simply case-insensitive. Attached patch solves this in another, more general way.

Scutua (approved):

PR #7123, opened 2025-03-16
[sql mode] Add identifier for PostgreSQL

vrana:


marijnh:
Thanks! Strange that this wasn't already handled.

PR #7125, opened 2025-03-31
clip-path issue fixed in Chrome 106

zaiddabaeen:
In Chrome 105 due to a bug, `clip-path` needed to be set as pointer events were ineffective outside CodeMirror editor instance but within paddings and margins of its elements. This has been resolved in Chrome 106 as seen [here](https://issues.chromium.org/issues/40863245#comment21).


marijnh:
Thanks!

PR #7129, opened 2025-06-24
fix bug in show-hints inset

alirrah:
When the show-hints box does not have space to be displayed below, the box will be displayed in the wrong place, which is also reported in [this issue](https://github.com/codemirror/codemirror5/issues/6993). I fixed this issue by making some changes, which you can see in the videos below.

The reason I added `marginBottom` is so that I can give the box some space from the bottom (I didn't want the box to stick to the bottom of the page).

before:

[vokoscreenNG-2025-06-24_09-52-21.webm](https://github.com/user-attachments/assets/b79adcc4-659e-4aa7-bfbe-3bb4e2aabcbe)

after:

[vokoscreenNG-2025-06-24_09-49-06.webm](https://github.com/user-attachments/assets/a4267a16-af4b-4c31-a6c8-45958aa136c4)

### DIFF
--- a/addon/hint/show-hint.js
+++ b/addon/hint/show-hint.js
@@ -255,6 +255,7 @@
     }
 
     var container = completion.options.container || ownerDocument.body;
+    var marginBottom = completion.options.marginBottom || 0;
     var pos = cm.cursorCoords(completion.options.alignWithWord ? data.from : null);
     var left = pos.left, top = pos.bottom, below = true;
     var offsetLeft = 0, offsetTop = 0;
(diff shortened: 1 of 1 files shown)


marijnh:
Do you have a minimal HTML page that shows the current issue happening? My response to the issue you linked suggests I haven't been able reproduce the issue.

marijnh:
I can confirm that that plus should be a minus. I've corrected that in attached patch. I don't think adding an option like `marginBottom` is a good idea—firstly, it suggests we'd also need `marginTop`, `marginLeft`, and `marginRight`, and secondly, I don't think your implementation, as it stands, properly implements a bottom margin (just blindly moving the tooltip up will cause all the other computations to work with the wrong position). As such, I don't intend to merge that part, and suggest you work with something like transparent borders around the tooltip to get an effect like this.

